PR #49, opened 2016-06-17
santa-driver: Re-implement caching to avoid OSDictionary.

russellhancox:
OSDictionary is not well-suited to our needs and locking is quite expensive.
This commit:

  + Replaces all uses of OSDictionary with a new SantaCache class, which
    is a size-limited array hash table with per-bucket locking. It works with
    uint64_t keys, which is perfect for our needs.
  + Adds a unit test for SantaCache.
  + Removes SantaCachedDecision and SantaPIDAndPPID, which only existed
    because OSDictionary can only store OSObject subclasses.
  + Removes a lot of locking logic from SantaDecisionManager as the
    locking is now handled inside SantaCache and is therefore and is
    much more granular.
  + Removes the timed cache expiration for ALLOW decisions. This was
    originally to ensure executions were logged regularly but as we're
    logging all executions nowadays this is longer particularly useful.

SantaCache's configured load factor and hashing function may need tweaking
over-time but this is already a little faster and uses less memory
than what existed before.


Source/santa-driver/SantaCache.h
```diff
@@ -0,0 +1,308 @@
...
+    Initialize a newly created cache.
+   
+    @param maximum_size The maximum number of entries in this cache. Once this
+        number is reached all the entries will be purged.
+    @param per_bucket The maximum number of entries in each bucket. A higher
+        number will result in better performance but higher memory usage.
+        Cannot be higher than 126 due to type requirements.
+  */
+  SantaCache(uint64_t maximum_size = 10000, uint8_t per_bucket = 5) {
+    if (unlikely(per_bucket > 126)) per_bucket = 126;
```
tburgin: Add a check to make sure `per_bucket` != 0.

tburgin:
Still wading through. More soon.

Source/santa-driver/SantaCache.h
```diff
@@ -0,0 +1,308 @@
...
+      }
+    }
+    unlock(bucket);
+    return zero_;
+  }
+
+  /**
+    Set an element in the cache.
+   
+    @return if an existing value was replaced, the previous value, otherwise zero_
```
tburgin: Add the `maximum_size` comment here too.

tburgin:
Don't know why the travis tests fail. All test pass for me.
[LGTM]

PR #50, opened 2016-06-26
Optimize logging by using one pre-allocated NSMutableString

georgekola:
I am breaking it into small commits to make it easy to be reviewed and merged in


georgekola:
Travis test is failing because of unrelated issue  in [SantaCacheTest testCount]

PR #52, opened 2016-06-29
Correctly use pread

georgekola:
pread can return less than the chunk size (e.g. signal caught in the
middle) and hence we need to handle it. This change also cleans up the
hash function and makes it more performant.

### DIFF
--- a/Source/common/SNTFileInfo.m
+++ b/Source/common/SNTFileInfo.m
@@ -102,7 +102,8 @@ extern NSString *const NSURLQuarantinePropertiesKey WEAK_IMPORT_ATTRIBUTE;
 #pragma mark Hashing
 
 - (void)hashSHA1:(NSString **)sha1 SHA256:(NSString **)sha256 {
-  const int chunkSize = 262144;
+  const int MAX_CHUNK_SIZE = 256 * 1024;  // 256 KB
+  const size_t chunkSize = _fileSize > MAX_CHUNK_SIZE ? MAX_CHUNK_SIZE : _fileSize;
   char chunk[chunkSize];
 
   CC_SHA1_CTX c1;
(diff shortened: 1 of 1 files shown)


Source/common/SNTFileInfo.m
```diff
@@ -112,52 +113,48 @@ - (void)hashSHA1:(NSString **)sha1 SHA256:(NSString **)sha256 {
   if (sha256) CC_SHA256_Init(&c256);
 
   int fd = self.fileHandle.fileDescriptor;
-  
+
   fcntl(fd, F_RDAHEAD, 1);
   fcntl(fd, F_RDADVISE, 1);
-
+  ssize_t bytesRead;
```
russellhancox: Please add an empty line after the two fcntl calls.

Source/common/SNTFileInfo.m
```diff
@@ -112,52 +113,48 @@ - (void)hashSHA1:(NSString **)sha1 SHA256:(NSString **)sha256 {
...
-    } else if (errno == EINTR) {
+    } else if (bytesRead == -1 && errno == EINTR) {
       continue;
     } else {
       return;
     }
   }
-
-  fcntl(fd, F_RDAHEAD, 0);
   fcntl(fd, F_RDADVISE, 0);
```
russellhancox: Please restore the empty lines around this call, it's clearer when it's separate. Also, why remove the call to disable F_RDAHEAD?

Source/common/SNTFileInfo.m
```diff
@@ -112,52 +113,48 @@ - (void)hashSHA1:(NSString **)sha1 SHA256:(NSString **)sha256 {
...
-    unsigned char dgst[CC_SHA1_DIGEST_LENGTH];
-    CC_SHA1_Final(dgst, &c1);
+    unsigned char digest[CC_SHA1_DIGEST_LENGTH];
+    CC_SHA1_Final(digest, &c1);
+    char sha1hex[CC_SHA1_DIGEST_LENGTH * 2 + 1];
 
-    NSMutableString *buf = [[NSMutableString alloc] initWithCapacity:CC_SHA1_DIGEST_LENGTH * 2];
-    for (int i = 0; i < CC_SHA1_DIGEST_LENGTH; ++i) {
-      [buf appendFormat:@"%02x", (unsigned char)dgst[i]];
+    for (int i = 0; i < CC_SHA1_DIGEST_LENGTH / 4; i += 4) {
```
russellhancox: Remove the `/ 4`, otherwise only quarter of the hash is processed.

Source/common/SNTFileInfo.m
```diff
@@ -112,52 +113,48 @@ - (void)hashSHA1:(NSString **)sha1 SHA256:(NSString **)sha256 {
...
+    *sha1 = [[NSString alloc] initWithBytes:sha1hex
+                                     length:CC_SHA1_DIGEST_LENGTH * 2
+                                   encoding:NSUTF8StringEncoding];
   }
   if (sha256) {
-    unsigned char dgst[CC_SHA256_DIGEST_LENGTH];
-    CC_SHA256_Final(dgst, &c256);
+    unsigned char digest[CC_SHA256_DIGEST_LENGTH];
+    CC_SHA256_Final(digest, &c256);
+    char sha256hex[CC_SHA256_DIGEST_LENGTH + 1];
```
russellhancox: The digest length needs to be doubled

Source/common/SNTFileInfo.m
```diff
@@ -112,52 +113,48 @@ - (void)hashSHA1:(NSString **)sha1 SHA256:(NSString **)sha256 {
...
 
-    NSMutableString *buf = [[NSMutableString alloc] initWithCapacity:CC_SHA256_DIGEST_LENGTH * 2];
-    for (int i = 0; i < CC_SHA256_DIGEST_LENGTH; ++i) {
-      [buf appendFormat:@"%02x", (unsigned char)dgst[i]];
+    for (int i = 0; i < CC_SHA256_DIGEST_LENGTH / 4; i += 4) {
```
russellhancox: Again remove the `/ 4`.

russellhancox:
If we're going for performance, why not remove the loops and have a single `sprintf` call? Seems to be about 2-4% faster.

PR #58, opened 2016-08-09
Performance improvements

russellhancox:
Mostly perf improvements:

* Use msleep/wakeup instead of sleeping for a fixed period while waiting for responses from daemon
* Cache user/group id->name lookups as getpwuid/getgrgid can be quite slow
* Rename the Dismiss button to Ignore

### DIFF
--- a/Source/santad/SNTEventLog.m
+++ b/Source/santad/SNTEventLog.m
@@ -29,6 +29,13 @@
 @interface SNTEventLog ()
 @property NSMutableDictionary *detailStore;
 @property dispatch_queue_t detailStoreQueue;
+
+// Caches for uid->username and gid->groupname lookups.
+// Both dictionaries must be accessed from the nameMapQueue
+// to enforce thread-safety.
+@property NSMutableDictionary *userNameMap;
+@property NSMutableDictionary *groupNameMap;
+@property dispatch_queue_t nameMapQueue;
 @end
 
 @implementation SNTEventLog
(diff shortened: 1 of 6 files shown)


Source/santad/SNTEventLog.m
```diff
@@ -29,6 +29,10 @@
 @interface SNTEventLog ()
 @property NSMutableDictionary *detailStore;
 @property dispatch_queue_t detailStoreQueue;
+
+@property NSMutableDictionary *userNameMap;
```
tburgin: Add a comment to userNameMap and groupNameMap. Something like:
Do not access directly, this is used for caching. Only access from the nameMapQueue.

Source/santad/SNTExecutionController.m
```diff
@@ -365,44 +373,36 @@ - (void)printMessage:(NSString *)msg toTTYForPID:(pid_t)pid {
     return;
   }
 
-  NSString *devPath = [NSString stringWithFormat:@"/dev/%s", devname(taskInfo.e_tdev, S_IFCHR)];
-  int fd = open(devPath.UTF8String, O_WRONLY | O_NOCTTY);
-  @try {
-    NSFileHandle *fh = [[NSFileHandle alloc] initWithFileDescriptor:fd closeOnDealloc:YES];
-    [fh writeData:[msg dataUsingEncoding:NSUTF8StringEncoding]];
-  } @catch (NSException *) { /* do nothing */ }
+  char devPath[32] = "/dev/";
```
tburgin: 32 seems arbitrary. Either use `PATH_MAX` or go the other way and count the max tty sessions. `kern.tty.ptmx_max` seems to take a max of 999. So the max "/dev/tty999" is 12 bytes with null termination.

Source/santad/SNTEventLog.m
russellhancox: Added a comment

Source/santad/SNTExecutionController.m
russellhancox: Added a comment

tburgin:
LGTM

PR #90, opened 2016-09-09
santa-driver: Remove the static wrappers in SantaDriverClient

russellhancox:
SantaDriverClient was implemented to have static functions that call instance
methods passing appropriate arguments. While this works and is 'technically correct' (best kind),
it's a bit messy and hard to read.

### DIFF
--- a/Source/santa-driver/SantaDriverClient.cc
+++ b/Source/santa-driver/SantaDriverClient.cc
@@ -110,10 +110,15 @@ IOReturn SantaDriverClient::clientMemoryForType(
 
 #pragma mark Callable Methods
 
-IOReturn SantaDriverClient::open() {
-  if (isInactive()) return kIOReturnNotAttached;
+IOReturn SantaDriverClient::open(
+    OSObject *target,
+    void *reference,
+    IOExternalMethodArguments *arguments) {
+  SantaDriverClient *me = OSDynamicCast(SantaDriverClient, target);
+  if (!me) return kIOReturnBadArgument;
 
-  if (!myProvider->open(this)) {
+  if (me->isInactive()) return kIOReturnNotAttached;
+  if (!me->myProvider->open(me)) {
     LOGW("A second client tried to connect.");
     return kIOReturnExclusiveAccess;
   }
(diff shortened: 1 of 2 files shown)


Source/santa-driver/SantaDriverClient.cc
```diff
@@ -110,97 +110,78 @@ IOReturn SantaDriverClient::clientMemoryForType(
...
 
-IOReturn SantaDriverClient::static_clear_cache(
-    com_google_SantaDriverClient *target,
-    void *reference,
-    IOExternalMethodArguments *arguments) {
-  if (!target) return kIOReturnBadArgument;
-  return target->clear_cache();
-}
+IOReturn SantaDriverClient::cache_count(
+    OSObject *target, void *reference, IOExternalMethodArguments *args) {
```
tburgin: /s/args/arguments

Source/santa-driver/SantaDriverClient.cc
```diff
@@ -110,97 +110,78 @@ IOReturn SantaDriverClient::clientMemoryForType(
...
-    void *reference,
-    IOExternalMethodArguments *arguments) {
-  if (!target) return kIOReturnBadArgument;
-  return target->open();
-}
+IOReturn SantaDriverClient::allow_binary(
+    OSObject *target, void *reference, IOExternalMethodArguments *arguments) {
+  SantaDriverClient *me = OSDynamicCast(SantaDriverClient, target);
+  if (!me) return kIOReturnBadArgument;
+  if (arguments->scalarInput == nullptr) return kIOReturnBadArgument;
```
tburgin: If the userland caller of IOConnectCallScalarMethod promises an `input` with `inputCnt` but does not give one, the caller will crash.

```
Thread 11 Crashed:: Dispatch queue: com.apple.NSXPCConnection.user.15028
0   libsystem_platform.dylib           	0x00007fff986d9fbd _platform_memmove$VARIANT$Haswell + 157
1   libsystem_c.dylib                  	0x00007fff913d7f13 __memcpy_chk + 22
2   com.apple.framework.IOKit          	0x00007fff915b3b19 io_connect_method + 120
3   com.apple.framework.IOKit          	0x00007fff915543c4 IOConnectCallScalarMethod + 111
```

If the userland caller provides no `inputCnt` promise and no `input`, the decision threads are never woken up and I experience deadlock. We may want to look into other options for preventing deadlock if the vnode_ids don't make the round trip or are corrupted.

Issue #112, opened 2016-10-10
extra bytes prepended to (esp. json) output from `santactl fileinfo ... --key <path>` can be dropped?


arubdesu:
I noticed that I'm getting extra bytes in the front of a string outputted from `fileinfo's` `--key` flag, and the inimitable @pudquick figured out it was in the color codes [embedded in it](https://github.com/google/santa/blob/master/Source/santactl/Commands/SNTCommandFileInfo.m#L479). It definitely shouldn't be in there for `json` output, but I'd not expect colored output to be particularly useful for values returned with `--key`, either - more pressing is trying to use the python json module to `.loads` the value wasn't working for me.

russellhancox:
> It definitely shouldn't be in there for json output

It isn't?

```sh
santactl fileinfo --json --key Rule /usr/bin/yes
{
  "Rule" : "Blacklisted"
}
```

arubdesu:
Au contraire mon frère, to my chagrin
```python
#!/usr/bin/python
import json
import subprocess

path = "/Applications/WWDC.app"
cmd = ["/usr/local/bin/santactl", "fileinfo", "--cert-index", "1", "--key", "SHA-256", "--json",
       path]
output = subprocess.check_output(cmd)
with open("/tmp/jsony_output", 'wb') as outfile:
    outfile.write(output)
```

<img width="383" alt="screenshot 2016-10-10 17 12 17" src="https://cloud.githubusercontent.com/assets/388808/19251345/cc4a3918-8f0c-11e6-9253-28b0f3f0918e.png">

russellhancox:
Aha: [2K is not a color code, it's the sequence for _clear line_. It's the only escape sequence I can see [that isn't checking](https://github.com/google/santa/blob/master/Source/santactl/Commands/SNTCommandFileInfo.m#L479) for JSON output or the presence of a TTY.

Tom's fixing now.

PR #113, opened 2016-10-11
SNTCommandFileInfo global json bool shared between class and instance methods

tburgin:
*  https://github.com/google/santa/issues/112 Add missing pretty output check around the clear line escape sequence
  *  Collapses all pretty output checks to a single function
  *  Only run isatty() once saving some time

### DIFF
--- a/Source/santactl/Commands/SNTCommandFileInfo.m
+++ b/Source/santactl/Commands/SNTCommandFileInfo.m
@@ -50,6 +50,18 @@ static NSString *const kValidUntil = @"Valid Until";
 static NSString *const kSHA256 = @"SHA-256";
 static NSString *const kSHA1 = @"SHA-1";
 
+// global json output flag
+static BOOL json = NO;
+
+BOOL PrettyOutput() {
+  static int tty = 0;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    tty = isatty(STDOUT_FILENO);
+  });
+  return (tty && !json);
+}
+
 #pragma mark SNTCommandFileInfo
 
 @interface SNTCommandFileInfo : NSObject<SNTCommand>
@@ -87,9 +99,6 @@ typedef id (^SNTAttributeBlock)(SNTCommandFileInfo *);
 // Common Date Formatter
 @property(nonatomic) NSDateFormatter *dateFormatter;
 
-// CLI option
-@property(nonatomic) BOOL jsonOutput;
-
 // Block Helpers
 - (NSString *)humanReadableFileType:(SNTFileInfo *)fi;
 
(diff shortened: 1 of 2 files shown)


Source/santactl/Commands/SNTCommandFileInfo.m
```diff
@@ -134,9 +141,7 @@ - (SNTFileInfo *)fileInfo {
   if (!_fileInfo) {
     _fileInfo = [[SNTFileInfo alloc] initWithPath:self.filePath];
     if (!_fileInfo) {
-      if (isatty(STDOUT_FILENO) && !self.jsonOutput) {
-        printf("\rInvalid or empty file: %s\n", self.filePath.UTF8String);
-      }
+      if (PrettyOutput()) printf("\rInvalid or empty file: %s\n", self.filePath.UTF8String);
```
russellhancox: This error should go to stderr rather than stdout and be printed regardless of json/tty except for the \r.

Source/santactl/Commands/SNTCommandFileInfo.m
```diff
@@ -305,13 +310,13 @@ - (SNTAttributeBlock)rule {
           output = @"None".mutableCopy;
           break;
       }
-      if ((SNTEventStateAllow & s) && isatty(STDOUT_FILENO) && !fi.jsonOutput) {
+      if ((SNTEventStateAllow & s) && PrettyOutput()) {
```
russellhancox: I think the "PrettyOutput" condition should be on its own to avoid repeating it:

```objc
if (PrettyOutput()) {
  if (s & SNTEventStateAllow) {
    ...
  } else if, etc. etc.
```

Source/santactl/Commands/SNTCommandFileInfo.m
```diff
@@ -421,13 +421,10 @@ + (void)runWithArguments:(NSArray *)arguments daemonConnection:(SNTXPCConnection
...
     NSBlockOperation *hashOperation = [NSBlockOperation blockOperationWithBlock:^{
-      if (isatty(STDOUT_FILENO) && !jsonOutput) {
-        printf("\rCalculating %lu/%lu", ++hashed, filePaths.count);
-      }
+      if (PrettyOutput()) printf("\rCalculating %lu/%lu", ++hashed, filePaths.count);
 
       SNTCommandFileInfo *fi = [[self alloc] initWithFilePath:obj
-                                             daemonConnection:daemonConn
-                                                   jsonOutput:jsonOutput];
+                                             daemonConnection:daemonConn];
```
russellhancox: This looks like it'll fit on one line now.

Source/santactl/Commands/SNTCommandFileInfo.m
tburgin: Sending it to stderr. Still need the \r for formatting when viewing stdout and stderr together.

Source/santactl/Commands/SNTCommandFileInfo.m
tburgin: Done

Source/santactl/Commands/SNTCommandFileInfo.m
```diff
@@ -421,13 +421,10 @@ + (void)runWithArguments:(NSArray *)arguments daemonConnection:(SNTXPCConnection
 
   [filePaths enumerateObjectsUsingBlock:^(id obj, NSUInteger idx, BOOL *stop) {
     NSBlockOperation *hashOperation = [NSBlockOperation blockOperationWithBlock:^{
-      if (isatty(STDOUT_FILENO) && !jsonOutput) {
-        printf("\rCalculating %lu/%lu", ++hashed, filePaths.count);
-      }
+      if (PrettyOutput()) printf("\rCalculating %lu/%lu", ++hashed, filePaths.count);
```
tburgin: Since this function would only be used in two places, leaving the PrettyOutput() check.

Source/santactl/Commands/SNTCommandFileInfo.m
tburgin: Yup

PR #120, opened 2016-10-13
santad: Fix printer-proxy workaround

russellhancox:


arubdesu:
Huh, I just ran in to this today!
screenshot 2016-10-13 14 36 22

Issue #121, opened 2016-10-13
Push Notifications


tburgin:
An issue to track the integration of a push notification client for instant rule syncing.
Most likely a FCM client with an option to set a controlling server URL.

PR #129, opened 2016-10-30
run santactl as a sync daemon

tburgin:
-  New events are sent directly from santad --> santactl for immediate upload. No need to read from events.db for single event uploads. All events are still saved to events.db and those not uploaded immediately are uploaded during full syncs.
-  Continues to perform full syncs every 10 minutes. Only now this is scheduled and performed by santactl.
-  The event `idx` field in events.db is now randomly generated at event creation time instead of being set at regeneration time. This will allow santactl to remove events from events.db when it uploads a single event from santad.
-  First step in supporting FCM push notifications https://github.com/google/santa/issues/121

### DIFF
--- a/Source/santad/DataLayer/SNTEventTable.m
+++ b/Source/santad/DataLayer/SNTEventTable.m
@@ -15,7 +15,6 @@
 #import "SNTEventTable.h"
 
 #import "MOLCertificate.h"
-#import "SNTLogging.h"
 #import "SNTStoredEvent.h"
 
 @implementation SNTEventTable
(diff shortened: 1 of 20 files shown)


Source/santad/DataLayer/SNTEventTable.m
```diff
@@ -98,8 +98,9 @@ - (BOOL)addStoredEvent:(SNTStoredEvent *)event {
 
   __block BOOL success = NO;
   [self inTransaction:^(FMDatabase *db, BOOL *rollback) {
-    success = [db executeUpdate:@"INSERT INTO 'events' (filesha256, eventdata) VALUES (?, ?)",
-                                event.fileSHA256, eventData];
+    success = [db executeUpdate:@"INSERT INTO 'events' (idx, filesha256, eventdata)"
+                                @"VALUES (?, ?, ?)",
+                                event.idx, event.fileSHA256, eventData];
```
russellhancox: Can event.idx ever be unset?

tburgin: Not likely as it's only caller sets idx right before calling the method. Added idx to the nil early return check at the beginning of the method just in case.

PR #138, opened 2017-01-04
santactl sync: reachability and notification updates santad: syncd xpc updates

tburgin:
*  `santactl sync`: Retry failed full syncs once the url is reachable
*  `santactl sync`: rule_sync action messages contain a hash and the name of the binary rule generated for the machine. That data is cached and looked up later during the rule download phase. To avoid potentially displaying the wrong name for a current user action, I propose we display all downloaded rules names that can be found in the cache. We then remove each cached item as we go through the list. This will give more accurate feedback and mitigate laptops waking from sleep from displaying an older cached name in response to a current user action. If there are multiple rules generated while a machine is asleep, upon wake the notifications are grouped like so.
screen shot 2017-01-04 at 2 22 52 pm
*  `santad`: Allow only one syncd connection at any given time to be set.

### DIFF
--- a/Source/santactl/Commands/sync/SNTCommandSyncManager.m
+++ b/Source/santactl/Commands/sync/SNTCommandSyncManager.m
@@ -14,6 +14,8 @@
 
 #import "SNTCommandSyncManager.h"
 
+#import <SystemConfiguration/SystemConfiguration.h>
+
 #import <MOLAuthenticatingURLSession.h>
 #import <MOLFCMClient/MOLFCMClient.h>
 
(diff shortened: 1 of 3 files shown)


Source/santactl/Commands/sync/SNTCommandSyncManager.m
```diff
@@ -36,16 +38,30 @@
...
 @property(nonatomic) NSCache *dispatchLock;
 @property(nonatomic) NSCache *ruleSyncCache;
 @property MOLFCMClient *FCMClient;
 @property(nonatomic) SNTXPCConnection *daemonConn;
 @property BOOL targetedRuleSync;
+@property(nonatomic) BOOL reachable;
 @end
 
+// Called when the network state changes
+static void reachabilityHandler(SCNetworkReachabilityRef target, SCNetworkReachabilityFlags flags,
```
russellhancox:
```c
static void reachabilityHandler(
    SCNetworkReachabilityRef target, SCNetworkReachabilityFlags flags, void *info) {
```

tburgin:
Currently, executing "santactl sync --daemon" will connect to santad and set the listener to that santactl process. That new process would start receiving the syncd messages from santad. The santactl process that santad spins up would be left dangling.

This change "locks in" the first santactl to connect to santad. Since santad starts and re-spawns santactl if it falls over, the correct santactl should always be connected.

PR #143, opened 2017-02-03
santad: Clear cache when regexes change.

### LINKED ISSUE #142: Delayed effect of a BlacklistRegex update via a PreFlight response
The updated regex is visible using `sudo defaults read /var/db/santa/config.plist` right after a sync. The previously blocked apps are allowed, as expected, but the apps which are supposed to be blocked by the new regex are still working. If santad is restarted, they are blocked (so, not a regex issue.)

russellhancox:
When white/black-list regexes are changed clear the kernel cache so the regexes are able to take effect immediately. Fixes #142


tburgin:
Doesn't look like review comments send notifications.

Issue #151, opened 2017-03-09
Potential Memory Leak


LtMama:
I recently noticed a proportional trend between Santa app and santad's memory usage and a mac's uptime.
I've looked at multiple macs in our fleet (running 10.10/10.11/10.12) and it appears to be consistent with throughout.

We're running Santa pretty lightweight, Default configuration with 4 blacklisted binaries (OS X Installers)

Example of the uptime/usage trend:

Uptime | Total Santa Memory usage
------------ | -------------
1 hour | 7.7 MB
7 days | 230 MB
15 days | ~520MB
42 days | ~890MB
55 days | ~1200MB

![screen shot 2017-03-09 at 17 09 24](https://cloud.githubusercontent.com/assets/7572589/23761987/9b6d5130-04ec-11e7-803b-403389db5160.png)

Let me know if there's anything else I can provide to help

russellhancox:
Can you confirm you're running the latest version (0.9.16)? We're trying to reproduce but none of our machines seem to have unexpectedly high memory use. 

We think it could be due to fast-user-switching which we don't really use but Santa is designed to support. A recent commit (5e4b8350ab), which hasn't yet been included in a release could have inadvertently fixed this but we'd like to reproduce to confirm.

tburgin:
Could you also provide your config?
```
cat /var/db/santa/config.plist
```

LtMama:
@russellhancox 
Turns out majority of users were still running 0.9.14 (we packaged 0.9.16 back in Feb, but didn't actual push it out to everyone _**face** **palm**_ Pushed out now).

Having said that, the first 3 entries in that table _were_ running 0.9.16, so although smaller sample size, might still be the case. Also found another one running 0.9.16, uptime 28days, Santa using 1.8GB.

@tburgin 
Hmmm, there doesn't appear to be a santa config.plist on any mac that I've checked
> /var/db/santa
>-rw-------   1 root  wheel  5169152  9 Mar 19:19 events.db
>-rw-------   1 root  wheel    12288 10 Mar 09:51 rules.db
>-rw-------   1 root  wheel    12824 10 Mar 09:51 rules.db-journal

Currently, we've just left Santa in the default MONITOR mode, and been adding items to the blacklist by sending a payload-free package containing 4 santactl commands

tburgin:
Thanks, I am still unable to reproduce the issue. When you come across a machine that is in this state can you grab a memory graph and share it here?

```
[...]

PR #154, opened 2017-03-17
Running without a config fixes

tburgin:
*  	https://github.com/google/santa/issues/151 common: capture fileSystemRepresentation as a local variable to avoid a leak in it's implementation
    *  A big thanks to @LtMama for reporting this issue and providing memory graphs to track down the issue.
*  	santactl/status: check for instant notification status only when there is a sync server

### DIFF
--- a/Source/santactl/Commands/SNTCommandStatus.m
+++ b/Source/santactl/Commands/SNTCommandStatus.m
@@ -110,12 +110,15 @@ REGISTER_COMMAND_NAME(@"status")
   NSString *lastRuleSyncSuccessStr =
       [dateFormatter stringFromDate:lastRuleSyncSuccess] ?: lastSyncSuccessStr;
   BOOL syncCleanReqd = [[SNTConfigurator configurator] syncCleanRequired];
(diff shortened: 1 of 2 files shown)


Source/santactl/Commands/SNTCommandStatus.m
```diff
@@ -110,12 +110,15 @@ + (void)runWithArguments:(NSArray *)arguments daemonConnection:(SNTXPCConnection
...
       [dateFormatter stringFromDate:lastRuleSyncSuccess] ?: lastSyncSuccessStr;
   BOOL syncCleanReqd = [[SNTConfigurator configurator] syncCleanRequired];
-  __block BOOL pushNotifications;
-  dispatch_group_enter(group);
-  [[daemonConn remoteObjectProxy] pushNotifications:^(BOOL response) {
-    pushNotifications = response;
-    dispatch_group_leave(group);
-  }];
+
+  __block BOOL pushNotifications = FALSE;
```
russellhancox: s/FALSE/NO

PR #163, opened 2017-04-12
protect wakeup() from being called with 0

tburgin:
*  If wakeup() is called with 0 the machine will KP
*  I have been unable to reproduce a situation where the `.va_fsid` and `.va_fileid` would both be 0. Though I have had 2 or 3 KP reports in 6 months of exactly this. So let's prevent it.
*  Add branch prediction since this _should_ never happen.


tburgin:
CocoaPods 1.2.1 now requires XPC services to be embedded in their target :/
https://github.com/CocoaPods/CocoaPods/commit/16fae73efdc5ea0afae9f9ab49d9e323aa85b1dd

I have made this change and locally pods install and the targets link and compile successfully.
I think travis should pass now, but it keeps trying to link against `Pods-santabs` instead of the new `Pods-santad-santabs`.

Closing this pull, and going to re-open it, in hopes that travis will work?

PR #164, opened 2017-04-13
santa-driver: Refactor cache expiration calculation

msuozzo:
Cosmetic refactor of the cache expiration calculation in santa-driver.


tburgin:
Yay! We have swayed some random person to help! 👏

PR #167, opened 2017-04-14
protect wakeup() from being called with 0

tburgin:
*  Update CocoaPods to 1.2.1
*  If wakeup() is called with 0 the machine will KP
    *  I have been unable to reproduce a situation where the `.va_fsid` and `.va_fileid` would both be 0. Though I have had 2 or 3 KP reports in 6 months of exactly this. So let's prevent it.
    *  Add branch prediction since this _should_ never happen.


tburgin:
😢

tburgin:
Okay, should be good now. PTAL.

PR #173, opened 2017-05-22
project/config: Move /var/log/santa.log to /var/db/santa/santa.log

tburgin:
*  Move the default location of santa.log to mitigate against malware that wipes /var/log/*
  *  If /var/db/santa/* is wiped we will at least know it is targeting Santa


russellhancox:
Don't forget to call this out very clearly in the release notes.

tburgin:
Yes, this will be communicated in the release notes.

PR #175, opened 2017-06-02
santabs: use the ancestor bundle when searching for binaries

tburgin:
* Add useAncestorBundle property to SNTFileInfo. This gives us the benefit of keeping the API consistent for accessing bundle* and infoPlist methods, along with flexibility to switch which NSBundle backs those methods.

*  santabs will now look for the highest NSBundle in the path and use that to search for nested binaries.


Source/common/SNTFileInfo.m
```diff
@@ -285,23 +285,29 @@ - (BOOL)isMissingPageZero {
...
 ///  This method walks up the path until a bundle is found, if any.
 ///
+///  @param ancestor YES this will return the highest NSBundle found in the tree. No will return the
+///                  the lowest.
+///
+-(NSBundle *)findBundleWithAncestor:(BOOL)ancestor {
+  NSBundle *bundle;
+  NSMutableArray *pathComponents = [[self.path pathComponents] mutableCopy];
+
+  // Ignore the checking the root path "/". For some reason this produces a valid bundle.
```
russellhancox: nit, this reads weirdly. How about:
```
// Ignore the root path "/", for some reason this is considered a bundle.
```

Source/santabs/SNTBundleService.m
```diff
@@ -231,6 +236,7 @@ - (NSArray *)findRelatedBinaries:(SNTStoredEvent *)event progress:(NSProgress *)
       dispatch_group_async(group,
                            dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0), ^{
         @autoreleasepool {
+          fi.useAncestorBundle = YES;
```
russellhancox: I think there's an opportunity for an optimization here, particularly as we're making the bundle searching more time-consuming. Where you replace the bundle path on line 97, replace all the other bundle properties too, and then copy those to the new event below (lines 246-249).

Tests/LogicTests/SNTFileInfoTest.m
```diff
@@ -132,6 +132,65 @@ - (void)testBundle {
   XCTAssertEqualObjects([sut bundlePath], @"/Applications/Safari.app");
 }
 
+- (void)testAncestorBundle {
+  NSString *nestedBundleBinary =
+      @"/Applications/Safari.app/Contents/XPCServices/"
+      @"com.apple.Safari.BrowserDataImportingService.xpc/"
+      @"Contents/MacOS/com.apple.Safari.BrowserDataImportingService";
```
russellhancox: I realize a lot of these tests rely on system paths (which is bad anyway) but this one makes me nervous as I think it's likely to change in future releases. Could you create a bundle structure in the LogicTests/Resources folder and use that for this test?

Source/common/SNTFileInfo.m
tburgin: Done.

Source/santabs/SNTBundleService.m
tburgin: Good idea, done.

Tests/LogicTests/SNTFileInfoTest.m
tburgin: Could do, though testing the Ancestor Bundle idea wouldn't work because LogicTests.xctest the highest bundle is the path :/

tburgin: As discussed. Used LogicTests.xctest as the Ancestor Bundle.
PTAL.

PR #177, opened 2017-06-12
santabs: de-dupe generated events before upload

tburgin:


Source/santabs/SNTBundleService.m
```diff
@@ -25,6 +25,22 @@
 #import "SNTXPCConnection.h"
 #import "SNTXPCNotifierInterface.h"
 
+// For the SNTBundleService class, make SNTStoredEvent objects equal if the SHA256 matches.
+@implementation SNTStoredEvent (BundleEquatable)
```
russellhancox: I'm not sure about this. Quoth the [Apple documentation](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html):

> If the name of a method declared in a category is the same as a method in the original class, or a method in another category on the same class (or even a superclass), the behavior is undefined as to which method implementation is used at runtime.

The documentation used to be even stronger:

> Although the Objective-C language currently allows you to use a category to override methods the class inherits, or even methods declared in the class interface, you are strongly discouraged from doing so. A category is not a substitute for a subclass.

However, I'm not _exactly_ sure what a better solution looks like. 

- A subclass should work but then you're returning an array of the subclass rather than the stored event; this is fine unless the code using that array compares the objects in some way. 

- Another option would be to change relatedEvents on line 142 to a dictionary keyed by SHA-256 and only add to the dictionary if the key doesn't exist and then just return dict.allValues. You could avoid making the SNTStoredEvent objects at all this way instead of de-duping later.

Thoughts?

tburgin: Good catch on this, makes sense.

The subclass approach would require larger refactoring. XPC drops the replies since the collection member type doesn't match anymore.

I would like to keep this compare behavior just to santabs. The dictionary approach seems like a reasonable way forward.

tburgin:
PTAL, refactored it bit:
* `SNTFileInfo`: Added `initWithResolvedPath:error:` to avoid calling stat() on a file twice during bundle hashing.
* `SantaGUI`: Clean up bundle hashing artifacts on screen when santabs falls over.
* `santabs`: Use a C array to store pointers to SNTStoredEvent objects in favor of an NSMutableArray and locks. Use dispatch_apply() for optimized block scheduling.

Source/SantaGUI/SNTNotificationManager.m
```diff
@@ -192,12 +194,23 @@ - (void)setBundleServiceListener:(NSXPCListenerEndpoint *)listener {
...
+    }
+  };
+
   dispatch_semaphore_signal(self.bundleServiceSema);
 }
 
 #pragma mark SNTBundleNotifierXPC helper methods
 
 - (void)hashBundleBinariesForEvent:(SNTStoredEvent *)event {
+  self.currentWindowController.foundFileCountLabel.stringValue = @"Seaching for files...";
```
russellhancox: sp: Seaching

Source/common/SNTFileInfo.h
```diff
@@ -38,6 +38,18 @@
 ///
 - (instancetype)initWithPath:(NSString *)path;
 
+
+///
+///  Initializer for already resolved paths.
+///
+///  @param path The path of the file this instance is to represent. The path will be
```
russellhancox: remove the be at the end, it's repeated on the next line

Source/common/SNTFileInfo.m
```diff
@@ -59,11 +59,13 @@ @implementation SNTFileInfo
 
 extern NSString *const NSURLQuarantinePropertiesKey WEAK_IMPORT_ATTRIBUTE;
 
-- (instancetype)initWithPath:(NSString *)path error:(NSError **)error {
+- (instancetype)initWithPath:(NSString *)path
+                resolvedPath:(BOOL)resolvedPath
+                       error:(NSError **)error {
```
russellhancox: Instead of adding 2 new methods, why not just make this initWithResolvedPath:error and create a new initWIthPath:error: which just has the path resolving bit at the beginning?

Source/common/SNTFileInfo.m
```diff
@@ -105,10 +121,18 @@ - (instancetype)initWithPath:(NSString *)path error:(NSError **)error {
   return self;
 }
 
+- (instancetype)initWithPath:(NSString *)path error:(NSError **)error {
+  return [self initWithPath:path resolvedPath:NO error:NULL];
```
russellhancox: You should pass the error in..

Source/santabs/SNTBundleService.m
```diff
@@ -122,170 +130,145 @@ - (void)hashBundleBinariesForEvent:(SNTStoredEvent *)event
...
+  Find binaries within a bundle given the bundle's event. It will run until a timeout occurs,
+  or until the NSProgress is cancelled. Search is done within the bundle concurrently.
+
+  @param event The SNTStoredEvent to begin searching.
+  @return An NSDictionary object with keys of fileSHA256 and values of SNTStoredEvent objects.
+*/
+- (NSDictionary *)findRelatedBinaries:(SNTStoredEvent *)event progress:(NSProgress *)progress {
+  // Find all files and folders within the fileBundlePath
+  NSFileManager *fm = [NSFileManager defaultManager];
+  NSArray *dirEnum = [fm subpathsOfDirectoryAtPath:event.fileBundlePath error:NULL];
```
russellhancox: dirEnum no longer makes sense as a variable name

Source/santabs/SNTBundleService.m
```diff
@@ -122,170 +130,145 @@ - (void)hashBundleBinariesForEvent:(SNTStoredEvent *)event
...
         sentBinaryCount = binaryCount;
         [[self.notifierConnection remoteObjectProxy] updateCountsForEvent:event
                                                               binaryCount:binaryCount
-                                                                fileCount:fileCount];
+                                                                fileCount:i
+                                                              hashedCount:0];
       }
+
+      NSString *file = dirEnum[i];
+      if (!file) return;
```
russellhancox: Can this actually happen?

Source/santabs/SNTBundleService.m
```diff
@@ -122,170 +130,145 @@ - (void)hashBundleBinariesForEvent:(SNTStoredEvent *)event
...
+
+      NSString *newFile =
+          [event.fileBundlePath stringByAppendingPathComponent:file].stringByStandardizingPath;
+      SNTFileInfo *fi = [[SNTFileInfo alloc] initWithResolvedPath:newFile error:NULL];
+      if (!fi.isExecutable) return;
+
+      fis[i] = (void *)CFBridgingRetain(fi);
```
russellhancox: As this isn't a CF object that you're taking retained ownership of, I think I'd prefer a ```__bridge_retained``` instead. They're identical, it's just that CFBridgingRetain returns a ```CFTypeRef``` while the bridge can be done directly to void:

```objc
fis[i] = (__bridge_retained void *)fi
```

Source/santabs/SNTBundleService.m
```diff
@@ -122,170 +130,145 @@ - (void)hashBundleBinariesForEvent:(SNTStoredEvent *)event
...
+      OSAtomicIncrement64Barrier(&binaryCount);
     }
+  });
+
+  [progress resignCurrent];
+
+  NSMutableArray *fileInfos = [NSMutableArray array];
```
russellhancox: ```arrayWithCapacity:binaryCount```

Source/santabs/SNTBundleService.m
```diff
@@ -122,170 +130,145 @@ - (void)hashBundleBinariesForEvent:(SNTStoredEvent *)event
...
+  for (NSUInteger i = 0; i < dirEnum.count; i++) {
+    if (fis[i]) [fileInfos addObject:(SNTFileInfo *)CFBridgingRelease(fis[i])];
```
russellhancox: If you change ```CFBridgingRetain``` to ```__bridge_retained``` above, you'll also want to change ```CFBridgingRelease``` here to ```__bridge_transfer```

Source/santabs/SNTBundleService.m
```diff
@@ -122,170 +130,145 @@ - (void)hashBundleBinariesForEvent:(SNTStoredEvent *)event
...
-        binaryCount++;
-        pthread_mutex_unlock(&eventsMutex);
-
-        dispatch_semaphore_signal(sema);
+      dispatch_sync(dispatch_get_main_queue(), ^{
+        p.completedUnitCount++;
       });
-      if (progress && ((fileCount % 500) == 0 || binaryCount > sentBinaryCount)) {
+
+      if (progress && ((i % 500) == 0 || binaryCount > sentBinaryCount)) {
```
russellhancox: What's to stop two threads seeing that last condition at the same time?

Source/common/SNTFileInfo.m
tburgin: Done as discusses.

Source/santabs/SNTBundleService.m
tburgin: Nothing really, but it would just send an extra message to the gui. Just in case I moved this to the main thread.

Source/santabs/SNTBundleService.m
tburgin: Nope, removed.

Source/santabs/SNTBundleService.m
tburgin: Ah, nice. Done.

Source/santabs/SNTBundleService.m
tburgin: Done.

Source/santabs/SNTBundleService.m
tburgin: Done.

Source/common/SNTFileInfo.m
tburgin: :)

Source/common/SNTFileInfo.m
```diff
@@ -78,33 +74,17 @@ - (instancetype)initWithPath:(NSString *)path
       return nil;
```
russellhancox: I would change this error handling to be a bit clearer - this error suggests there was an error resolving the path but this is in the resolvedPath method, so this error should move to initWithPath (which will mean splitting out the path resolving from the init. This error should then call out the fact the path was empty.

tburgin: Makes sense. PTAL.

PR #185, opened 2017-07-31
santad: Modified log output format.

nguyen-phillip:
Moved path and args info to the end of the log output so that
extremely long path strings don't obscure process information.

### DIFF
--- a/Source/santad/SNTEventLog.m
+++ b/Source/santad/SNTEventLog.m
@@ -99,10 +99,8 @@
 
   // init the string with 2k capacity to avoid reallocs
   NSMutableString *outStr = [NSMutableString stringWithCapacity:2048];
-  [outStr appendFormat:@"action=%@|path=%@", action, [self sanitizeString:path]];
-  if (newpath) {
-    [outStr appendFormat:@"|newpath=%@", [self sanitizeString:newpath]];
-  }
+  [outStr appendFormat:@"action=%@", action];
+
   char ppath[PATH_MAX] = "(null)";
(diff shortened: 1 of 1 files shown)


Source/santad/SNTEventLog.m
```diff
@@ -99,17 +99,20 @@ - (void)logFileModification:(santa_message_t)message {
 
   // init the string with 2k capacity to avoid reallocs
   NSMutableString *outStr = [NSMutableString stringWithCapacity:2048];
-  [outStr appendFormat:@"action=%@|path=%@", action, [self sanitizeString:path]];
-  if (newpath) {
-    [outStr appendFormat:@"|newpath=%@", [self sanitizeString:newpath]];
-  }
+  [outStr appendFormat:@"action=%@", action];
```
russellhancox: Combine this with the larger appendFormat on line 107

Source/santad/SNTEventLog.m
```diff
@@ -212,6 +200,21 @@ - (void)logExecution:(santa_message_t)message withDecision:(SNTCachedDecision *)
                        message.gid, [self nameForGID:message.gid],
                        mode];
 
+  [outLog appendFormat:@"|sha256=%@|path=%@", cd.sha256, [self sanitizeString:@(message.path)]];
```
russellhancox: Combine this with the appendFormat call above

nguyen-phillip:
I restored logFileModification to its former state and rearranged the logExecution message so that path and args are at the very end.  Also combined adjacent appendFormats where able.

PR #188, opened 2017-08-02
common & santactl/sync: Added rule type for packages

nguyen-phillip:
Fixes problem where push notifications weren't being displayed after an
application bundle was whitelisted on the server.

### DIFF
--- a/Source/common/SNTCommonEnums.h
+++ b/Source/common/SNTCommonEnums.h
@@ -24,6 +24,7 @@ typedef NS_ENUM(NSInteger, SNTRuleType) {
 
   SNTRuleTypeBinary = 1,
   SNTRuleTypeCertificate = 2,
+  SNTRuleTypePackage = 3,
 };
 
 typedef NS_ENUM(NSInteger, SNTRuleState) {
--- a/Source/common/SNTCommonEnums.h
+++ b/Source/common/SNTCommonEnums.h
@@ -24,7 +24,6 @@ typedef NS_ENUM(NSInteger, SNTRuleType) {
 
   SNTRuleTypeBinary = 1,
   SNTRuleTypeCertificate = 2,
-  SNTRuleTypePackage = 3,
 };
 
 typedef NS_ENUM(NSInteger, SNTRuleState) {
--- a/Source/santactl/Commands/sync/SNTCommandSyncRuleDownload.m
+++ b/Source/santactl/Commands/sync/SNTCommandSyncRuleDownload.m
@@ -17,6 +17,7 @@
 #import "SNTCommandSyncConstants.h"
 #import "SNTCommandSyncState.h"
 #import "SNTRule.h"
+#import "SNTStoredEvent.h"
 #import "SNTXPCConnection.h"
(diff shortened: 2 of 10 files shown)


Source/common/SNTCommonEnums.h
```diff
@@ -24,6 +24,7 @@ typedef NS_ENUM(NSInteger, SNTRuleType) {
 
   SNTRuleTypeBinary = 1,
   SNTRuleTypeCertificate = 2,
+  SNTRuleTypePackage = 3,
```
tburgin: Since this new rule type is just used as an indication to show a notification we don't need this enum. The package rule will never be stored as a rule.

nguyen-phillip: Are the rules that are being added to the database being filtered in any way?

Source/santactl/Commands/sync/SNTCommandSyncRuleDownload.m
```diff
@@ -115,6 +115,8 @@ - (SNTRule *)ruleFromDictionary:(NSDictionary *)dict {
     newRule.type = SNTRuleTypeBinary;
   } else if ([ruleTypeString isEqual:kRuleTypeCertificate]) {
     newRule.type = SNTRuleTypeCertificate;
+  } else if ([ruleTypeString isEqual:kRuleTypePackage]) {
+    newRule.type = SNTRuleTypePackage;
```
tburgin: This should do the trick.
newRule.type = SNTRuleTypeUnknown;

Source/common/SNTCommonEnums.h
tburgin: Ah! Yeah, if the rule type is unknown it will be dropped.
https://github.com/google/santa/blob/master/Source/santad/DataLayer/SNTRuleTable.m#L165

Source/santactl/Commands/sync/SNTCommandSyncRuleDownload.m
tburgin: Ah crap, never mind. That would [choke](https://github.com/google/santa/blob/master/Source/santad/DataLayer/SNTRuleTable.m#L168) and stop processing the rules. We should filter package rules before L56 of SNTCommandSyncRuleDownload.m, then add back before L79.

nguyen-phillip:
I made a rough pass at how I think this should work, for the purpose of discussion.

Whenever a binary is blocked from executing, we store the event in a cache keyed by the binary's sha256 hash.  Whenever we download new rules, we check each new binary whitelist rule against the cache to see if the binary was recently blocked.  If so, we push out a notification to say it can now run.

The info in the FCM message triggering the rules sync is ignored because it doesn't necessarily correspond to the interesting rules being downloaded.

Probably a lot of issues still with the code, e.g. not sure if SNTExecutionController is the best place to be storing the cache.

tburgin:
Wow, I like your code 👍 
But, I was think of something a bit simpler. We should be able to do it without adding the SNTExecutionController critical path.
https://github.com/google/santa/compare/master...tburgin:bundle-notification

PR #193, opened 2017-08-15
Split kernel cache for root/non-root volume.

russellhancox:
Split the kernel cache into 2 separate caches, one for the root volume and one for secondary volumes. When an unmount happens, clear the non-root cache to ensure no overlap with filesystem IDs. Show the count for both caches in the santactl/status command.

Reduce the total cache size from **10k** entries to **5k (root) + 500 (non-root)**.

Also attempt to make logging slightly less bad in the kernel, using a gcc/clang varargs extension to avoid needing to call IOLog twice per output line.


tburgin:
You know what is needed

russellhancox:
PTAL.

russellhancox:
All kernel tests pass.

PR #195, opened 2017-08-16
santa-driver: Fix possible race condition in SDM::AddToCache

russellhancox:
SantaDecisionManager::AddToCache() rejects allow/deny entries being added to the
cache unless there's already an entry for that vnode with the REQUEST_BINARY
status, so that if a write to the file happens and the entry is removed while a
decision is being made, it's rejected.

However, as it's currently implemented there's a small window where the entry is
actually added to the cache. To fix that, this PR adds a way to use the
cache->set() method as a compare-and-swap.

While here, I also fixed the cache performance test in KernelTests so that it
measures walltime instead of CPU time.

Also fixed some line length violations.

### DIFF
--- a/Source/santa-driver/SantaCache.h
+++ b/Source/santa-driver/SantaCache.h
@@ -101,12 +101,19 @@ template<class T> class SantaCache {
   /**
     Set an element in the cache.
 
-    @note If the cache is full when this is called, this will empty the cache before
-    inserting the new value.
+    @note If the cache is full when this is called, this will
+          empty the cache before inserting the new value.
 
-    @return if an existing value was replaced, the previous value, otherwise zero_
+    @param key, The key
+    @param value, The value with parameterized type
+    @param previous_value, If the has_prev_value parameter is true the new
+        value will only be set if this parameter is equal to the provided value.
+        This allows set to become a CAS operation.
+    @param has_prev_value, Pass true if previous_value should be used.
+
+    @return the previous value (which may be zero_)
   */
-  T set(uint64_t key, T value) {
+  T set(uint64_t key, T value, T previous_value, bool has_prev_value) {
     struct bucket *bucket = &buckets_[hash(key)];
     lock(bucket);
     struct entry *entry = (struct entry *)((uintptr_t)bucket->head - 1);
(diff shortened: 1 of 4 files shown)


Source/santa-driver/SantaCache.h
```diff
@@ -166,6 +182,20 @@ template<class T> class SantaCache {
   }
 
   /**
+    Overload to allow setting without providing a previous value
+  */
+  T set(uint64_t key, T value) {
+    return set(key, value, {}, false);
```
tburgin: what is the benefit of using `{}` over `T()`?

Source/santa-driver/SantaCache.h
```diff
@@ -247,7 +277,7 @@ template<class T> class SantaCache {
   /**
     Holder for a 'zero' entry for the current type
   */
-  const T zero_ = T();
+  const T zero_ = {};
```
tburgin: Same ?

Source/santa-driver/SantaCache.h
russellhancox: That's C++11 list initialization, it has fewer limitations on provided types. It doesn't make much difference in our use but it's a nice switch.

Source/santa-driver/SantaCache.h
russellhancox: Same

Tests/KernelTests/main.mm
```diff
@@ -542,12 +545,12 @@ - (void)testCachePerformance {
   std::for_each(times.begin(), times.end(), [&](const double d) {
     accum += (d - mean) * (d - mean);
   });
-  double stdev = sqrt(accum / (times.size()-1));
+  double stdev = sqrt(accum / (times.size() - 1));
 
-  if (mean > 1000 || stdev > 150) {
-    TFAILINFO("μ: %-3.2f σ: %-3.2f", mean, stdev);
+  if (mean > 80 || stdev > 1) {
```
tburgin: 1 is a little low - especially when my machine is so much faster than yours 😝 
2 worked for me

russellhancox: I made it 10 to be safe.

PR #197, opened 2017-08-21
Bundle Notifications

nguyen-phillip:
Currently: FCM notification message format remains unchanged.  When rules are downloaded, rules about single binaries remain unchanged.  However for bundles, each whitelisting rule for a binary in a bundle includes two new fields: 

1. file_bundle_hash - identifying the bundle to which it belongs.
2. file_bundle_binary_count - which is the total number of binary rules associated with the bundle.

For bundles, a local whitelist notification is pushed to the user when all of the associated binary rules have been downloaded and added.

### DIFF
--- a/Source/santactl/Commands/sync/SNTCommandSyncManager.m
+++ b/Source/santactl/Commands/sync/SNTCommandSyncManager.m
@@ -37,7 +37,6 @@
 static NSString *const kFCMActionKey = @"action";
 static NSString *const kFCMFileHashKey = @"file_hash";
 static NSString *const kFCMFileNameKey = @"file_name";
-static NSString *const kFCMRuleCountKey = @"rule_count";
 static NSString *const kFCMTargetHostIDKey = @"target_host_id";
 
 @interface SNTCommandSyncManager () {
@@ -48,6 +47,9 @@ static NSString *const kFCMTargetHostIDKey = @"target_host_id";
 @property(nonatomic) dispatch_source_t ruleSyncTimer;
 
 @property(nonatomic) NSCache *dispatchLock;
+
+// Changed this to an NSMutableDictionary instead of NSCache so that we can iterate over its keys
+// when checking for bundle messages who've downloaded all of their associated binary rules.
 @property(nonatomic) NSMutableDictionary *pendingNotifications;
 
 @property NSUInteger FCMFullSyncInterval;
@@ -218,14 +220,17 @@ static void reachabilityHandler(
     return;
   }
 
-  // Store the file name and hash in a cache. When the rule is actually added, use the cache
-  // to build a user notification.
+  // Assumption: Incoming FCM message contains name of binary/bundle and a hash.  Rule count info
+  // for bundles will be sent out with the rules themselves.  If the message is related to a bundle,
+  // the hash is a bundle hash, otherwise it is just a hash for a single binary.
+  // For later use, we store a mapping of bundle/binary hash to a dictionary containing the
+  // binary/bundle name so we can send out relevant notifications once the rules are actually
+  // downloaded & added to local database.  We use a dictionary value so that we can later add a
+  // count field when we start downloading the rules and receive the count information.
   NSString *fileHash = message[kFCMFileHashKey];
   NSString *fileName = message[kFCMFileNameKey];
-  NSString *ruleCount = message[kFCMRuleCountKey];
-  if (fileName && fileHash && ruleCount) {
-    NSNumber *count = @([ruleCount integerValue]);
-    self.pendingNotifications[fileHash] = @{ @"name" : fileName, @"count" : count }.mutableCopy;
+  if (fileName && fileHash) {
+    self.pendingNotifications[fileHash] = @{ @"name" : fileName }.mutableCopy;
   }
 
   LOGD(@"Push notification action: %@ received", action);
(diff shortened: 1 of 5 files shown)


Source/santactl/Commands/sync/SNTCommandSyncRuleDownload.m
```diff
@@ -30,30 +30,16 @@ - (NSURL *)stageURL {
...
-
-  if (resp[kCursor]) {
-    return [self ruleDownloadWithCursor:resp[kCursor]];
-  }
-
-  if (!self.syncState.downloadedRules.count) return YES;
+  // Grab the new rules from server
+  NSArray<SNTRule *> *newRules = [self downloadNewRulesFromServer];
+  if (!newRules) return NO;         // encountered a problem with the download
+  if (!newRules.count) return YES;  // successfully downloaded rules, but nothing of interest
```
tburgin: Maybe something like:
successfully completed the request, but there are no new rules

nguyen-phillip: changed

Source/santactl/Commands/sync/SNTCommandSyncRuleDownload.m
```diff
@@ -124,6 +155,44 @@ - (SNTRule *)ruleFromDictionary:(NSDictionary *)dict {
...
+    // If the downloaded rule included count information, this initial value is (count - 1).
+    // If the downloaded rule had no count information, then it was a non-bundle rule and count is
+    // set to 0, indicating that the we've already downloaded all of the 1 rules associated with
+    // the binary.
+    [self.syncState.whitelistNotificationQueue addOperationWithBlock:^{
+      NSMutableDictionary *notifier = self.syncState.whitelistNotifications[primaryHash];
+      if (notifier) {
+        NSNumber *ruleCount = dict[kFileBundleBinaryCount];
+        NSNumber *notifierCount = notifier[kFileBundleBinaryCount];
+        if (notifierCount) {  // bundle rule with existing count
```
msuozzo: You could also write this something like (but more readable than) this:

```objc
if (!notifier[kFileBundleBinaryCount]) {
  notifier[kFileBundleBinaryCount] = ruleCount ? @[ruleCount intValue] : 1;
}
notifier[kFileBundleBinaryCount]--;
```

That way you'd avoid having both ruleCount and notifierCount floating around as variables.

Source/santactl/Commands/sync/SNTCommandSyncRuleDownload.m
```diff
@@ -124,6 +155,44 @@ - (SNTRule *)ruleFromDictionary:(NSDictionary *)dict {
...
+    // it is simply the binary hash.
+    NSString *primaryHash = dict[kFileBundleHash];
+    if (primaryHash.length != 64) {
+      primaryHash = newRule.shasum;
+    }
+
+    // If we have already seen a rule with the same primary hash, then decrement the count of the
+    // corresponding pending notification.  Otherwise, if this is the first time we've seen this
+    // primary hash, add a count field to the pending notfication and set its initial value.
+    // If the downloaded rule included count information, this initial value is (count - 1).
```
msuozzo: this sentence kinda reiterates the code. you might include a point about _why_ you set it to `(count-1)`.

Source/santactl/Commands/sync/SNTCommandSyncManager.m
```diff
@@ -47,7 +47,11 @@ @interface SNTCommandSyncManager () {
 @property(nonatomic) dispatch_source_t ruleSyncTimer;
 
 @property(nonatomic) NSCache *dispatchLock;
-@property(nonatomic) NSCache *ruleSyncCache;
+
+// Changed this to an NSMutableDictionary instead of NSCache so that we can iterate over its keys
```
nguyen-phillip: Sorry, I was using a lot of these comments as temporary documentation.  I'll go ahead and fix them.  Let me know if they still seem weird / too verbose.

Source/santactl/Commands/sync/SNTCommandSyncManager.m
```diff
@@ -47,7 +47,11 @@ @interface SNTCommandSyncManager () {
...
+// when checking for bundle messages who've downloaded all of their associated binary rules.
+@property(nonatomic) NSMutableDictionary *whitelistNotifications;
+@property(nonatomic) NSOperationQueue *whitelistNotificationQeue;
```
nguyen-phillip: typo + autocomplete.  fixed.

Source/santactl/Commands/sync/SNTCommandSyncRuleDownload.m
nguyen-phillip: rewrote, but still might be too wordy

Source/santactl/Commands/sync/SNTCommandSyncRuleDownload.m
nguyen-phillip: This is what that would look like:
```
        if (!notifier[kFileBundleBinaryCount]) {
            notifier[kFileBundleBinaryCount] = ruleCount ?: @1;
        }
        notifier[kFileBundleBinaryCount] = @([notifier[kFileBundleBinaryCount] intValue] - 1);
```
which is definitely terser, but seems a bit more confusing to me.  I think maybe the variable notifierCount just needs a better name, say notifierRemaining or notifierLeft to distinguish it from ruleCount which is conceptually different.

PR #202, opened 2017-09-11
sync: start reachability handler to retry to upload blocked event when initial upload fails

nguyen-phillip:


nguyen-phillip:
PTAL

nguyen-phillip:
PTAL

Source/santactl/Commands/sync/SNTCommandSyncManager.m
```diff
@@ -146,22 +145,21 @@ - (void)postEventsToSyncServer:(NSArray<SNTStoredEvent *> *)events isFromBundle:
...
   SNTCommandSyncEventUpload *p = [[SNTCommandSyncEventUpload alloc] initWithState:syncState];
-  if (event && [p uploadEvents:@[event]]) {
+  if ([p uploadEvents:@[event]]) {
     if ([syncState.bundleBinaryRequests containsObject:event.fileBundleHash]) {
-      reply(SNTBundleEventActionSendEvents);
+      reply(YES); // store and immediately send related bundle events
       LOGD(@"Needs related events");
     } else {
-      reply(SNTBundleEventActionDropEvents);
       LOGD(@"Bundle event upload complete");
```
tburgin: Now this case doesn't reply back to santad. I am pretty sure santad will keep around the block and captured objects forever. It may also prevent subsequent calls to `postBundleEventToSyncServer:reply:` from being sent since XPC operates serially. Testing now...

tburgin: Objects that are captured within a reply block that is never executed do "leak".

Client:
```objc
  NSXPCConnection *c =
      [[NSXPCConnection alloc] initWithServiceName:@"com.google.corp.XPCTestService"];
  c.remoteObjectInterface =
      [NSXPCInterface interfaceWithProtocol:@protocol(XPCTestServiceProtocol)];
  [c resume];

  NSArray *hudge = [self arrayWithNumberOfStrings:1000000];

  [[c remoteObjectProxy] sortArray:[self arrayWithNumberOfStrings:42] withReply:^(NSArray *a) {
    NSLog(@"hudge %lu", hudge.count);
    NSLog(@"%lu", a.count);
  }];
```

Service:
```objc
- (void)sortArray:(NSArray *)array withReply:(void (^)(NSArray *))reply {
  NSLog(@"%@ called with count %lu", NSStringFromSelector(_cmd), array.count);
  if (array.count == 42) {
    reply([array sortedArrayUsingSelector:@selector(localizedCaseInsensitiveCompare:)]);
  }
}
```

The first run is with a reply block that fires, the second is one that does not fire.

screen shot 2017-09-13 at 4 23 32 pm

To my second point. Since the method you are calling does return, it does not block subsequent calls.

In summery we still need the enum for the reply.

nguyen-phillip: Oh well, I guess it was too good to be true

tburgin: I like the enum approach, it is explicit. Which is nice for when we re-read this code 1 year from now 😉

nguyen-phillip:
PTAL

Source/santad/SNTSyncdQueue.m
```diff
@@ -49,16 +49,18 @@ - (void)addEvents:(NSArray<SNTStoredEvent *> *)events isFromBundle:(BOOL)isFromB
...
   [self dispatchBlockOnSyncdQueue:^{
-    [self.syncdConnection.remoteObjectProxy postBundleEventToSyncServer:event
-                                                                  reply:^(BOOL needRelatedEvents) {
-      // Remove the backoff entry for the inital block event. The same event will be included in the
-      // related events synced using addEvents:isFromBundle:.
-      if (needRelatedEvents) [self.uploadBackoff removeObjectForKey:event.fileBundleHash];
-      reply(needRelatedEvents);
-    }];
+    [self.syncdConnection.remoteObjectProxy
+     postBundleEventToSyncServer:event reply:^(SNTBundleEventAction action) {
```
tburgin: Indent +3

nguyen-phillip:
PTAL

Source/santad/SNTSyncdQueue.m
```diff
@@ -53,14 +53,14 @@ - (void)addBundleEvent:(SNTStoredEvent *)event reply:(void (^)(SNTBundleEventAct
...
     [self.syncdConnection.remoteObjectProxy
-     postBundleEventToSyncServer:event reply:^(SNTBundleEventAction action) {
-       // Remove the backoff entry for the inital block event. The same event will be included in
-       // the related events synced using addEvents:isFromBundle:.
-       if (action == SNTBundleEventActionSendEvents) {
-         [self.uploadBackoff removeObjectForKey:event.fileBundleHash];
-       }
-       reply(action);
-     }];
+        postBundleEventToSyncServer:event reply:^(SNTBundleEventAction action) {
```
tburgin: This is good now.

Source/santad/SNTSyncdQueue.m
```diff
@@ -53,14 +53,14 @@ - (void)addBundleEvent:(SNTStoredEvent *)event reply:(void (^)(SNTBundleEventAct
...
+          // Remove the backoff entry for the inital block event. The same event will be included in
```
tburgin: The rest of the lines only need to be indented +2 from L55

nguyen-phillip:
PTAL

PR #218, opened 2017-11-15
santa-driver: Prevent possible infinite loop if decision requests fai…

### LINKED ISSUE #215: Checking IOSharedDataQueue available space before enqueuing new items
Hi, 

According to the documentation of `IOSharedDataQueue` in `enqueue` method it says : 

"When attempting to add a new entry towards the end of the queue and there isn't enough space at the end, it wraps back to the beginning"

In Santa, I couldn't see any restriction for enqueueing new item to the decision_dataqueue_ , and I wonder if the design was to accept loosing older decision requests for the sake of new ones ? 

thanks

russellhancox:
…l to be retrieved

When enqueue'ing on the decision data queue, if the queue is full the new message will overwrite the oldest. In this scenario it's possible for that overwritten request to get stuck in an infinite loop - as far as the driver is concerned there's a request pending that the driver should be picking up and responding to but the daemon has never actually received the request. The only way out of this loop is for the file being executed to be written to. This change adds an expiration to pending requests (of 5s) so that if this scenario were to happen the pending request would be removed, breaking out of the inner decision loop to the outer loop where the request is sent to the daemon.

This change also removes a pointless dequeue in the log queue, it was intended to try and help reduce the queue size to get logs flowing again but it doesn't really help.

Fixes #215

### DIFF
--- a/Source/santa-driver/SantaDecisionManager.h
+++ b/Source/santa-driver/SantaDecisionManager.h
@@ -79,8 +79,9 @@ class SantaDecisionManager : public OSObject {
   kern_return_t StartListener();
 
   /**
-    Stops the kauth listeners. After stopping new callback requests, waits until all
-    current invocations have finished before clearing the cache and returning.
+    Stops the kauth listeners. After stopping new callback requests, waits 
+    until all current invocations have finished before clearing the cache and
+    returning.
   */
   kern_return_t StopListener();
 
@@ -89,7 +90,10 @@ class SantaDecisionManager : public OSObject {
                   const santa_action_t decision,
                   const uint64_t microsecs = GetCurrentUptime());
 
-  ///  Fetches a response from the cache, first checking to see if the entry has expired.
+  /**
+    Fetches a response from the cache, first checking to see if the entry 
+    has expired.
+  */
   santa_action_t GetFromCache(uint64_t identifier);
 
   /// Checks to see if a given identifier is in the cache and removes it.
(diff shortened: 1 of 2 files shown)


Source/santa-driver/SantaDecisionManager.h
```diff
@@ -137,19 +144,34 @@ class SantaDecisionManager : public OSObject {
...
+
   /// Maximum number of entries in the in-kernel cache.
   static const uint32_t kMaxCacheSize = 10000;
 
   /// Maximum number of PostToDecisionQueue failures to allow.
   static const uint32_t kMaxDecisionQueueFailures = 10;
 
-  /// The maximum number of messages can be kept in the decision data queue at any time.
+  /**
+    The maximum number of messages can be kept in the decision data queue at
```
nguyen-phillip: nit: ... messages that/which can be kept ...

Source/santa-driver/SantaDecisionManager.h
```diff
@@ -137,19 +144,34 @@ class SantaDecisionManager : public OSObject {
...
+    any time.
+  */
   static const uint32_t kMaxDecisionQueueEvents = 512;
 
-  /// The maximum number of messages can be kept in the logging data queue at any time.
+  /**
+    The maximum number of messages can be kept in the logging data queue at
```
nguyen-phillip: nit: ... messages that can be kept ...

PR #219, opened 2017-11-21
santa-driver: Add timestamp support for ACTION_REQUEST_BINARY

tburgin:
https://github.com/google/santa/commit/133814cd733dce84d30a71aad336b3739d7953bc introduced a hanging issue when making decisions are larger executables. This stems from prematurely evicting ACTION_REQUEST_BINARY entries from the kernel cache, causing the decision queue to fill up with re-requests.

To be able to evict stale ACTION_REQUEST_BINARY requests we need to:
  *  store a timestamp at request time
  *  Send the timestamp along with the request to user-space and receive the timestamp along with the decision back in the kernel-space
  *  Use the timestamp along with ACTION_REQUEST_BINARY as the expected CAS value before updating the decision.

### DIFF
--- a/Source/common/SNTKernelCommon.h
+++ b/Source/common/SNTKernelCommon.h
@@ -82,6 +82,7 @@ typedef enum {
 // Message struct that is sent down the IODataQueue.
 typedef struct {
   santa_action_t action;
+  uint64_t timestamp;
   uint64_t vnode_id;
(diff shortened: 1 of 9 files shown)


tburgin:
Closing in favor of https://github.com/google/santa/pull/220

Issue #209, opened 2017-10-11
Add support for configuring Santa via mobileconfig


jesseendahl:
Hi!

I think it'd be a really great addition if Santa could read config from a `.mobileconfig`. There are several advantages to supporting this functionality, but I'll just mention the biggest one. The biggest benefit to Google and the Mac admin community is that it would make it possible to easily deploy Santa using only MDM, without the need for any other configuration management tooling (e.g. Puppet/Chef) or custom packages.

This would already be hugely beneficial today, but would also serve to future proof Santa for a potential world where the only way to manage/configure apps on macOS is via MDM.

Thoughts on implementation:

- Santa would need to ignore any existing configuration on disk once a `.mobileconfig` got installed.
- Currently, when operating in sync server mode, `santad` won't read configuration changes without an unload/reload of its launch daemon. This would have to be changed so that `santad` would new config from a `.mobileconfig` automatically, without an unload/reload.

russellhancox:
I think this is great idea and would definitely make managing Santa easier for everyone.

We can use NSUserDefaults' `objectIsForcedForKey:` and `objectIsForcedForKey:inDomain:` to only allow this configuration from a .mobileconfig rather than from the user themselves.

All of the changes should be limited to the SNTConfigurator class, which should make this relatively easy to implement. If you (or anyone) have time to work on this we'd be happy to review contributions, otherwise we'll get to this in the near future.

jesseendahl:
Awesome! Glad you agree :) Unfortunately we don't have the bandwidth to implement it ourselves  right now, but super excited to hear you can make it happen on your end soon(ish).

tburgin:
`objectIsForcedForKey:` and `objectIsForcedForKey:inDomain:` doesn't seem to be able to differentiate between mobile config payload scopes [0]. We would only want to pay attention to the System scope. Allowing users to flip config values would not be good.

We could use `CFPreferencesCopyValue()` to specify the System scope, but that will not tell us if the value is forced or not :/

In light of this, I think CFPreferencesCopyValue is the best way forward . This way we at least know if the config was written by an admin.

Thoughts?

[0] 
```
<key>PayloadScope</key>
<string>User</string>
```
```
<key>PayloadScope</key>
<string>System</string>
```

jesseendahl:
I talked this over with @mikeymike and yes  `CFPreferencesCopyValue` seems like the way to go.

[...]

PR #222, opened 2017-12-01
SNTConfigurator: use mobileconfigs

tburgin:
*  `/var/db/santa/config.plist` is now ignored and not used
*  `/var/db/santa/sync-state.plist` now holds state from the sync server
*  `com.google.santa` mobileconfig is now used for configuration
*  If a mobileconfig manages the `SyncBaseURL` the ` kClientModeKey` `kWhitelistRegexKey` `kBlacklistRegexKey` `kFileChangesRegexKey` keys will be ignored within the mobile config.

Fixes https://github.com/google/santa/issues/209

~~One note: The sync-state.plist could include values from a mobile config if the sync server is not setting them. This "effective" sync-state will only make a difference if the mobile config changed values while santad was not running and the sync server was not managing those values.~~

### DIFF
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -359,7 +359,6 @@ static NSString *const kSyncCleanRequired = @"SyncCleanRequired";
                                               error:&error];
   if (!readData) {
     LOGE(@"Could not read sync state file: %@, replacing.", [error localizedDescription]);
-    [self saveSyncStateToDisk];
     return;
   }
 
@@ -370,7 +369,6 @@ static NSString *const kSyncCleanRequired = @"SyncCleanRequired";
                                                   error:&error];
   if (!syncState) {
     LOGE(@"Could not parse sync state file: %@, replacing.", [error localizedDescription]);
-    [self saveSyncStateToDisk];
     return;
   }
 
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -27,9 +27,6 @@
 
 /// A NSUserDefaults object set to use the com.google.santa suite.
 @property(readonly, nonatomic) NSUserDefaults *defaults;
-
-/// Array of keys that cannot be changed while santad is running if santad didn't make the change.
-@property(readonly) NSArray *protectedKeys;
 @end
 
 @implementation SNTConfigurator
@@ -100,7 +97,7 @@ static NSString *const kSyncCleanRequired = @"SyncCleanRequired";
   return sharedConfigurator;
 }
 
-#pragma mark Protected Keys
+#pragma mark Keys
 
 - (NSArray *)syncServerKeys {
   static NSArray *keys;
@@ -430,7 +427,7 @@ static NSString *const kSyncCleanRequired = @"SyncCleanRequired";
     }
   }
   if (config[kSyncBaseURLKey] || self.syncBaseURL) {
-    for (NSString *key in [self protectedKeys]) {
+    for (NSString *key in [self syncServerKeys]) {
       if ([key isEqualToString:kSyncBaseURLKey]) continue;
       [config removeObjectForKey:key];
     }
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -426,7 +426,7 @@ static NSString *const kSyncCleanRequired = @"SyncCleanRequired";
       config[key] = [self.defaults objectForKey:key];
     }
   }
-  if (config[kSyncBaseURLKey] || self.syncBaseURL) {
+  if (config[kSyncBaseURLKey]) {
     for (NSString *key in [self syncServerKeys]) {
       if ([key isEqualToString:kSyncBaseURLKey]) continue;
       [config removeObjectForKey:key];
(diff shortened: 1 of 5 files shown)


Source/common/SNTConfigurator.m
```diff
@@ -82,19 +89,44 @@ - (instancetype)initWithFilePath:(NSString *)filePath {
...
   return sharedConfigurator;
 }
 
-#pragma mark Protected Keys
+#pragma mark Keys
 
-- (NSArray *)protectedKeys {
-  return @[ kClientModeKey, kWhitelistRegexKey, kBlacklistRegexKey,
-            kFileChangesRegexKey, kSyncBaseURLKey ];
+- (NSArray *)syncServerKeys {
```
nguyen-phillip: This could be declared as a readonly property and initialized in init.

Source/common/SNTConfigurator.m
```diff
@@ -82,19 +89,44 @@ - (instancetype)initWithFilePath:(NSString *)filePath {
...
+  dispatch_once(&onceToken, ^{
+    keys = @[
+        kClientModeKey, kWhitelistRegexKey, kBlacklistRegexKey, kFileChangesRegexKey,
+        kFullSyncLastSuccess, kRuleSyncLastSuccess, kSyncCleanRequired
+    ];
+  });
+  return keys;
+}
+
+- (NSArray *)mobileConfigKeys {
```
nguyen-phillip: Same as for syncServerKeys, this could be a readonly property.

Source/common/SNTConfigurator.m
```diff
@@ -322,70 +354,85 @@ - (NSString *)machineID {
...
-    [self saveConfigToDisk];
-    return;
-  };
+  }
+
+  // Nothing else to do if a sync server is not involed
+  if (!self.configData[kSyncBaseURLKey]) return;
+
+  // Load the last known sync state
+  NSMutableDictionary *syncState =
```
nguyen-phillip: It looks like the syncState variable is always reassigned in line 393 before it is ever used, so you could just get rid of it and change line 397 to read directly from syncStateData.

Source/common/SNTConfigurator.m
```diff
@@ -322,70 +354,85 @@ - (NSString *)machineID {
...
+      [NSMutableDictionary dictionaryWithCapacity:[self syncServerKeys].count];
+  if (![[NSFileManager defaultManager] fileExistsAtPath:kSyncStateFilePath]) return;
```
nguyen-phillip: Just to be clear: if the syncStateFile doesn't exist, we don't want to call saveSyncStateToDisk?

Source/common/SNTConfigurator.m
```diff
@@ -322,70 +354,85 @@ - (NSString *)machineID {
...
+  self.configData = [self mobileConfig];
+  if (!self.configData[kClientModeKey]) {
+    // Default to Monitor if the config is missing or invalid
     self.configData[kClientModeKey] = @(SNTClientModeMonitor);
-    [self saveConfigToDisk];
-    return;
-  };
+  }
+
+  // Nothing else to do if a sync server is not involed
```
nguyen-phillip: typo: involved

Source/common/SNTConfigurator.m
```diff
@@ -322,70 +354,85 @@ - (NSString *)machineID {
...
+      [NSMutableDictionary dictionaryWithCapacity:[self syncServerKeys].count];
+  for (NSString *key in [self syncServerKeys]) {
+    syncState[key] = self.configData[key];
+  }
+  [syncState writeToFile:kSyncStateFilePath atomically:YES];
+}
+
+
+///
+///  Returns a config provided by a com.google.santa mobileconfig or nil if
```
nguyen-phillip: This method never returns nil.

Source/common/SNTConfigurator.m
tburgin: YES

Source/common/SNTConfigurator.m
tburgin: Yes, and actually we don't need to write to disk here at all. The sync-state should only be updated when a sync happens causing  santactl to send an XPC message to santad to update the file.

jessfs:
@russellhancox how's the testing going?

Source/common/SNTConfigurator.m
```diff
@@ -322,70 +340,79 @@ - (NSString *)machineID {
...
-- (void)saveConfigToDisk {
+- (void)saveSyncStateToDisk {
   if (geteuid() != 0) return;
-  [self.configData writeToFile:self.configFilePath atomically:YES];
+  NSMutableDictionary *syncState =
+      [NSMutableDictionary dictionaryWithCapacity:[self syncServerKeys].count];
+  for (NSString *key in [self syncServerKeys]) {
+    syncState[key] = self.configData[key];
+  }
+  [syncState writeToFile:kSyncStateFilePath atomically:YES];
```
russellhancox: Probably want to ensure here that the permissions aren't broken.

russellhancox: It's probably worth watching the sync state file for changes and re-writing the file if it changes behind our back. santad should be the only process writing to the file.

jesseendahl:
Happy new year!! 🎉😀 How is this going? Is there a bit more work to be done to take into account @russellhancox last comment?

tburgin:
Yes, a bit more :) I will jump on it as soon as I get a moment.

tburgin:
PTAL

PR #234, opened 2018-02-02
 config: use KVO

tburgin:
Whenever there is a system wide mobileconfig change, all on-disk representations are deleted and re-created...twice (Thanks Apple). This flapping causes Santa to reload its config many times over a short period of time. In some cases this can result in an incorrect running config.

*  Build and merge the new config in temporary variables.
*  Let the caller of `reloadConfigData` know if a reload is taking place.
*  Update callers to not do unnecessary work.

### DIFF
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -31,7 +31,7 @@
 @property(readonly, nonatomic) NSUserDefaults *defaults;
 
 /// Holds the configuration from a sync server.
-@property(nonatomic, readonly) NSMutableDictionary *syncState;
+@property NSMutableDictionary *syncState;
 
 /// Used to determine if the underlying values have changed.
 @property SNTClientMode clientModeLastSeen;
@@ -122,7 +122,9 @@ static NSString *const kSyncCleanRequired = @"SyncCleanRequired";
     return cm;
   }
 
-  LOGE(@"Client mode was set to bad value: %ld. Defaulting to MONITOR.", cm);
+  if (self.clientModeLastSeen != cm) {
+    LOGE(@"Client mode was set to bad value: %ld. Defaulting to MONITOR.", cm);
+  }
   if (update) self.clientModeLastSeen = cm;
   return SNTClientModeMonitor;
 }
@@ -337,6 +339,8 @@ static NSString *const kSyncCleanRequired = @"SyncCleanRequired";
 ///  Read the saved syncState.
 ///
 - (NSMutableDictionary *)readSyncStateFromDisk {
+  // Only read the sync state if a sync server is configured.
+  if (!self.syncBaseURL) return nil;
   NSMutableDictionary *syncState =
       [NSMutableDictionary dictionaryWithContentsOfFile:kSyncStateFilePath];
   if (!syncState) LOGE(@"Could not read sync state file: %@", kSyncStateFilePath);
@@ -448,6 +452,7 @@ static NSString *const kSyncCleanRequired = @"SyncCleanRequired";
   NSURL *newSyncBaseURL = [self syncBaseURLAndUpdateLastSeen:NO];
   if (![newSyncBaseURL.absoluteString isEqualToString:self.syncBaseURLLastSeen.absoluteString]) {
     self.syncBaseURLLastSeen = newSyncBaseURL;
+    self.syncState = [self readSyncStateFromDisk];
     if ([self.delegate conformsToProtocol:@protocol(SNTConfiguratorReceiver)]) {
       [self.delegate syncBaseURLDidChange:newSyncBaseURL];
     }
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -22,20 +22,16 @@
 #import "SNTSystemInfo.h"
 
 @interface SNTConfigurator ()
-/// Creating NSRegularExpression objects is not fast, so cache them.
-@property NSRegularExpression *cachedFileChangesRegex;
-@property NSRegularExpression *cachedWhitelistDirRegex;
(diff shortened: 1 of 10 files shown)


Source/common/SNTConfigurator.m
```diff
@@ -101,7 +103,8 @@ - (instancetype)init {
         kMachineOwnerKey, kMachineIDKey, kMachineOwnerPlistFileKey, kMachineOwnerPlistKeyKey,
         kMachineIDPlistFileKey, kMachineIDPlistKeyKey
     ];
-    [self reloadConfigData];
+    _reloading = 0;
+    [self reloadConfigDataHelper];
```
russellhancox: This should call reloadConfigData so the barrier can be alive even during init.

tburgin: I avoided doing that because of the 5 sec blocking time. Every component of Santa initializes an instance of SNTConfigurator very early on and will block until complete:

[2018-02-02T00:52:41.882Z] I santad: Started, version 0.9.21
[2018-02-02T00:52:47.118Z] I santad: Connected to driver, activating.

⇒  date && santactl status && date
Thu Feb  1 19:57:05 EST 2018
>>> Daemon Info
  Mode                      | Lockdown
  File Logging              | Yes
  Watchdog CPU Events       | 0  (Peak: 0.22%)
  Watchdog RAM Events       | 0  (Peak: 12.68MB)
...
Thu Feb  1 19:57:10 EST 2018

russellhancox: GitHub didn't save my updated comment - yeah the 5s sleep needs to be avoided but you still need the barrier. Either do the atomic stuff around the call in init or make the sleep length a parameter

tburgin: ack

tburgin:
I ran this fix on my much slower MacBook Pro and the lag time is upwards of 10+ seconds before the NSUserDefaults values settle. So the 5 sec sleep will not fix all cases. We should investigate other options for mobileconfig updates.

A trivial example of how terrible NSUserDefaults is:
```objc

int main(int argc, char *argv[]) {
	@autoreleasepool {
		NSUserDefaults *d = [[NSUserDefaults alloc] initWithSuiteName:@"com.google.santa"];
		NSString *key = @"SyncBaseURL";
		while (1) {
			if ([d objectIsForcedForKey:key]) {
				if (![d objectForKey:key]) {
					NSLog(@"enforced miss");
				}
			} else {
				NSLog(@"miss");
			} 
		}
	}
}
```

A com.google.santa profile is installed with a SyncBaseURL set.
Here is output when removing an unrelated profile:
```
⇒  sudo ~/Desktop/cas
2018-02-01 22:41:05.821 cas[37780:9409158] miss
<snip>
2018-02-01 22:41:14.727 cas[37780:9409158] miss
2018-02-01 22:41:15.343 cas[37780:9409158] enforced miss
2018-02-01 22:41:15.347 cas[37780:9409158] miss
<snip>
2018-02-01 22:41:15.369 cas[37780:9409158] miss
```

tburgin:
@russellhancox figured out a use for `NSUserDefaultsDidChangeNotification` that works santad's context. I will update this pull next week with those changes.

tburgin:
PTAL

*  The mobileconfig is now queried from the getters. No more watching files for changes.
*  The sync state is held in memory and written to disk anytime it is changed.
*  syncd will now start and stop when the config changes

tburgin:
PTAL
*  SNTApplication will now respond to config changes
*  All config properties now support KVO

Source/santad/SNTApplication.m
```diff
@@ -73,18 +70,37 @@ - (instancetype)init {
...
+    self.notQueue = [[SNTNotificationQueue alloc] init];
     SNTSyncdQueue *syncdQueue = [[SNTSyncdQueue alloc] init];
 
     // Restart santactl if it goes down
     syncdQueue.invalidationHandler = ^{
       [self startSyncd];
     };
-    
+
+    // Listen for clientMode and syncBaseURL changes
```
russellhancox: Comment is incorrect

Source/santad/SNTApplication.m
```diff
@@ -270,4 +242,77 @@ void diskDisappearedCallback(DADiskRef disk, void *context) {
...
+  int ret = kill(self.syncdPID, SIGKILL);
+  LOGD(@"kill(%i, 9) = %i", self.syncdPID, ret);
+  self.syncdPID = 0;
+}
+
+- (void)observeValueForKeyPath:(NSString *)keyPath
+                      ofObject:(id)object
+                        change:(NSDictionary<NSString *,id> *)change
+                       context:(void *)context {
+  LOGD(@"%@ --> %@", keyPath, change);
```
russellhancox: Needed?

Source/common/SNTConfigurator.m
```diff
@@ -441,25 +489,58 @@ - (void)syncStateFileChanged:(unsigned long)data {
...
+
+- (NSMutableDictionary *)readForcedConfig {
+  NSMutableDictionary *forcedConfig = [NSMutableDictionary dictionary];
+  for (NSString *key in self.forcedConfigKeyTypes) {
+    id obj = [self forcedConfigValueForKey:key];
+    forcedConfig[key] = [obj isKindOfClass:self.forcedConfigKeyTypes[key]] ? obj : nil;
+    // Create the regex objects now
+    if ([key isEqualToString:kWhitelistRegexKey] ||
+        [key isEqualToString:kBlacklistRegexKey] ||
+        [key isEqualToString:kFileChangesRegexKey]) {
```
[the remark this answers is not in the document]
tburgin: Good idea, done throughout.

PR #242, opened 2018-02-26
sync-state: perform sync-state operations on a serial q

tburgin:
The sync-state is read once at santad startup. Continually watching the file and reverting the changes provides no real advantages.

With this change: If someone modifies the sync-state file while santad is running nothing happens. The next sync will overwrite the sync-state file with proper content and permissions. Still the sync-state file is not read from until santad restarts.

The only detractor is if someone modifies the sync-state file then santad restarts before a sync takes place. This is really no different to someone modifying the sync-state file while santad is not running. Again, the next sync will overwrite any out-of-band changes to the sync state in santad and on disk.

The motivation for this change is to avoid adding more complexity to the file watcher code to solve:

```
Application Specific Information:
*** Terminating app due to uncaught exception 'NSGenericException', reason: '*** Collection <__NSDictionaryM: 0x7fd658c1cfd0> was mutated while being enumerated.'
abort() called
terminating with uncaught exception of type NSException

Application Specific Backtrace 1:
0   CoreFoundation                      0x00007fff5756333b __exceptionPreprocess + 171
1   libobjc.A.dylib                     0x00007fff7e2a7942 objc_exception_throw + 48
2   CoreFoundation                      0x00007fff575f2d8a __NSFastEnumerationMutationHandler + 298
3   santad                              0x000000010348787f -[SNTConfigurator syncStateFileChanged:] + 303
4   santad                              0x0000000103488659 __40-[SNTConfigurator readSyncStateFromDisk]_block_invoke_2 + 41
5   santad                              0x000000010347dda9 __35-[SNTFileWatcher startWatchingFile]_block_invoke.19 + 57
6   libdispatch.dylib                   0x00007fff7ee62591 _dispatch_call_block_and_release + 12
7   libdispatch.dylib                   0x00007fff7ee5ad50 _dispatch_client_callout + 8
8   libdispatch.dylib                   0x00007fff7ee6de76 _dispatch_continuation_pop + 472
9   libdispatch.dylib                   0x00007fff7ee5d4bf _dispatch_source_invoke + 1811
10  libdispatch.dylib                   0x00007fff7ee5c941 _dispatch_root_queue_drain + 515
11  libdispatch.dylib                   0x00007fff7ee5c6ed _dispatch_worker_thread3 + 101
12  libsystem_pthread.dylib             0x00007fff7f11f1ca _pthread_wqthread + 1387
13  libsystem_pthread.dylib             0x00007fff7f11ec4d start_wqthread + 13
```


russellhancox:
A better fix would be to make `-syncStateFileChanged` do its work on a serial queue.

tburgin:
I figured you would say that, and yet I am fresh out of comebacks. I will make it so.

tburgin:
PTAL - syncState changes and reloads are now done serially

tburgin:
PTAL - "I'll enter the same calculations using what we like to call The Right Way."

PR #246, opened 2018-03-27
santad: event logger

tburgin:
* Create a new `EventLogType` option. This will allow santad to send event logs to syslog (ASL / ULS) or directly to a file.
* Enables building with 10.12+ SDKs without loosing the event log.


Source/common/SNTConfigurator.m
```diff
@@ -429,6 +442,16 @@ - (NSString *)machineID {
   return machineId.length ? machineId : [SNTSystemInfo hardwareUUID];
 }
 
+- (NSString *)eventLogType {
+  NSString *s = self.configState[kEventLogType];
+  if ([s isEqualToString:@"Syslog"] || [s isEqualToString:@"FileLog"]) return s;
```
russellhancox: make these comparisons case insensitive

Source/common/SNTConfigurator.h
```diff
@@ -79,6 +79,23 @@
...
 @property(readonly, nonatomic) BOOL enablePageZeroProtection;
 
+///
+///  Defines how event logs are stored. Options are:
+///    Syslog: Sent to ASL or ULS (if built with the 10.12 SDK or later).
+///    Filelog: Sent to a file on disk. Use eventLogPath to specify a path.
+///
+///  @note: This property is KVO compliant, but should only be read once at santad startup.
+///
+@property(readonly, nonatomic) NSString *eventLogType;
```
russellhancox: Document that FileLog is the default

Source/santad/Logs/SNTEventLog.h
```diff
@@ -23,15 +23,33 @@
...
+
+// String formatter helpers.
+- (void)addArgsForPid:(pid_t)pid toString:(NSMutableString *)str;
+- (NSString *)diskImageForDevice:(NSString *)devPath;
+- (NSString *)nameForUID:(uid_t)uid;
+- (NSString *)nameForGID:(gid_t)gid;
+- (NSString *)sanitizeCString:(const char *)str ofLength:(NSUInteger)length;
+- (NSString *)sanitizeString:(NSString *)inStr;
+- (NSString *)serialForDevice:(NSString *)devPath;
+- (NSString *)originalPathForTranslocation:(santa_message_t)message;
```
russellhancox: These should be moved to the syslog class, as they're not generic enough for all cases.

Source/santad/Logs/SNTEventLog.h
```diff
@@ -23,15 +23,33 @@
...
+
+// A cache for usernames and groups.
+@property(readonly, nonatomic) NSCache<NSNumber *, NSString *> *userNameMap;
+@property(readonly, nonatomic) NSCache<NSNumber *, NSString *> *groupNameMap;
+
+// A UTC Date formatter.
+@property(readonly, nonatomic) NSDateFormatter *dateFormatter;
```
russellhancox: Isn't this only used in the filelog class?

Source/santad/SNTApplication.m
```diff
@@ -68,7 +69,13 @@ - (instancetype)init {
       return nil;
     }
 
-    _eventLog = [[SNTEventLog alloc] init];
+    // Choose an event logger.
+    SNTConfigurator *configurator = [SNTConfigurator configurator];
+    if ([[configurator eventLogType] isEqualToString:@"Syslog"]) {
```
russellhancox: Also case-insensitive.

Alternatively, have the getters in SNTConfigurator parse these strings and convert to an enum so you can use a switch here.

Source/santad/Logs/SNTEventLog.h
```diff
@@ -23,15 +23,33 @@
...
 - (void)logDiskDisappeared:(NSDictionary *)diskProperties;
-
 - (void)logFileModification:(santa_message_t)message;
-
-- (void)saveDecisionDetails:(SNTCachedDecision *)cd;
 - (void)logDeniedExecution:(SNTCachedDecision *)cd withMessage:(santa_message_t)message;
 - (void)logAllowedExecution:(santa_message_t)message;
-
 - (void)logBundleHashingEvents:(NSArray<SNTStoredEvent *> *)events;
+- (void)writeLog:(NSString *)log;
```
russellhancox: This should be move to the syslog class

Source/santad/Logs/SNTEventLog.h
russellhancox: Consider removing from header, move to an extension

Source/common/SNTConfigurator.h
tburgin: Done

Source/santad/SNTApplication.m
tburgin: Done, moved to enum.

Source/common/SNTConfigurator.m
tburgin: Done.

Source/santad/Logs/SNTEventLog.h
tburgin: Just removed it entirely.

Source/santad/Logs/SNTEventLog.h
tburgin: can't :(

Source/common/SNTConfigurator.h
```diff
@@ -81,15 +81,16 @@
 
 ///
 ///  Defines how event logs are stored. Options are:
-///    Syslog: Sent to ASL or ULS (if built with the 10.12 SDK or later).
-///    Filelog: Sent to a file on disk. Use eventLogPath to specify a path.
+///    SNTEventLogTypeSyslog: Sent to ASL or ULS (if built with the 10.12 SDK or later).
+///    SNTEventLogTypeFilelog: Sent to a file on disk. Use eventLogPath to specify a path.
+///    Defaults to SNTEventLogTypeFilelog.
```
russellhancox: Document the strings these map to.

tburgin: Done

tburgin:
PTAL
* Ensure SNTFileEventLog objects can don't over retain themselves
* While profiling santad for SNTFileEventLog objects I noticed I was leaking an SNTXPCConnection object when starting up santabs. santad acts as a proxy to establish a direct connection between santabs <--> SantaGUI. Invalidate the proxy connection between santad <--> santabs once this is complete.

PR #248, opened 2018-04-09
sync-state: Mitigate com.apple.ManagedClient flapping

tburgin:
* If the SyncBaseURL disappears, keep the syncState active for 5 seconds before clearing.

### DIFF
--- a/Source/santad/SNTApplication.m
+++ b/Source/santad/SNTApplication.m
@@ -318,11 +318,17 @@ void diskDisappearedCallback(DADiskRef disk, void *context) {
 - (void)syncBaseURLDidChange:(NSURL *)syncBaseURL {
   if (syncBaseURL) {
     LOGI(@"Starting santactl with new SyncBaseURL: %@", syncBaseURL);
+    [NSObject cancelPreviousPerformRequestsWithTarget:[SNTConfigurator configurator]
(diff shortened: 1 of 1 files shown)


Source/santad/SNTApplication.m
```diff
@@ -318,11 +318,17 @@ - (void)clientModeDidChange:(SNTClientMode)clientMode {
...
     LOGI(@"Starting santactl with new SyncBaseURL: %@", syncBaseURL);
+    [NSObject cancelPreviousPerformRequestsWithTarget:[SNTConfigurator configurator]
+                                             selector:@selector(clearSyncState)
+                                               object:nil];
     [self startSyncd];
   } else {
     LOGI(@"SyncBaseURL removed, killing santactl pid: %i", self.syncdPID);
     [self stopSyncd];
-    [[SNTConfigurator configurator] clearSyncState];
+    // Keep the syncState active for 5 seconds in case com.apple.ManagedClient is flapping.
```
russellhancox: Why not make this something like 10 minutes, just to be on the safe side? Sure, it means removing the profile will take a while to take effect but that seems like the safest course of action anyway.

tburgin: SGTM, If the remover wants more immediate action. Restart santad.

PR #257, opened 2018-05-07
fileinfo: whitelist bundle extensions when performing an ancestor search

tburgin:
* Directories with a "Contents/Info.plist" entry can be mistaken as a bundle. Fix this by checking the ancestor candidate's extension against a whitelist.


Source/common/SNTFileInfo.m
```diff
@@ -317,6 +317,27 @@ - (BOOL)isMissingPageZero {
...
+  dispatch_once(&onceToken, ^{
+    NSArray *list = @[
+      @"app",
+      @"bundle",
+      @"framework",
+      @"kext",
+      @"xctest",
+      @"xpc",
+    ];
+    set = [NSSet setWithArray:list];
```
russellhancox: Combine:

```objc
set = [NSSet setWithArray:@[
  @"app",
  ...
]];
```

PR #258, opened 2018-05-08
Project: Use MOLXPCConnection

tburgin:
* Migrate from SNTXPCConnection to MOLXPCConnection
* Cleanup some leaking NSXPCConnection objects


Tests/LogicTests/SNTXPCConnectionTest.m
```diff
@@ -16,16 +16,15 @@
 
 #import <MOLCodesignChecker/MOLCodesignChecker.h>
 #import <OCMock/OCMock.h>
+#import <MOLXPCConnection/MOLXPCConnection.h>
 
-#import "SNTXPCConnection.h"
-
-@interface SNTXPCConnectionTest : XCTestCase
+@interface MOLXPCConnectionTest : XCTestCase
```
russellhancox: This exists: https://github.com/google/macops-MOLXPCConnection/blob/master/Tests/MOLXPCConnectionTests.m

Source/santad/SNTDaemonControlController.m
```diff
@@ -214,14 +215,16 @@ - (void)setBundlesEnabled:(BOOL)bundlesEnabled reply:(void (^)())reply {
 #pragma mark GUI Ops
 
 - (void)setNotificationListener:(NSXPCListenerEndpoint *)listener {
-  SNTXPCConnection *c = [[SNTXPCConnection alloc] initClientWithListener:listener];
+  // This will leak the underlying NSXPCConnection when "fast user switching" occurs.
+  // It is not worth the trouble to fix. Maybe furtue self will feel differently.
```
russellhancox: Spelling

russellhancox: Also, is fixing this not just a case of invalidating any existing listener?

tburgin: That would work until it doesn't. Invalidating the existing listener would prompt that listener to reconnect. Invaliding the new listener. Prompting the new listener to reconnect. Invalidating the previous,  existing listener...You get the point.

This *should* be mitigated by [0], but it's racey [1] and not guaranteed to run before the new user becomes active.

This could also be triggered by manually killing the existing Santa process.

[0] https://github.com/google/santa/blob/master/Source/SantaGUI/SNTAppDelegate.m#L45-L47
[1] https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPMultipleUsers/Concepts/UserSwitchNotifications.html#Event%20Timing

PR #264, opened 2018-05-21
Update project for Xcode 9.1, add codesigning flags

russellhancox:


Source/santad/DataLayer/SNTRuleTable.m
```diff
@@ -133,7 +133,8 @@ - (SNTRule *)ruleForBinarySHA256:(NSString *)binarySHA256
 
 #pragma mark Adding
 
-- (BOOL)addRules:(NSArray *)rules cleanSlate:(BOOL)cleanSlate error:(NSError **)error {
+- (BOOL)addRules:(NSArray *)rules cleanSlate:(BOOL)cleanSlate
+           error:(NSError * __autoreleasing *)error {
```
tburgin: I would double check this doesn't release too early.

Santa.xcodeproj/project.pbxproj
```diff
@@ -2014,13 +2038,37 @@
...
 				GCC_OPTIMIZATION_LEVEL = fast;
+				GCC_WARN_64_TO_32_BIT_CONVERSION = YES;
+				GCC_WARN_ABOUT_RETURN_TYPE = YES;
+				GCC_WARN_UNDECLARED_SELECTOR = YES;
+				GCC_WARN_UNINITIALIZED_AUTOS = YES;
+				GCC_WARN_UNUSED_FUNCTION = YES;
+				GCC_WARN_UNUSED_VARIABLE = YES;
 				LLVM_LTO = YES;
 				MACOSX_DEPLOYMENT_TARGET = 10.9;
+				OTHER_CODE_SIGN_FLAGS = "-o library,kill";
```
tburgin: "kill" is compatible for all the mach-os we compile?

Source/santad/DataLayer/SNTRuleTable.m
russellhancox: It doesn't, this is the fix recommended by clang (see [here](http://lists.llvm.org/pipermail/cfe-commits/Week-of-Mon-20170213/185409.html))

Santa.xcodeproj/project.pbxproj
russellhancox: As far as I know yes. I've also removed these flags from debug builds as it doesn't play nice with asan.

PR #266, opened 2018-05-21
santa-driver: do not invalidate cached decisions on KAUTH_VNODE_ACCESS

tburgin:
Also, catch KAUTH_VNODE_APPEND_DATA.

### DIFF
--- a/Source/santa-driver/SantaDecisionManager.cc
+++ b/Source/santa-driver/SantaDecisionManager.cc
@@ -614,11 +614,11 @@ extern "C" int vnode_scope_callback(
     sdm->DecrementListenerInvocations();
     return result;
   } else if (action & KAUTH_VNODE_WRITE_DATA || action & KAUTH_VNODE_APPEND_DATA) {
+    sdm->IncrementListenerInvocations();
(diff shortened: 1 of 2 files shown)


Source/santa-driver/SantaDecisionManager.cc
```diff
@@ -615,7 +613,11 @@ extern "C" int vnode_scope_callback(
                                     reinterpret_cast<int *>(arg3));
     sdm->DecrementListenerInvocations();
     return result;
-  } else if (action & KAUTH_VNODE_WRITE_DATA) {
+  } else if (action & KAUTH_VNODE_WRITE_DATA || action & KAUTH_VNODE_APPEND_DATA) {
+    if (!(action & KAUTH_VNODE_ACCESS)) {
+      auto vnode_id = sdm->GetVnodeIDForVnode(reinterpret_cast<vfs_context_t>(arg0), vp);
+      sdm->RemoveFromCache(vnode_id);
+    }
```
russellhancox: This needs to be below the `IncrementListenerInvocations` call below

tburgin: Oops, done.

Issue #265, opened 2018-05-21
log denied executions from dirty blocks


tburgin:
https://github.com/google/santa/blob/master/Source/santa-driver/SantaDecisionManager.cc#L459

PR #267, opened 2018-05-25
log deny because of dirty vnode

tburgin:


tburgin:
#265

Source/santa-driver/SantaDecisionManager.cc
```diff
@@ -459,6 +459,11 @@ int SantaDecisionManager::VnodeCallback(const kauth_cred_t cred,
   if (vnode_hasdirtyblks(vp)) {
     RemoveFromCache(vnode_id);
     returnedAction = ACTION_RESPOND_DENY;
+
+    char path[MAXPATHLEN];
+    int len = MAXPATHLEN;
+    path[MAXPATHLEN - 1] = 0;
+    LOGW("file has dirty blocks: %s", (vn_getpath(vp, path, &len) == ENOSPC) ? "unknown" : path);
```
russellhancox: vn_getpath can fail for reasons other than ENOSPC. Change this to:

```cpp
LOGW("file has dirty blocks: %s", vn_getpath(vp, path, &len) ? "unknown" : path);
```

tburgin:
PTAL

PR #271, opened 2018-05-30
SantaCache: Templatize key types

russellhancox:


Source/santa-driver/SantaDecisionManager.cc
```diff
@@ -29,9 +29,9 @@ bool SantaDecisionManager::init() {
   decision_dataqueue_lock_ = lck_mtx_alloc_init(sdm_lock_grp_, sdm_lock_attr_);
   log_dataqueue_lock_ = lck_mtx_alloc_init(sdm_lock_grp_, sdm_lock_attr_);
 
-  root_decision_cache_ = new SantaCache<uint64_t>(5000, 2);
-  non_root_decision_cache_ = new SantaCache<uint64_t>(500, 2);
-  vnode_pid_map_ = new SantaCache<uint64_t>(2000, 5);
+  root_decision_cache_ = new SantaCache<uint64_t, uint64_t>(5000, 2);
+  non_root_decision_cache_ = new SantaCache<uint64_t, uint64_t>(500, 2);
```
tburgin: Until we fix the issue where system binaries are executing on a non-root VFS we should bump the non_root_decision_cache_ up some.

Tests/LogicTests/SantaCacheTest.mm
```diff
@@ -120,57 +110,105 @@ - (void)testThreading {
 }
 
 - (void)testCount {
-  auto sut = new SantaCache<uint64_t>();
+  auto sut = SantaCache<uint64_t, int>();
```
tburgin: I assume you wanted to test this with int?

Source/santa-driver/SantaDecisionManager.cc
russellhancox: I have a follow up PR that gets rid of the split caches

Tests/LogicTests/SantaCacheTest.mm
russellhancox: I did

Source/santa-driver/SantaCache.h
```diff
@@ -40,17 +40,29 @@
...
 #endif // KERNEL
 
+/**
+  A type to specialize to help SantaCache with its hashing.
+
+  The default works for numeric types with a multiplicative hash
+  using a prime near to the golden ratio, per Knuth.
+*/
+template<typename T> uint64_t SantaCacheHasher(T const& t) {
+  return t * 11400714819323198549UL;
```
nguyen-phillip: Just a minor aside: after looking this up online, most versions that I've seen tend to apply a bit shift after the multiplication so that only the high bits of the result are used.  That's more difficult to accomplish with 64-bit operands but, I think currently this hash function is biased towards the low bits of t.  And taking the modulo of the hash would be sort of the same as just taking the modulo of t, throwing away most of the info in the high bits.

tburgin: I had that same thought when we started discussing added templated keys. Here is some terrible code I wrote at the time to ensure the hash distribution was okay for a struct with 2 uint64_t values. Good enough distribution for our purposes.

```
 
typedef struct {
	uint64_t fs_id;
	uint64_t vn_id;
} vnode_t;
 
uint64_t hash(vnode_t vnode, uint64_t buckets) {
	uint64_t h = vnode.fs_id * 11400714819323198549ul;
	h += vnode.vn_id * 11400714819323198549ul;
	return h % buckets;
}
 
int main(int argc, char *argv[]) {
	@autoreleasepool {
		NSMutableDictionary *cache = [NSMutableDictionary dictionaryWithCapacity:0xFFFFFF];
		for (uint64_t i = 0; i < 0xFFFFFF; ++i) {
			vnode_t vnode = {i < 0xFF ? 123413 : 567891, i};
			uint64_t h = hash(vnode, 10000);
			cache[@(h)] = @([cache[@(h)] unsignedLongLongValue] + 1);
		}
		uint64_t min = UINT64_MAX;
		uint64_t max = 0;
		for (int i = 0; i < 10000; ++i) {
			uint64_t value = [cache[@(i)] unsignedLongLongValue];
			if (value < min) {
				 min = value;
			} else if (value > max) {
				max = value;
			}
		}
		NSLog(@"%min: %llu max: %llu", min, max);
	}
}
 
2018-05-18 13:26:08.168 Untitled 2[76301:1359961] min: 1670 max: 1686
```

tburgin: You may have a point 🥇 
```
uint64_t hash(vnode_t vnode, uint64_t buckets) {
	return (vnode.fs_id + vnode.vn_id) % buckets;
}
Untitled 2[4515:47916] min: 1677 max: 1678
```

tburgin: The above test is quite limited and probably flawed.

nguyen-phillip: Ah cool!  I was thinking of writing some test code but wasn't quite sure what the best way to do it would be.  For testing plain integer hashing, we'd probably want to make sure that sets of similar keys still got spread out uniformly.  The (t * 11400714819323198549) hash function will always send even numbers to even numbers when the number of buckets is even for instance.  So in that case you'd have to rely on the keys being somewhat uniformly distributed in order to have a chance of the hashed indices being uniformly distributed.

russellhancox: Coming back to this, the histogram added in #275 running against the hashing code in #276 shows a pretty uniform distribution of actual executed binaries. After executing all the binaries in /usr/bin, I got:

```
There are 7199 empty buckets
   1: 926
   2: 57
   3: 10
```

PR #275, opened 2018-06-01
SantaCache: Add command to print histogram of bucket distribution

russellhancox:
The command only exists in Debug builds.

This currently only prints the distribution of the non-root cache. In the near future I'll unify the caches again which stops this being a problem.

### DIFF
--- a/Source/common/SNTKernelCommon.h
+++ b/Source/common/SNTKernelCommon.h
@@ -21,9 +21,6 @@
 #ifndef SANTA__COMMON__KERNELCOMMON_H
 #define SANTA__COMMON__KERNELCOMMON_H
 
-// Defines the lengths of paths and Vnode IDs passed around.
-#define MAX_VNODE_ID_STR 21  // digits in UINT64_MAX + 1 for NULL-terminator
-
 // Defines the name of the userclient class and the driver bundle ID.
 #define USERCLIENT_CLASS "com_google_SantaDriver"
 #define USERCLIENT_ID "com.google.santa-driver"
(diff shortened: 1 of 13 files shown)


Source/common/SNTKernelCommon.h
```diff
@@ -99,4 +97,10 @@ typedef struct {
   char pname[MAXPATHLEN];
 } santa_message_t;
 
+// Used for the kSantaUserClientCacheBucketCount request.
+typedef struct {
+  uint16_t per_bucket[1024];
```
tburgin: Make 1024 a constant and use throughout. Or use `sizeof(counts.per_bucket) / sizeof(uint16_t)` where you are using 1024 throughout.

russellhancox:
PTAL: I've switched from printing the count in every bucket to printing the number of buckets that match a count. For easier reading I've also defaulted to just printing the numbers, unless you pass the -g flag.

Source/common/SNTKernelCommon.h
russellhancox: Done.

Source/santactl/Commands/SNTCommandCacheHistogram.m
```diff
@@ -0,0 +1,81 @@
...
+          @"  Use -g to get 'graphical' output\n"
+          @"Only available in DEBUG builds.");
+}
+
+- (void)runWithArguments:(NSArray *)arguments {
+  [[self.daemonConn remoteObjectProxy] cacheBucketCount:^(NSArray *counts) {
+    NSMutableDictionary<NSNumber *, NSNumber *> *k = [NSMutableDictionary dictionary];
+    [counts enumerateObjectsUsingBlock:^(id  _Nonnull obj, NSUInteger idx, BOOL * _Nonnull stop) {
+      if (!k[obj]) {
+        k[obj] = @1;
```
tburgin: This shouldn't be needed. Your L59 should handle setting 1 as the initial value of k[obj].

PR #276, opened 2018-06-01
santa-driver: Switch to a struct for vnode IDs, holding both the file…

russellhancox:
…system ID and vnode ID.

Also drop the separate caches for root/non-root as this doesn't offer any benefit anymore.


Source/santa-driver/SantaDecisionManager.cc
```diff
@@ -19,6 +19,10 @@ OSDefineMetaClassAndStructors(SantaDecisionManager, OSObject);
 
 #pragma mark Object Lifecycle
 
+template<> uint64_t SantaCacheHasher<santa_vnode_id_t>(santa_vnode_id_t const& s) {
+  return SantaCacheHasher<uint64_t>(s.fsid) ^ (SantaCacheHasher<uint64_t>(s.fileid) << 1);
```
nguyen-phillip: Given the current hash function and that almost all the variation is in s.fileid, maybe it would be better to use (s.fsid << 1) ^ s.fileid instead.  Because otherwise, if there's only one s.fsid in use, the last bit on the output of this function will always be the last bit of s.fsid, and this will trickle down to SantaCache's hash function where the modulo by a power of 2 will preserve that last bit, and so we'll end up using only half of the available buckets, either all the even indices or all the odd indices, depending on the parity of s.fsid.

Source/common/SNTKernelCommon.h
```diff
@@ -80,10 +80,30 @@ typedef enum {
...
+#ifdef __cplusplus
+  bool operator==(const santa_vnode_id_t& rhs) const {
+    return fsid == rhs.fsid && fileid == rhs.fileid;
+  }
+  bool operator!=(const santa_vnode_id_t& rhs) const {
+    return !(*this == rhs);
+  }
+  // This _must not_ be used for anything security-sensitive. It exists solely to make
+  // the msleep/wakeup calls easier.
+  uint64_t simple_id() const {
```
nguyen-phillip: Maybe rename this to unsafe_simple_id?

russellhancox: Good call

Source/santa-driver/SantaCache.h
```diff
@@ -206,7 +206,7 @@ template<typename KeyT, typename ValueT> class SantaCache {
       lock(bucket);
       struct entry *entry = (struct entry *)((uintptr_t)bucket->head - 1);
       while (entry != nullptr) {
-        if (entry->key != 0) ++count;
+        if (entry->key != key_zero_) ++count;
```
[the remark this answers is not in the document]
russellhancox: > I don't think there's anything special about entry->key being zero is there?

No, the counting code was wrong and was checking for a zero key rather than a zero value, which is wrong.

Source/santa-driver/SantaCache.h
```diff
@@ -338,7 +338,8 @@ template<typename KeyT, typename ValueT> class SantaCache {
   /**
     Holder for a 'zero' entry for the current type
   */
-  const ValueT zero_ = {};
+  KeyT key_zero_ = {};
```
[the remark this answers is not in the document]
russellhancox: No, see above.

Source/santa-driver/SantaDecisionManager.cc
russellhancox: Good call, that does help with distribution.

Source/santad/SNTDriverManager.m
```diff
@@ -186,30 +186,31 @@ - (kern_return_t)postToKernelAction:(santa_action_t)action forVnodeID:(uint64_t)
...
                                    0) == KERN_SUCCESS;
 }
 
-- (santa_action_t)checkCache:(uint64_t)vnodeID {
-  uint32_t input_count = 1;
-  uint64_t vnode_action = 0;
+- (santa_action_t)checkCache:(santa_vnode_id_t)vnodeID {
+  struct {
+    uint64_t vnode_action;
+  } output;
```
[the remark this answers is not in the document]
russellhancox: It did for the IOConnectCallStruct method but IOConnectCallMethod handles both types so I've switched.

PR #278, opened 2018-06-04
santad: Wait for driver appearance using IOKit notifications.

russellhancox:


Source/santad/SNTDriverManager.m
```diff
@@ -86,6 +70,61 @@ - (void)dealloc {
...
+    dispatch_semaphore_signal(sema);
+  };
+
+  LOGI(@"Waiting for Santa driver to become available");
+
+  IOServiceAddMatchingNotification(notificationPort,
+                                   kIOMatchedNotification,
+                                   matchingDict,
+                                   driverAppearedHandler,
+                                   (__bridge_retained void *)block,
```
tburgin: Does the block fire on a different thread? Just want to make sure the below dispatch wait won’t block the execution of the block due to notification.

russellhancox: It did, I've changed the scope of the PR.

Source/common/SNTXPCControlInterface.h
```diff
@@ -37,6 +37,7 @@
 - (void)flushCache:(void (^)(BOOL))reply;
 - (void)cacheBucketCount:(void (^)(NSArray *))reply;
 - (void)checkCacheForVnodeID:(santa_vnode_id_t)vnodeID withReply:(void (^)(santa_action_t))reply;
+- (void)connectionEstablished:(void (^)(BOOL))reply;
```
tburgin: What connection? The driver?

Source/santactl/Commands/SNTCommandStatus.m
```diff
@@ -190,6 +196,7 @@ - (void)runWithArguments:(NSArray *)arguments {
     printf("%s\n", [statsStr UTF8String]);
   } else {
     printf(">>> Daemon Info\n");
+    printf("  %-25s | %s\n", "Connected", driverConnected ? "Yes" : "No");
```
tburgin: Something like: "Driver Connected"?
Also, add this to the JSON output.

Source/santad/SNTDriverManager.m
```diff
@@ -23,67 +23,97 @@
...
+}
 
-      if (!serviceObject) {
-        LOGD(@"Waiting for Santa driver to become available");
-        sleep(delay);
-        if (delay < MAX_DELAY) delay *= 2;
-      }
-    } while (!serviceObject);
-    CFRelease(classToMatch);
+- (void)dealloc {
```
tburgin: Given SNTDriverManager is essentially a singleton, this should never be called. Add a note.

Source/santad/SNTDriverManager.m
```diff
@@ -23,67 +23,97 @@
...
+// Helper function used with IOServiceAddMatchingNotification which expects
+// a DriverAppearedBlock to be passed as the 'reference' argument. The block
+// will be called for each object iterated over and if the block wants to keep
+// the object, it should call IOObjectRetain().
+typedef void (^DriverAppearedBlock)(io_object_t object);
+static void driverAppearedHandler(void *info, io_iterator_t iterator) {
+  DriverAppearedBlock block = (__bridge DriverAppearedBlock)info;
+  io_object_t object = 0;
+  while ((object = IOIteratorNext(iterator))) {
+    block(object);
```
tburgin: You may want to double check block() does indeed exist.

Source/santad/SNTDriverManager.m
```diff
@@ -23,67 +23,97 @@
...
+  DriverAppearedBlock block = ^(io_object_t object) {
     // This calls `initWithTask`, `attach` and `start` in `SantaDriverClient`
-    kr = IOServiceOpen(serviceObject, mach_task_self(), 0, &_connection);
-    IOObjectRelease(serviceObject);
+    kern_return_t kr = IOServiceOpen(object, mach_task_self(), 0, &_connection);
     if (kr != kIOReturnSuccess) {
-      LOGD(@"Failed to open Santa driver service");
-      return nil;
+      LOGE(@"Failed to open santa-driver service: 0x%X", kr);
+      exit(1);
```
tburgin: Is exiting the best option for all these cases? I could see santad flapping.

russellhancox: I can't see a better option - if establishing a connection failed after the driver appeared there's no point waiting for the driver to appear again and there's not a notification that whatever caused connection failures is now fixed. Exiting and hoping that the next time starting might lead to a better outcome is about all we can do.

Source/santad/SNTDriverManager.m
russellhancox: Done

Source/santad/SNTDriverManager.m
russellhancox: We might treat SNTDriverManager like a singleton but it isn't implemented as one.

Source/santactl/Commands/SNTCommandStatus.m
russellhancox: Done

Source/common/SNTXPCControlInterface.h
russellhancox: I've renamed to make this clearer

tburgin:
LGTM

PR #281, opened 2018-06-08
santa-driver: Re-factor some destruction methods

russellhancox:
This fixes IOServiceClose() being called on the driver

Also don't delete and re-enqueue the log data queue if it fails to enqueue, this is pointless.

### DIFF
--- a/Source/santa-driver/SantaDriverClient.cc
+++ b/Source/santa-driver/SantaDriverClient.cc
@@ -48,23 +48,21 @@ bool SantaDriverClient::start(IOService *provider) {
 }
 
 void SantaDriverClient::stop(IOService *provider) {
-  super::stop(provider);
   myProvider = nullptr;
   decisionManager = nullptr;
+  super::stop(provider);
+}
+
+IOReturn SantaDriverClient::clientDied() {
+  LOGI("Client died.");
+  decisionManager->DisconnectClient(true);
+  return terminate() ? kIOReturnSuccess : kIOReturnError;
 }
 
 IOReturn SantaDriverClient::clientClose() {
-  decisionManager->DisconnectClient(true);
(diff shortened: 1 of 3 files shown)


Source/santa-driver/SantaDriverClient.h
```diff
@@ -49,9 +49,7 @@ class com_google_SantaDriverClient : public IOUserClient {
 
   ///  Called when a client disconnects
   IOReturn clientClose() override;
-
-  ///  Called when the driver is shutting down
-  bool terminate(IOOptionBits options) override;
+  IOReturn clientDied() override;
```
tburgin: Add a comment

Source/santa-driver/SantaDriverClient.cc
```diff
@@ -48,23 +48,21 @@ bool SantaDriverClient::start(IOService *provider) {
...
-bool SantaDriverClient::terminate(IOOptionBits options) {
-  decisionManager->DisconnectClient();
+IOReturn SantaDriverClient::clientClose() {
   LOGI("Client disconnected.");
-
-  if (myProvider && myProvider->isOpen(this)) myProvider->close(this);
-
-  return super::terminate(options);
+  decisionManager->DisconnectClient(false);
+  return terminate() ? kIOReturnSuccess : kIOReturnError;
```
tburgin: Is there any chance the client is still connected when calling terminate here? Should kIOServiceSynchronous or kIOServiceRequired be used?

Source/santa-driver/SantaDriverClient.h
russellhancox: Done

Source/santa-driver/SantaDriverClient.cc
russellhancox: There shouldn't be, these methods are called in response to a client disconnecting and we only allow a single client.

PR #284, opened 2018-06-20
Add CMake support, implement fuzzers

alessandrogario:
The new CMake project allows the user to select which SDK version to use. The Xcode path is also configurable to support non-standard installation paths and/or systems with multiple versions installed.

Code signing can now be configured via command line, using the CODESIGN_IDENTITY environment variable.

New fuzzing targets (libFuzzer)
 - SantaCache
 - santactl
 - santad

New make targets:
 - tests: Runs the tests
 - fuzz: Runs the fuzzer
 - redist: Regenerates the redistributable folder
 - install: Installs Santa

A small issue has been found in santactl when the the sync-server returns a valid JSON that does not get deserialized to a dictionary object (not yet fixed inside this PR).


alessandrogario:
I have signed the CLA! 👍

Fuzzing/santactl/src/main.mm
```diff
@@ -0,0 +1,48 @@
...
+  if ([response objectForKey:kRules] == nil) {
+    return 0;
+  }
+
+  SNTCommandSyncState *state = [[SNTCommandSyncState alloc] init];
+  if (state == nil) {
+    return 0;
+  }
+
+	SNTCommandSyncRuleDownload *obj = [[SNTCommandSyncRuleDownload alloc] initWithState:state];
```
russellhancox: nit: indented too far

Source/santa-driver/CMakeLists.txt
```diff
@@ -0,0 +1,111 @@
...
+  )
+
+  if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+    target_compile_definitions("${PROJECT_NAME}" PRIVATE DEBUG)
+  else()
+    target_compile_definitions("${PROJECT_NAME}" PRIVATE NDEBUG)
+  endif()
+
+  target_include_directories("${PROJECT_NAME}" SYSTEM PRIVATE
+    "${XCODE_ROOT_FOLDER}/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX${MACOSX_VERSION_MIN}.sdk/System/Library/Frameworks/Kernel.framework/PrivateHeaders"
```
russellhancox: If you remove the reference to MACOSX_VERSION_MIN here it should be less brittle and will compile with whatever the latest SDK is. Alternatively, we'll need a separate variable to differentiate the minimum supported version vs. the SDK version.

Source/santa-driver/SantaDriver.cc
```diff
@@ -53,3 +53,17 @@ SantaDecisionManager *SantaDriver::GetDecisionManager() const {
...
+extern "C" {
+extern kern_return_t _start(kmod_info_t *ki, void *data);
+extern kern_return_t _stop(kmod_info_t *ki, void *data);
+
+__attribute__((visibility("default"))) KMOD_EXPLICIT_DECL(com.google.santa-driver, SANTA_VERSION, _start, _stop)
+__private_extern__ kmod_start_func_t *_realmain = 0;
+__private_extern__ kmod_stop_func_t *_antimain = 0;
+__private_extern__ int _kext_apple_cc = __APPLE_CC__ ;
+}
+#endif
```
russellhancox: Could this section instead be in a separate file, main.c or similar?

Fuzzing/santactl/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+
+  add_executable("${PROJECT_NAME}" EXCLUDE_FROM_ALL ${PROJECT_SOURCEFILES})
+
+  target_compile_options("${PROJECT_NAME}" PRIVATE
+    -fobjc-arc -Wno-everything -fmodules -fcxx-modules
+    -mmacosx-version-min=${MACOSX_VERSION_MIN}
+  )
+
+  target_compile_definitions("${PROJECT_NAME}" PRIVATE
+    SNT_OS_LOG_CATEGORY=\"${PROJECT_NAME}\"
```
russellhancox: What's this define for?

Fuzzing/santactl/src/main.mm
```diff
@@ -0,0 +1,48 @@
...
+#include <cstdint>
+#include <vector>
+
+#include <SNTCommandSyncRuleDownload.h>
+#include <SNTCommandSyncState.h>
+#include <SNTCommandSyncConstants.h>
+#include <SNTRule.h>
+
+extern "C" int LLVMFuzzerTestOneInput(const std::uint8_t *data, std::size_t size) {
+  NSData* buffer = [NSData dataWithBytes:static_cast<const void *>(data) length:size];
```
russellhancox: To be consistent with the rest of the source, the * should be with the variable name:

`NSData *buffer`

Fuzzing/santactl/src/main.mm
```diff
@@ -0,0 +1,48 @@
...
+  if (buffer == nil) {
```
russellhancox: Also consistency: we generally use the form `if (buffer) {` here.

Fuzzing/santad/src/checkCacheForVnodeID.mm
```diff
@@ -0,0 +1,45 @@
...
+
+#import <MOLXPCConnection/MOLXPCConnection.h>
+
+#import "SNTCommandController.h"
+#import "SNTRule.h"
+#import "SNTXPCControlInterface.h"
+
+#pragma pack(push, 1)
+
+#pragma pack(pop)
```
russellhancox: Are these 2 pragma's leftovers from something?

Tests/santa-driver/main.mm
```diff
@@ -0,0 +1,51 @@
+#import "SNTDriverManager.h"
```
russellhancox: Can you add a comment at the top of this file describing what it's testing. I'm guessing it's a test that attempts to connect to the driver from multiple threads simultaneously doesn't cause crashes.

alessandrogario: You are right! I have added a comment at the top of the file

alessandrogario:
@russellhancox Thanks so much for the review! I've updated branch with the changes you have requested

russellhancox:
Thanks again for the submission!

PR #287, opened 2018-07-05
Update MOLXPCConnection; add support for unprivileged XPC interfaces

alessandrogario:


Source/santactl/Commands/sync/SNTCommandSync.m
```diff
@@ -57,6 +57,9 @@ + (NSString *)longHelpText {
 }
 
 - (void)runWithArguments:(NSArray *)arguments {
+  // Connect to santad while we are root, so that we pass the XPC authentication
+  [self.daemonConn resume];
```
tburgin: All instances of `santactl sync` must now run as root. `requiresRoot` should return YES.

PR #288, opened 2018-07-06
Various fixes (documentation, and an additional check on the JSON received from the syncserver)

alessandrogario:
This PR contains some fixes on the documentation (caused warnings when compiling with CMake) and a small check on the santactl executable when retrieving JSON from the syncserver (found with libFuzzer).

### DIFF
--- a/Source/santactl/Commands/sync/SNTCommandSyncRuleDownload.m
+++ b/Source/santactl/Commands/sync/SNTCommandSyncRuleDownload.m
@@ -82,7 +82,6 @@
   do {
     NSDictionary *requestDict = cursor ? @{kCursor : cursor} : @{};
     NSDictionary *response = [self performRequest:[self requestWithDictionary:requestDict]];
-    if (!response) return nil;
 
     if (![response isKindOfClass:[NSDictionary class]]) {
       return nil;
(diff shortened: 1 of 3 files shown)


Source/santactl/Commands/sync/SNTCommandSyncRuleDownload.m
```diff
@@ -83,6 +83,11 @@ - (BOOL)sync {
     NSDictionary *requestDict = cursor ? @{kCursor : cursor} : @{};
     NSDictionary *response = [self performRequest:[self requestWithDictionary:requestDict]];
     if (!response) return nil;
+
+    if (![response isKindOfClass:[NSDictionary class]]) {
```
russellhancox: Might as well remove the nil check on line 85 - this check will handle both.

alessandrogario:
Hey @russellhancox! You are sooo fast at looking at PR! Thank you so much! :)

Do you think there is any value in only allowing executables owned by root:root when santad is not connected?

russellhancox:
I could imagine that leading to some unexpected consequences and without santad connected there's no (clean) way to inform the user why something wasn't allowed to execute, which could be very confusing. Open an issue and we can discuss further

PR #296, opened 2018-09-19
santad: add critical system binaries

### LINKED ISSUE #295: System freeze on MacOS Mojave
Hello,

Santa (update 0.9.28) causes random occasional system freezes on Mojave beta 10 and 11.

From the end-user perspective, it feels like a kernel level lock up and happens within a half hour or so of boot on beta 10 and 11. Seems to especially be triggered by WindowServer context switches (wake from sleep, switch virtual desktops, etc.).

tburgin:
* In Mojave `/usr/libexec/trustd` needs to run in order to check arbitrary code signatures. This may be a departure from previous version of macOS. If `trustd` is killed, Santa can cause a deadlock. This fix pre-validates and whitelists `trustd` at `santad` startup time.
* Fixes #295

### DIFF
--- a/Source/santad/DataLayer/SNTRuleTable.m
+++ b/Source/santad/DataLayer/SNTRuleTable.m
@@ -17,7 +17,9 @@
 #import <MOLCertificate/MOLCertificate.h>
 #import <MOLCodesignChecker/MOLCodesignChecker.h>
 
+#import "SNTCachedDecision.h"
 #import "SNTConfigurator.h"
+#import "SNTFileInfo.h"
 #import "SNTLogging.h"
 #import "SNTRule.h"
 
@@ -31,6 +33,7 @@ static const NSUInteger kTransitiveRuleExpirationSeconds = 6 * 30 * 24 * 3600;
 @property NSString *santadCertSHA;
 @property NSString *launchdCertSHA;
 @property NSDate *lastTransitiveRuleCulling;
+@property NSDictionary *criticalSystemBinaries;
 @end
 
 @implementation SNTRuleTable
@@ -64,7 +67,8 @@ static const NSUInteger kTransitiveRuleExpirationSeconds = 6 * 30 * 24 * 3600;
   // Save hashes of the signing certs for launchd and santad.
   // Used to ensure rules for them are not removed.
   self.santadCertSHA = [[[[MOLCodesignChecker alloc] initWithSelf] leafCertificate] SHA256];
-  self.launchdCertSHA = [[[[MOLCodesignChecker alloc] initWithPID:1] leafCertificate] SHA256];
+  MOLCodesignChecker *launchdCSInfo = [[MOLCodesignChecker alloc] initWithPID:1];
+  self.launchdCertSHA = launchdCSInfo.leafCertificate.SHA256;
 
   // Ensure the certificates used to sign the running launchd/santad are whitelisted.
   // If they weren't previously and the database is not new, log an error.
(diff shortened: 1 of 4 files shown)


Source/santad/SNTExecutionController.m
```diff
@@ -116,21 +116,28 @@ - (void)validateBinaryWithMessage:(santa_message_t)message {
...
-  // Actually make the decision (and refresh rule access timestamp).
-  SNTCachedDecision *cd = [self.policyProcessor decisionForFileInfo:binInfo
+  // If needed, actually make the decision (and refresh rule access timestamp).
+  if (!cd) {
+    cd = [self.policyProcessor decisionForFileInfo:binInfo
                                                          fileSHA256:nil
                                                   certificateSHA256:csInfo.leafCertificate.SHA256];
-  cd.certCommonName = csInfo.leafCertificate.commonName;
+    cd.certCommonName = csInfo.leafCertificate.commonName;
+  }
```
nguyen-phillip: I think this might be easier to follow if it were structured as
```
if (!cd) {
  // declare csInfo
  if (binInfo.isMachO) {
    // get csInfo
  }
  // get cd from policyProcessor
}
```
instead of having two separate checks for !cd

Source/santad/DataLayer/SNTRuleTable.m
```diff
@@ -87,6 +91,31 @@ - (uint32_t)initializeDatabase:(FMDatabase *)db fromVersion:(uint32_t)version {
...
+      cd.decision = SNTEventStateAllowBinary;
+      cd.decisionExtra = @"critical system binary";
+      cd.sha256 = binInfo.SHA256;
+
+      // Not needed, but nice for logging.
+      cd.certSHA256 = csInfo.leafCertificate.SHA256;
+      cd.certCommonName = csInfo.leafCertificate.commonName;
+
+      bins[binInfo.SHA256] = cd;
+    }
```
nguyen-phillip: log error if signing info doesn't match?

Source/santad/DataLayer/SNTRuleTable.m
```diff
@@ -87,6 +91,31 @@ - (uint32_t)initializeDatabase:(FMDatabase *)db fromVersion:(uint32_t)version {
         self.launchdCertSHA, @(SNTRuleStateWhitelist), @(SNTRuleTypeCertificate)];
   }
 
+  // Setup critical system binaries
+  // TODO(tburgin): Add the Santa components to this feature and remove the santadCertSHA rule.
+  NSMutableDictionary *bins = [NSMutableDictionary dictionary];
+  for (NSString *path in @[ @"/usr/libexec/trustd" ]) {
+    SNTFileInfo *binInfo = [[SNTFileInfo alloc] initWithPath:path];
```
nguyen-phillip: Does this fail the right way if there's no file at path?  I'm guessing csInfo will end up nil and then the call to signingInformationMatches: will return NO?

Source/santad/SNTExecutionController.m
```diff
@@ -116,21 +116,28 @@ - (void)validateBinaryWithMessage:(santa_message_t)message {
   }
 
   // Get codesigning info about the file but only if it's a Mach-O.
+  // If the binary is a critical system binary, don't check its signiture. The binary was validated
```
nguyen-phillip: typo: signiture -> signature

tburgin:
PTAL

ameyah:
Thanks a ton for this fix @tburgin 

I was wondering when the next release would be? :)

russellhancox:
A couple of hours, most likely

ameyah:
Awesome, Thanks again!

PR #302, opened 2018-09-26
santa-driver: add back the root and non-root caches

tburgin:
* santa-driver now supports 64 bit file system ids. Identifying root and non-root file system ids on executing binaries seems to now work consistently.
* Clarify buckets and entries in non-graphical cachehistogram output.

This pull is mainly a line by line cherrypick revert of 15fa53d744368752ff1b55dcbdd62020e060c629.

### DIFF
--- a/Source/santa-driver/SantaDecisionManager.cc
+++ b/Source/santa-driver/SantaDecisionManager.cc
@@ -41,7 +41,8 @@ bool SantaDecisionManager::init() {
   decision_dataqueue_lock_ = lck_mtx_alloc_init(sdm_lock_grp_, sdm_lock_attr_);
   log_dataqueue_lock_ = lck_mtx_alloc_init(sdm_lock_grp_, sdm_lock_attr_);
 
-  decision_cache_ = new SantaCache<santa_vnode_id_t, uint64_t>(10000, 2);
+  root_decision_cache_ = new SantaCache<santa_vnode_id_t, uint64_t>(5000, 2);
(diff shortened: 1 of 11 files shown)


Source/santa-driver/SantaDecisionManager.cc
```diff
@@ -102,6 +105,17 @@ void SantaDecisionManager::ConnectClient(pid_t pid) {
 
   client_pid_ = pid;
 
+  // Determine root fsid
+  vfs_context_t ctx = vfs_context_create(NULL);
```
russellhancox: Use nullptr

Source/santa-driver/SantaDecisionManager.h
```diff
@@ -123,10 +123,15 @@ class SantaDecisionManager : public OSObject {
...
 
   /// Returns the number of entries in the cache.
-  uint64_t CacheCount() const;
+  uint64_t RootCacheCount() const;
+  uint64_t NonRootCacheCount() const;
+
+  /**
+   Clears the cache(s). If non_root_only is true, only the non-root cache
+   is cleared.
+   */
```
russellhancox: Indent -1

Source/santa-driver/SantaDecisionManager.cc
tburgin: Done.

Source/santa-driver/SantaDecisionManager.h
tburgin: Done.

PR #311, opened 2018-10-09
config: Ensure syncBaseURL ends with a /

### LINKED ISSUE #310: SyncBaseURL with a path requires a slash
This works
```
<key>SyncBaseURL</key>
<string>https://santa:8080/api/v1/</string>
```
This causes a 404
```
<key>SyncBaseURL</key>
<string>https://santa:8080/api/v1</string>
```

Probably causes by a strings join instead of parsing the URL and building on the path.

russellhancox:
Without the trailing / the last path component is removed by `URLWithString:relativeToURL:`

Fixes #310

### DIFF
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -352,6 +352,7 @@ static NSString *const kSyncCleanRequired = @"SyncCleanRequired";
 
 - (NSURL *)syncBaseURL {
   NSString *urlString = self.configState[kSyncBaseURLKey];
+  if (![urlString hasSuffix:@"/"]) urlString = [urlString stringByAppendingString:"/"];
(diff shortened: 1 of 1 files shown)


Source/common/SNTConfigurator.m
```diff
@@ -352,6 +352,7 @@ - (NSRegularExpression *)fileChangesRegex {
 
 - (NSURL *)syncBaseURL {
   NSString *urlString = self.configState[kSyncBaseURLKey];
+  if (![urlString hasSuffix:@"/"]) urlString = [urlString stringByAppendingString:"/"];
```
nguyen-phillip: oops, missed this: @"/"

russellhancox: Good catch.

PR #313, opened 2018-11-01
add in-kernel filemod prefix filter

tburgin:
* Allow santa-driver to ignore file changes with a prefix filter.
* Stops unnecessary generation of events that would be ignored by santad with a regex or lack of one.
* Prefixes are configurable with the mobileconfig.

### DIFF
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -76,13 +76,14 @@
 ///  A list of ignore prefixes which are checked in-kernel.
 ///  This is more performant than FileChangesRegex when ignoring whole directory trees.
 ///
-///  For example adding a prefix of "/tmp/" will turn off file change log generation in-kernel
-//   for that entire tree. Since they are ignored by the kernel, they never reach santad and are
-///  not seen by the fileChangesRegex. Note the trailing "/", without it any file or directory
-///  starting with "/tmp" would be ignored.
+///  For example adding a prefix of "/private/tmp/" will turn off file change log generation
+///  in-kernel for that entire tree. Since they are ignored by the kernel, they never reach santad
+///  and are not seen by the fileChangesRegex. Note the trailing "/", without it any file or
+///  directory starting with "/private/tmp" would be ignored.
 ///
 ///  By default "/." and "/dev/" are added.
-///  32 additional entries are allowed. Anything over 32 will be ignored.
+///  Memory in the kernel is precious. 256 total bytes are allowed, use them wisely. Anything
+///  totaling up to more than 256 bytes will be ignored.
 ///
 ///  To disable file change logging completely add "/".
 ///  Filters are only applied on santad startup.
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -109,7 +109,9 @@
 ///  "/🖖"
 ///
 ///                          -> [0xa4] -> [0x98]
+///
 ///  [/] -> [0xf0] -> [0x9f]
+///
 ///                          -> [0x96] -> [0x96]
 ///
 ///  To disable file change logging completely add "/".
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -83,10 +83,11 @@
 ///
 ///  By default "/." and "/dev/" are added.
 ///
-///  Memory in the kernel is precious. A total of 256 nodes are allowed for this configuration.
+///  Memory in the kernel is precious. A total of MAXPATHLEN (1024) nodes are allowed.
+///  Using all 1024 nodes will result in santa-driver allocating ~2MB of wired memory.
 ///  An ASCII character uses 1 node. An UTF-8 encoded Unicode character uses 1-4 nodes.
-///  Prefixes are added to the running config in-order one by one. The prefix will be ignored if
-///  (the running config's current size) + (the prefix's size) totals up to more than 256 nodes.
+///  Prefixes are added to the running config in-order, one by one. The prefix will be ignored if
+///  (the running config's current size) + (the prefix's size) totals up to more than 1024 nodes.
 ///  The running config is stored in a prefix tree.
 ///  Prefixes that share prefixes are effectively de-duped; their shared node sized components only
 ///  take up 1 node. For example these 3 prefixes all have a common prefix of "/private/".
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -83,14 +83,14 @@
 ///
 ///  By default "/." and "/dev/" are added.
 ///
-///  Memory in the kernel is precious. A total of 256 bytes are allowed for this configuration.
-///  An ASCII character uses 1 byte. An UTF-8 encoded Unicode character uses 1-4 bytes.
+///  Memory in the kernel is precious. A total of 256 nodes are allowed for this configuration.
+///  An ASCII character uses 1 node. An UTF-8 encoded Unicode character uses 1-4 nodes.
 ///  Prefixes are added to the running config in-order one by one. The prefix will be ignored if
-///  (the running config's current size) + (the prefix's size) totals up to more than 256 bytes.
-///  The running config is stored in a prefix tree. Each node in the tree represents 1 byte.
-///  Prefixes that share prefixes are effectively de-duped; their shared byte sized components only
-///  take up 1 byte. For example these 3 prefixes all have a common prefix of "/private/".
-///  They will only take up 21 bytes instead of 39.
+///  (the running config's current size) + (the prefix's size) totals up to more than 256 nodes.
+///  The running config is stored in a prefix tree.
+///  Prefixes that share prefixes are effectively de-duped; their shared node sized components only
+///  take up 1 node. For example these 3 prefixes all have a common prefix of "/private/".
+///  They will only take up 21 nodes instead of 39.
 ///
 ///  "/private/tmp/"
 ///  "/private/var/"
@@ -103,7 +103,7 @@
 ///                                                              -> [n] -> [e] -> [w] -> [/]
 ///
 ///  Prefixes with Unicode characters work similarly. Assuming a UTF-8 encoding these two prefixes
-///  are actually the same for the first 3 bytes. They take up 7 bytes instead of 10.
+///  are actually the same for the first 3 nodes. They take up 7 nodes instead of 10.
 ///
 ///  "/🤘"
 ///  "/🖖"
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -82,11 +82,41 @@
 ///  directory starting with "/private/tmp" would be ignored.
 ///
 ///  By default "/." and "/dev/" are added.
-///  Memory in the kernel is precious. 256 total bytes are allowed, use them wisely. Anything
-///  totaling up to more than 256 bytes will be ignored.
+///
+///  Memory in the kernel is precious. A total of 256 bytes are allowed for this configuration.
+///  An ASCII character uses 1 byte. An UTF-8 encoded Unicode character uses 1-4 bytes.
+///  Prefixes are added to the running config in-order one by one. The prefix will be ignored if
+///  (the running config's current size) + (the prefix's size) totals up to more than 256 bytes.
+///  The running config is stored in a prefix tree. Each node in the tree represents 1 byte.
+///  Prefixes that share prefixes are effectively de-duped; their shared byte sized components only
+///  take up 1 byte. For example these 3 prefixes all have a common prefix of "/private/".
+///  They will only take up 21 bytes instead of 39.
+///
+///  "/private/tmp/"
+///  "/private/var/"
+///  "/private/new/"
+///
+///                                                              -> [t] -> [m] -> [p] -> [/]
+///
+///  [/] -> [p] -> [r] -> [i] -> [v] -> [a] -> [t] -> [e] -> [/] -> [v] -> [a] -> [r] -> [/]
+///
+///                                                              -> [n] -> [e] -> [w] -> [/]
+///
+///  Prefixes with Unicode characters work similarly. Assuming a UTF-8 encoding these two prefixes
+///  are actually the same for the first 3 bytes. They take up 7 bytes instead of 10.
+///
+///  "/🤘"
+///  "/🖖"
+///
+///                          -> [0xa4] -> [0x98]
+///  [/] -> [0xf0] -> [0x9f]
+///                          -> [0x96] -> [0x96]
 ///
 ///  To disable file change logging completely add "/".
+///  TODO(bur): Make this default if no FileChangesRegex is set.
+///
 ///  Filters are only applied on santad startup.
+///  TODO(bur): Support add / remove of filters while santad is running.
 ///
 @property(readonly, nonatomic) NSArray *fileChangesPrefixFilters;
 
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -73,6 +73,23 @@
 @property(readonly, nonatomic) NSRegularExpression *fileChangesRegex;
 
 ///
+///  A list of ignore prefixes which are checked in-kernel.
+///  This is more performant than FileChangesRegex when ignoring whole directory trees.
+///
+///  For example adding a prefix of "/tmp/" will turn off file change log generation in-kernel
+//   for that entire tree. Since they are ignored by the kernel, they never reach santad and are
+///  not seen by the fileChangesRegex. Note the trailing "/", without it any file or directory
+///  starting with "/tmp" would be ignored.
+///
+///  By default "/." and "/dev/" are added.
+///  32 additional entries are allowed. Anything over 32 will be ignored.
+///
+///  To disable file change logging completely add "/".
+///  Filters are only applied on santad startup.
+///
+@property(readonly, nonatomic) NSArray *fileChangesPrefixFilters;
+
+///
 ///  Enable __PAGEZERO protection, defaults to YES
 ///  If this flag is set to NO, 32-bit binaries that are missing
 ///  the __PAGEZERO segment will not be blocked.
--- a/Source/santa-driver/SantaDecisionManager.cc
+++ b/Source/santa-driver/SantaDecisionManager.cc
@@ -58,16 +58,7 @@ bool SantaDecisionManager::init() {
   root_fsid_ = 0;
 
   // Setup file modification prefix filter.
-  filemod_prefix_filter_ = new SantaPrefixTree;
-  if (!filemod_prefix_filter_->init()) {
-    filemod_prefix_filter_->release();
-    filemod_prefix_filter_ = nullptr;
-    LOGE("Unable to initialize file modification filter tree.");
-  } else {
-    // Default prefixes to filter.
-    filemod_prefix_filter_->AddPrefix("/.");
-    filemod_prefix_filter_->AddPrefix("/dev/"); // Notice the trailing the "/".
-  }
+  filemod_prefix_filter_ = new SantaPrefixTree();
 
   return true;
 }
@@ -107,7 +98,7 @@ void SantaDecisionManager::free() {
   OSSafeReleaseNULL(decision_dataqueue_);
   OSSafeReleaseNULL(log_dataqueue_);
 
-  OSSafeReleaseNULL(filemod_prefix_filter_);
+  delete filemod_prefix_filter_;
 
   super::free();
 }
@@ -705,7 +696,7 @@ void SantaDecisionManager::FileOpCallback(
 
   // Filter out modifications to locations that are definitely
   // not useful or made by santad.
-  if (!filemod_prefix_filter_ || !filemod_prefix_filter_->HasPrefix(path)) {
+  if (!filemod_prefix_filter_->HasPrefix(path)) {
     auto message = NewMessage(nullptr);
     strlcpy(message->path, path, sizeof(message->path));
     if (new_path) strlcpy(message->newpath, new_path, sizeof(message->newpath));
--- a/Source/santa-driver/SantaDecisionManager.cc
+++ b/Source/santa-driver/SantaDecisionManager.cc
@@ -125,10 +125,6 @@ void SantaDecisionManager::ConnectClient(pid_t pid) {
   // connected should be cleared
   ClearCache();
 
-  // Reset the filter from any previous connections.
-  // The daemon will add prefixes as needed.
-  FilemodPrefixFilterReset();
-
   failed_decision_queue_requests_ = 0;
   failed_log_queue_requests_ = 0;
 }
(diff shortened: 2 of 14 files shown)


Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -0,0 +1,147 @@
...
+
+  SantaPrefixNode *node = root_;
+  size_t len = strlen(prefix);
+  for (int i = 0; i < len; ++i) {
+    // If there is a node in the path that is considered a prefix, stop adding.
+    // For our purposes we only care about the shortest path that matches.
+    if (node->isPrefix) break;
+
+    // Create a single char string.
+    const char value[2] = {prefix[i], '\0'};
```
russellhancox: Is this a string just to satisfy OSDictionary?

Source/santa-driver/SantaPrefixTree.h
```diff
@@ -0,0 +1,67 @@
...
+class SantaPrefixNode : public OSObject {
+  OSDeclareDefaultStructors(SantaPrefixNode);
+
+ public:
+  bool init() override;
+  void free() override;
+
+  bool isPrefix;
+  // TODO(bur): After writing this I noticed OSDictionary has an O(n) runtime.
+  //            Switch to an int[256] lookup table and get rid of SantaPrefixNode.
```
russellhancox: I'd do this to start with as changing away from OSDictionary allows a bunch of simplifications elsewhere (you can nest the classes, do away with the OSDeclare* macros, use single chars instead of 'string' keys, etc.)

Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -0,0 +1,147 @@
...
+
+bool SantaPrefixNode::init() {
+  if (!super::init()) return false;
+  children = OSDictionary::withCapacity(1);
+  isPrefix = false;
+  return true;
+}
+
+void SantaPrefixNode::free() {
+  children->OSCollection::iterateObjects(^bool(OSObject *key) {
```
russellhancox: This shouldn't, I believe, be necessary. All the OSCollection classes retain their keys and values when added and release when removed - as long as the free method calls release on the collection all of the objects in that collection will be released, cascading down.

Source/santa-driver/SantaDecisionManager.h
```diff
@@ -164,6 +164,10 @@ class SantaDecisionManager : public OSObject {
   */
   bool IsCompilerProcess(pid_t pid);
 
+  void AddFilemodPrefixFilter(const char *prefix) {
```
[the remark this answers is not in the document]
russellhancox: This isn't strictly necessary (it'll be inlined regardless) but it makes it clearer and consistent with the surrounding code.

Source/santa-driver/SantaPrefixTree.cc
tburgin: Without it the nodes are leaked :/
```
⇒  sudo kextunload -b com.google.santa-driver
(kernel) Can't unload kext com.google.santa-driver; classes have instances:
(kernel)     Kext com.google.santa-driver class SantaPrefixNode has 29 instances.
Failed to unload com.google.santa-driver - (libkern/kext) kext is in use or retained (cannot unload).
```
refcount = 0
`SantaPrefixNode *newNode = new SantaPrefixNode;`
refcount = 1
`node->children->setObject(value, newNode);`
refcount = 2

To release them:

`children->getObject((OSString *)key)->release();`
refcount = 1
`children->flushCollection();`
refcount = 0

The docs for `flushCollection` say:
>Removes and releases all keys and objects within the dictionary.
I take this as, "call release()" on all keys and objects.

Anyways, removing this code in favor of the char lookup array.

russellhancox: That's because you're not releasing them after adding to the dictionary.

```
SantaPrefixNode *newNode = new SantaPrefixNode;  // newNode refcount = 1
children->setObject(value, newNode);  // newNode refcount = 2
newNode->release()  // newNode refcount = 1

...

children->release();  // newNode refcount = 0
```

tburgin: Ah, thats better.

Source/santa-driver/SantaPrefixTree.h
nguyen-phillip: BTW, if going this route, using half-byte keys instead of chars would reduce the space requirements by a factor of 8, because each level then only needs an array of 2^4 = 16 pointers instead of 2^8 = 256 pointers, while the effective length of the prefix doubles.

Source/santad/SNTDriverManager.m
```diff
@@ -278,4 +278,31 @@ - (NSArray *)cacheBucketCount {
...
+  // doesn't have to acquire locks etc...
+  size_t count = 0;
+  for (NSString *filter in filters) {
+    char buffer[MAXPATHLEN];
+    if (![filter getFileSystemRepresentation:buffer maxLength:MAXPATHLEN]) {
+      LOGE(@"Invalid filemod prefix filter: %@", filter);
+      continue;
+    }
+
+    count += strlen(buffer);
```
russellhancox: Does strlen count UTF-8 strings correctly?

Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -0,0 +1,130 @@
...
+  lck_rw_unlock_exclusive(spt_lock_);
+}
+
+bool SantaPrefixTree::HasPrefix(const char *string) {
+  lck_rw_lock_shared(spt_lock_);
+
+  auto found = false;
+
+  SantaPrefixNode *node = root_;
+  size_t len = strlen(string);
```
nguyen-phillip: Using strlen forces you to loop over the string twice.  An alternative is something like:

for (char *p = string, node = root_; node && *p && !node->isPrefix; node = node->children[*p++]);
found = node && node->isPrefix;

Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -0,0 +1,130 @@
...
+
+    // Get ready for the next iteration.
+    node = (SantaPrefixNode *)node->children[value];
+
+    // If this is the end, mark the node as a prefix.
+    if (i + 1 == len) {
+      LOGD("Added prefix: %s", prefix);
+      node->isPrefix = true;
+      break;  // Unnecessary, but clear.
+    }
```
nguyen-phillip: This can go outside the for loop: the last value of node will be the end of the prefix.  (Though should probably check for zero-length prefix at beginning of function and return early if yes.) Saves a comparison each loop.

Source/santa-driver/SantaPrefixTree.h
```diff
@@ -0,0 +1,82 @@
...
+  ///
+  ///  Lookups of children are O(1).
+  ///
+  ///  Having the nodes represented by a smaller width, such as a nibble (1/2 byte), would
+  ///  drastically decrease the memory footprint but would double required dereferences.
+  ///
+  class SantaPrefixNode {
+   public:
+    bool isPrefix;
+    void *children[256];
```
nguyen-phillip: It's probably not worth the 9 bytes that you'd save per node, but children[0] will never be used, so you could declare an array of length 255 and adjust the indexing.

Source/santa-driver/SantaPrefixTree.h
```diff
@@ -0,0 +1,82 @@
...
+        children[i] = 0;
+      }
+    }
+  };
+
+  SantaPrefixNode *root_;
+
+  // Each node takes up ~2k, max out at ~512k.
+  static const uint32_t kMaxNodes = 256;
+  uint32_t current_nodes_;
```
nguyen-phillip: suggest renaming this to something involving count instead of current.

Source/santa-driver/SantaPrefixTree.h
tburgin: Ack

Source/santa-driver/SantaPrefixTree.cc
tburgin: Added a TODO

Source/santa-driver/SantaPrefixTree.h
```diff
@@ -0,0 +1,82 @@
...
+  ///  Lookups of children are O(1).
+  ///
+  ///  Having the nodes represented by a smaller width, such as a nibble (1/2 byte), would
+  ///  drastically decrease the memory footprint but would double required dereferences.
+  ///
+  class SantaPrefixNode {
+   public:
+    bool isPrefix;
+    void *children[256];
+
```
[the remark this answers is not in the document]
tburgin: As far as I know, yep.

Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -0,0 +1,238 @@
...
+  // Seed the "stack" with a starting node.
+  auto target_pointer = OSNumber::withNumber((uint64_t)(void *)target, 64);
+  stack->setObject(target_pointer);
+  target_pointer->release();
+
+  // Start at the target node and walk the tree to find all the sub-nodes.
+  while (stack->getCount()) {
+    auto pointer = OSDynamicCast(OSNumber, stack->getLastObject());
+    if (!pointer) {
+      LOGE("Unable complete prune!");
```
russellhancox: missing to

Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -0,0 +1,238 @@
...
+      stack->setObject(child_pointer);
+      child_pointer->release();
+    }
+  }
+
+  // Delete all the seen nodes.
+  seen->iterateObjects(^bool(OSObject *object) {
+    auto pointer = OSDynamicCast(OSNumber, object);
+    if (!pointer) {
+      LOGE("Unable delete node!");
```
russellhancox: missing to

Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -0,0 +1,238 @@
...
+    --node_count_;
+    return false;
+  });
+
+  OSSafeReleaseNULL(stack);
+  OSSafeReleaseNULL(seen);
+}
+
+SantaPrefixTree::~SantaPrefixTree() {
+  // The locking is probably not necessary here, but I am scared to remove it.
```
russellhancox: The locking is definitely necessary - while it might not currently be possible for AddPrefix or HasPrefix to be called while this object is being destroyed it's very easy for that to be changed without realizing it could lead to crashes and they may not appear immediately.

Source/santa-driver/SantaPrefixTree.h
```diff
@@ -0,0 +1,86 @@
...
+  // Check if the tree has a prefix for string.
+  bool HasPrefix(const char *string);
+
+  // Reset the tree.
+  // Optionally pass node_count to get the number of nodes after the reset. It should be 1.
+  void Reset(uint32_t *node_count = nullptr);
+
+  SantaPrefixTree();
+  ~SantaPrefixTree();
```
russellhancox: You don't need to declare these

Source/santa-driver/SantaPrefixTree.h
```diff
@@ -0,0 +1,86 @@
...
+  ///
+  ///  TODO(bur): Potentially convert this into a full on radix tree.
+  ///
+  class SantaPrefixNode {
+   public:
+    bool isPrefix;
+    void *children[256];
+  };
+
+  // PruneNode will remove the passed in node from the tree.
```
russellhancox: The node and all of its children are deleted but nothing sets the pointer in the parent's children array to nullptr. You should call this out.

Source/santa-driver/SantaDriverClient.cc
```diff
@@ -248,6 +248,28 @@ IOReturn SantaDriverClient::cache_bucket_count(
...
+  SantaDriverClient *me = OSDynamicCast(SantaDriverClient, target);
+  if (!me) return kIOReturnBadArgument;
+
+  const char *prefix = reinterpret_cast<const char *>(arguments->structureInput);
+
+  IOReturn ret = kIOReturnSuccess;
+  uint32_t node_count = 0;
+
+  // If the client sends a bad address the kernel will kill the client before it reaches here.
+  if (prefix[0] == '\0') {
```
russellhancox: I don't like the empty string being used as a magic value to force a reset - I think the empty string should be ignored (and probably logged) and another client method should be added to do a reset.

Source/santa-driver/SantaDecisionManager.cc
```diff
@@ -57,6 +57,9 @@ bool SantaDecisionManager::init() {
   client_pid_ = 0;
   root_fsid_ = 0;
 
+  // Setup file modification prefix filter.
+  filemod_prefix_filter_ = new SantaPrefixTree();
```
russellhancox: I think you should add a reset call in ConnectClient() to avoid leaving stale prefixes.

Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -0,0 +1,238 @@
...
+      if (!lck_rw_lock_shared_to_exclusive(spt_lock_)) {
+        // Grab a new exclusive lock.
+        lck_rw_lock_exclusive(spt_lock_);
+      }
+
+      // Is there enough room for the rest of the prefix?
+      if (!branched && (node_count_ + (len - i)) > kMaxNodes) {
+        LOGE("Prefix tree is full, can not add: %s", prefix);
+        if (node_count) *node_count = node_count_;
+        lck_rw_unlock_exclusive(spt_lock_);
```
russellhancox: Unless I'm mistaken[1] there's a potential threading bug here where 2 threads are adding prefixes at the same time and the count goes over kMaxNodes but part of a prefix has been added - it's not a huge problem as the partial prefix won't end with isPrefix=true so it'll never actually trigger anything but the nodes will be taking up space in the tree.

1: I'm not sure but the branched flag _might_ prevent this

Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -0,0 +1,238 @@
...
+
+  lck_rw_unlock_exclusive(spt_lock_);
+}
+
+void SantaPrefixTree::PruneNode(SantaPrefixNode *target) {
+  if (!target) return;
+
+  // For deep trees a recursive approach will generate too many stack frames.
+  auto stack = OSArray::withCapacity(1);
+  auto seen = OSSet::withCapacity(1);
```
russellhancox: Start with the capacity of node_count_ - you might not need all that space as you might be starting part way down the tree but better that then to keep having to re-double the size.

Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -0,0 +1,238 @@
...
+  // Start at the target node and walk the tree to find all the sub-nodes.
+  while (stack->getCount()) {
+    auto pointer = OSDynamicCast(OSNumber, stack->getLastObject());
+    if (!pointer) {
+      LOGE("Unable complete prune!");
+      break;  // Bail walking the tree, but still delete any seen nodes.
+    }
+
+    seen->setObject(pointer);
+    stack->removeObject(stack->getCount() - 1);
```
russellhancox: This might be premature - when you created this OSNumber object and added it to stack you released it, so the stack has the only reference to it. If you remove it from the stack the reference count should drop to 0 and be freed.

Source/common/SNTConfigurator.h
```diff
@@ -72,6 +72,54 @@
...
+///  By default "/." and "/dev/" are added.
+///
+///  Memory in the kernel is precious. A total of 256 bytes are allowed for this configuration.
+///  An ASCII character uses 1 byte. An UTF-8 encoded Unicode character uses 1-4 bytes.
+///  Prefixes are added to the running config in-order one by one. The prefix will be ignored if
+///  (the running config's current size) + (the prefix's size) totals up to more than 256 bytes.
+///  The running config is stored in a prefix tree. Each node in the tree represents 1 byte.
+///  Prefixes that share prefixes are effectively de-duped; their shared byte sized components only
+///  take up 1 byte. For example these 3 prefixes all have a common prefix of "/private/".
+///  They will only take up 21 bytes instead of 39.
```
nguyen-phillip: Should this be 21 nodes instead of 21 bytes?  Otherwise it seems a bit misleading since each node is actually  1 byte for bool + 256 * 8 bytes for pointer array = 2 kilobytes

Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -0,0 +1,238 @@
...
+    if (!node->children[value]) {
+      // Upgrade the shared lock.
+      // If the upgrade fails, the shared lock is released.
+      if (!lck_rw_lock_shared_to_exclusive(spt_lock_)) {
+        // Grab a new exclusive lock.
+        lck_rw_lock_exclusive(spt_lock_);
+      }
+
+      // Is there enough room for the rest of the prefix?
+      if (!branched && (node_count_ + (len - i)) > kMaxNodes) {
```
nguyen-phillip: Once you've branched, what is the reason for continuing to check the size?

Source/santa-driver/SantaPrefixTree.h
```diff
@@ -0,0 +1,86 @@
...
+  ///
+  ///  Having the nodes represented by a smaller width, such as a nibble (1/2 byte), would
+  ///  drastically decrease the memory footprint but would double required dereferences.
+  ///
+  ///  TODO(bur): Potentially convert this into a full on radix tree.
+  ///
+  class SantaPrefixNode {
+   public:
+    bool isPrefix;
+    void *children[256];
```
nguyen-phillip: Maybe I'm missing something, but I think this can be declared as SantaPrefixNode *children[256];

Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -0,0 +1,238 @@
...
+    seen->setObject(pointer);
+    stack->removeObject(stack->getCount() - 1);
+
+    auto node = (SantaPrefixNode *)pointer->unsigned64BitValue();
+    for (int i = 0; i < 256; ++i) {
+      if (!node->children[i]) continue;
+      auto child_pointer = OSNumber::withNumber((uint64_t)node->children[i], 64);
+      stack->setObject(child_pointer);
+      child_pointer->release();
+    }
```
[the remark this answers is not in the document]
tburgin: Brilliant!

Source/santa-driver/SantaPrefixTree.cc
tburgin: Looking into it.

Source/santa-driver/SantaPrefixTree.cc
tburgin: I was adding it seen so it was retained. Got rid of seen, so it can now free the OSNumber as soon as it gets its value.

Source/santa-driver/SantaPrefixTree.h
russellhancox: Oh, of course, you're not defining them here. Carry on.

Source/santa-driver/SantaDecisionManager.cc
russellhancox: If you add a reset in ConnectClient you get 2 benefits: 1) the tests won't break if you don't unload before running them and don't have to add reset calls to the test, 2) You can remove the reset on santad startup and only use the reset call when the config changes from when santad first starts.

Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -0,0 +1,224 @@
...
+
+#include "SNTLogging.h"
+
+#include <libkern/c++/OSArray.h>
+#include <libkern/c++/OSNumber.h>
+#include <libkern/c++/OSSet.h>
+
+SantaPrefixTree::SantaPrefixTree() {
+  root_ = new SantaPrefixNode();
+  ++node_count_;
```
russellhancox: Explicitly set this to 1.

Source/santa-driver/SantaDecisionManager.cc
tburgin: Done.

Source/santa-driver/SantaPrefixTree.cc
tburgin: I think I fixed this by adding the rest of the prefix to the tree when branching.

Source/santa-driver/SantaPrefixTree.cc
tburgin: Thanks, updated with something similar.

tburgin:
PTAL - I ran clang-format on KernelTests/main.mm, sorry!

Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -0,0 +1,220 @@
...
+  if (node_count) *node_count = node_count_;
+
+  lck_rw_unlock_exclusive(spt_lock_);
+}
+
+void SantaPrefixTree::PruneNode(SantaPrefixNode *target) {
+  if (!target) return;
+
+  // For deep trees a recursive approach will generate too many stack frames.
+  auto stack = OSArray::withCapacity(node_count_);
```
nguyen-phillip: If node_count_ is accurate I _think_ it's impossible for the stack to ever have more than (node_count_+1) items.  If that's right, then this can be simplified by declaring the stack as 

SantaPrefixNode *stack[node_count_+1]; 
int count = 0;

with push as stack[count++] = item, and pop as item = stack[--count].

You'd lose the dynamic resizing of OSArray, but I don't think it's ever resizing anyway since it's already being initialized with a capacity of node_count_ (which should maybe have been +1 to include root_?).  The advantage is no more casting and no more wrapping up pointers in OSNumber objects, which would drop the memory usage by at least half.

Source/santa-driver/SantaPrefixTree.h
```diff
@@ -0,0 +1,88 @@
...
+ public:
+  // Add a prefix to the tree.
+  // Optionally pass node_count to get the number of nodes after the add.
+  IOReturn AddPrefix(const char *, uint64_t *node_count = nullptr);
+
+  // Check if the tree has a prefix for string.
+  bool HasPrefix(const char *string);
+
+  // Reset the tree.
+  // Optionally pass node_count to get the number of nodes after the reset. It should be 1.
```
nguyen-phillip: 1 --> 0?

Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -0,0 +1,220 @@
...
+      break;
+    }
+  }
+
+  lck_rw_unlock_shared(spt_lock_);
+
+  return found;
+}
+
+void SantaPrefixTree::Reset(uint64_t *node_count) {
```
nguyen-phillip: Is it actually possible for this function to return and *node_count not be zero?  Or is this for implementing some kind of wait on the value of *node_count?

Source/santa-driver/SantaPrefixTree.cc
tburgin: Good point. I think this idea should work, but I will have to allocate the the space for the "stack" on the heap. The default kernel stack size is 16k. Putting the "stack" on the stack would take up ~8k. Not a lot of wiggle room, especially if there are a lot of previous frames.

Source/santa-driver/SantaPrefixTree.cc
tburgin: This was left over from testing. Removed.

Source/santa-driver/SantaPrefixTree.h
tburgin: Removed

PR #168, opened 2017-04-20
santactl/sync: Sync Server to set FCM interval and deadline

tburgin:
*  Handle `fcm_full_sync_interval` and `fcm_global_rule_deadline` values during preflight
*  If there is no SyncBaseURL exit
*  Remove support for `backoff`

PR #317, opened 2018-11-08
Paragraph formatting #157 to #168 (proposal)

kant:


russellhancox:
Thank you for your contribution!

PR #323, opened 2018-11-19
document vnode_pid_map

tburgin:
* remove vnode_pid_map: It is only used for logging execs. Creating a new message will get the executing PID and PPID.
* fix logic tests under Xcode 10.1


russellhancox:
From the [commit](bf51049fbf439ed265284a6be49a3a1bb0761f81) that introduced vnode_pid_map:

> Previously the execution logging from fileop didn't work when using posix_spawn as proc_selfpid/proc_selfppid still refer to the process calling posix_spawn. We can get the correct pid/ppid from the vfs_context in the vnode scope but we can't log executions from there as the arguments end up being wrong. Instead, save the vnode_id->pid/ppid mapping in the vnode scope and use that in the fileop scope for logging.

tburgin:
Ah, that makes sense. Added a note so I don't forget again.

PR #324, opened 2018-11-27
 SantaPrefixTree: Fix a bug and add some more tests

tburgin:
* Serialize add prefix requests.
* Add a bunch of ugly macros to make SantaPrefixTree usable in userspace.

### DIFF
--- a/Source/santa-driver/SantaPrefixTree.cc
+++ b/Source/santa-driver/SantaPrefixTree.cc
@@ -14,26 +14,67 @@
 
 #include "SantaPrefixTree.h"
 
-#include "SNTLogging.h"
+#ifdef KERNEL
+#include <libkern/locks.h>
 
-SantaPrefixTree::SantaPrefixTree() {
+#include "SNTLogging.h"
+#else // KERNEL
+// Support for unit testing.
+// Requires c++17 / macOS 10.12.
+// TODO(bur): Handle warnings from bumping target version of the tests to 10.12.
+#include <shared_mutex>
+
+#define lck_grp_attr_alloc_init() nullptr
+#define lck_grp_alloc_init(name, attr) nullptr
+#define lck_attr_alloc_init() nullptr
+
+#define lck_attr_free(attr) do {} while(0)
+#define lck_grp_free(grp) do {} while(0)
+#define lck_grp_attr_free(grp_attr) do {} while(0)
+
+#define lck_rw_lock_shared(l) l.lock_shared()
+#define lck_rw_unlock_shared(l) l.unlock_shared()
+#define lck_rw_lock_exclusive(l) l.lock()
+#define lck_rw_unlock_exclusive(l) l.unlock()
+
+#define lck_rw_lock_shared_to_exclusive(l) [this] () { l.unlock_shared(); return false;}()
+#define lck_rw_lock_exclusive_to_shared(l) l.unlock(); l.lock_shared()
+
+#define lck_mtx_lock(l) l.lock()
+#define lck_mtx_unlock(l) l.unlock()
+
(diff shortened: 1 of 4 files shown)


Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -14,26 +14,61 @@
...
   spt_lock_ = lck_rw_alloc_init(spt_lock_grp_, spt_lock_attr_);
+  spt_add_lock_ = lck_mtx_alloc_init(spt_lock_grp_, spt_lock_attr_);
+  #endif
 }
 
 IOReturn SantaPrefixTree::AddPrefix(const char *prefix, uint64_t *node_count) {
+  // Serialize requests to AddPrefix. Otherwise one AddPrefix thread could overwrite whole
+  // branches of another. HasPrefix is still free to read the tree, until AddPrefix needs to
+  // modify it.
+  lck_mtx_lock(spt_add_lock_);
```
nguyen-phillip: Since AddPrefix is now serialized, is it possible to reduce/simplify the amount of locking in the rest of this function?  (It may not be, but I thought I'd ask.)

tburgin: We could simplify it by grabbing the exclusive rw lock during the whole duration of AddPrefix, but that would prevent reads while we figure out if the tree needs to be modified. The extra complexity feels worth it to keep the reads flowing.

Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -14,26 +14,61 @@
...
+SantaPrefixTree::SantaPrefixTree(uint32_t max_nodes) {
   root_ = new SantaPrefixNode();
   node_count_ = 0;
+  max_nodes_ = max_nodes;
 
   spt_lock_grp_attr_ = lck_grp_attr_alloc_init();
   spt_lock_grp_ = lck_grp_alloc_init("santa-prefix-tree-lock", spt_lock_grp_attr_);
   spt_lock_attr_ = lck_attr_alloc_init();
+
+  #ifdef KERNEL
```
russellhancox: Could you avoid these ifdef's by defining lck_rw_alloc_init, lck_rw_t, lck_mtx_t, etc.?

```cpp
#define lck_rw_t std::shared_mutex
#define lck_mtx_t std::mutex

#define lck_rw_alloc_init(a, b) new std::shared_mutex
#define lck_rw_lock_shared(l) l->lock_shared()
...

#define lck_mtx_alloc_init(a, b) new std::mutex
#define lck_mtx_lock(l) l->lock();
```

tburgin: The templates get angry: `No viable overloaded '='`

Source/santa-driver/SantaPrefixTree.cc
```diff
@@ -14,26 +14,61 @@
...
 IOReturn SantaPrefixTree::AddPrefix(const char *prefix, uint64_t *node_count) {
+  // Serialize requests to AddPrefix. Otherwise one AddPrefix thread could overwrite whole
+  // branches of another. HasPrefix is still free to read the tree, until AddPrefix needs to
+  // modify it.
+  lck_mtx_lock(spt_add_lock_);
+
   // Don't allow an empty prefix.
   if (prefix[0] == '\0') return kIOReturnBadArgument;
 
+  #ifdef KERNEL
```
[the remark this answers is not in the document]
tburgin: smwart

Source/santa-driver/SantaPrefixTree.cc
tburgin: ack

Issue #303, opened 2018-09-26
Standardize build system


russellhancox:
Santa currently has 3 build systems:

1. The original Xcode project - this is still officially the right way to edit and build the project
2. The CMake build rules added in #284, which was largely to support fuzzing parts of Santa
3. A Bazel BUILD file, which is used internally at Google for building releases because that allows some automation for us.

Maintaining all 3 of these is a nuisance as any changes now have to be repeated in all 3 systems. I think the obvious answer is to settle on using Bazel everywhere and delete CMake/Xcode files from the repository. [Tulsi](https://tulsi.bazel.build/) can be used to create an Xcode project that uses Bazel for the actual build and the fuzzing rules in CMake can be ported to Bazel too. The other nice thing is that Bazel WORKSPACE files can be used to import external Git repositories without having to mess around with submodules.

tburgin:
+1
Once https://github.com/bazelbuild/rules_apple/commit/d197cbb59d445373beb966d4ba509e5217517465 is merged. It will make the move to bazel pretty seamless.

russellhancox:
This is almost complete, just the fuzzing rules to finish.

russellhancox:
I also need to split these rules out, the BUILD file is over 500 lines because it contains everything when really there should be separate BUILD files for separate components.

The fuzzing rules are taking longer than I expected because of mixing objc and cc code and targeting macOS. I'm working on some helper rules for the tests to make them clearer.

pmarkowsky:
There's also https://github.com/bazelbuild/rules_fuzzing now.  Though looks like macos doesn't work out of the box.

russellhancox:
Opened #815 to track getting the fuzzing rules added, closing this.

PR #326, opened 2018-12-13
Proj: Convert to Bazel build, remove other build systems.

russellhancox:
This necessitated fixing some warnings and updating the resource locations inside some tests.

Once this is submitted I have some follow-ups to do:

1) Add targets for the fuzz tests. This is a bit fiddly due to the way libFuzzer links
2) Possibly split the BUILD file into several BUILD files. This would make the BUILD files easier to read and easier to have smaller targets that can be pulled in (inside of several huge globs) but it makes versioning a little bit more difficult.

Part of #303


nguyen-phillip:
This seems to have broken Travis

russellhancox:
It took a few attempts but that's now fixed too

PR #329, opened 2019-01-03
Project: Convert to bazel, part 2

russellhancox:
The main thing to call out in this PR is that Santa.app is now embedded inside santa-driver.kext along with everything else. The package will handle updating this automatically but it should be called out in release notes to make everyone aware.

* Switch to using macos_kernel_extension and macos_xpc_service, stop using product_type.
* Have Bazel embed all related binaries inside santa-driver.kext, including Santa.app. This simplifies the :release and :reload rules.
* Add commands for unload, load and reload, removing any need to keep the Rakefile around 😃
* Make the :kernel_tests rule a command that replicates what the Rakefile did for this.
* Added a project-wide .bazelrc that always generates dSYMs

Documentation changes to follow, as all the building instructions are now out of date.


BUILD
```diff
@@ -253,44 +258,71 @@ macos_application(
...
+sudo cp $${BUILD_WORKSPACE_DIRECTORY}/conf/com.google.santa.asl.conf /etc/asl
+sudo cp $${BUILD_WORKSPACE_DIRECTORY}/conf/com.google.santa.newsyslog.conf /etc/newsyslog.d/
+sudo /usr/bin/killall -HUP syslogd
+
+echo "Loading new Santa"
+sudo chown -R root:wheel /Library/Extensions/santa-driver.kext
+sudo launchctl load /Library/LaunchDaemons/com.google.santad.plist
+launchctl load /Library/LaunchAgents/com.google.santagui.plist
+
+echo "Time to stop being naughty"
```
tburgin: 😜

cmd.bzl
```diff
@@ -0,0 +1,12 @@
+def run_command(name, cmd, **kwargs):
```
tburgin: Clever

BUILD
```diff
@@ -253,44 +258,57 @@ macos_application(
...
+    name = "reload",
+    cmd = """
+set -e
+
+unzip -o -d /tmp/bazel_santa_reload \
+    $${BUILD_WORKSPACE_DIRECTORY}/bazel-bin/santa-driver.zip >/dev/null
+echo "You may be asked for your password for sudo"
+sudo BINARIES=/tmp/bazel_santa_reload CONF=$${BUILD_WORKSPACE_DIRECTORY}/Conf \
+    $${BUILD_WORKSPACE_DIRECTORY}/Conf/install.sh
+rm -rf /tmp/bazel_santa_reload
```
tburgin: I would move this above the unzip incase this the install.sh is interrupted.

PR #330, opened 2019-01-04
Project: Update docs for switch to bazel

russellhancox:
I also made it so the santactl compilation won't include debug-only commands in release builds and the release rule will fail on a non-opt build.


russellhancox:
Ptal

PR #336, opened 2019-01-16
BUILD: buildifier formatting

tburgin:
What do you think about adding buildifier to Santa's WORKSPACE?

https://github.com/bazelbuild/buildtools/tree/master/buildifier

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -7,8 +7,8 @@ exports_files(["LICENSE"])
 load(
     "@build_bazel_rules_apple//apple:macos.bzl",
     "macos_application",
-    "macos_kernel_extension",
     "macos_command_line_application",
+    "macos_kernel_extension",
     "macos_unit_test",
     "macos_xpc_service",
 )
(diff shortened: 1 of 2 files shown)


russellhancox:
> What do you think about adding buildifier to Santa's WORKSPACE?

I dunno, seems a bit overkill given how much needs to be added and maintained. Let's come back to that if we keep having to submit buildifier fixes 😄

BUILD
```diff
@@ -177,11 +176,11 @@ objc_library(
     sdk_dylibs = ["libz"],
     sdk_frameworks = ["IOKit"],
     deps = [
-        "@FMDB//:fmdb",
-        "@MOLAuthenticatingURLSession//:MOLAuthenticatingURLSession",
-        "@MOLCodesignChecker//:MOLCodesignChecker",
-        "@MOLFCMClient//:MOLFCMClient",
-        "@MOLXPCConnection//:MOLXPCConnection",
+        "@FMDB",
```
russellhancox: This fails - the rule inside the FMDB package is fmdb not FMDB.

tburgin: Did you see the change to the WORKSPACE? It works for me...and Travis.

russellhancox: Oh the WORKSPACE change, nevermind.

tburgin:
>Let's come back to that if we keep having to submit buildifier fixes.
Agreed

PR #338, opened 2019-01-18
Project: Split BUILD file into several. Part 1/3

russellhancox:
The tests fail in this PR because the rules need updating. I'll fix them in a follow-up PR.


Source/SantaGUI/BUILD
```diff
@@ -0,0 +1,51 @@
...
+    xibs = [
+        "Resources/AboutWindow.xib",
+        "Resources/MessageWindow.xib",
+    ],
+    deps = [
+        "//Source/common:SNTBlockMessage",
+        "//Source/common:SNTConfigurator",
+        "//Source/common:SNTXPCControlInterface",
+        "//Source/common:SNTXPCNotifierInterface",
+        "@MOLCodesignChecker//:MOLCodesignChecker",
```
tburgin: nit: @MOLCodesignChecker
Maybe just run buildifier on all the BUILD files?

russellhancox: Done

russellhancox:
I was in two minds about the glob() thing - I like being quite specific rather than relying on glob but the thing that pushed me into avoiding it is that I'm moving the test source files to live with the libs they're testing and keeping the exclusion list to exclude those is annoying.

Tests/LogicTests/BUILD
```diff
@@ -0,0 +1,53 @@
...
+        "DiskArbitration",
+        "IOKit",
+        "SecurityInterface",
+    ],
+    structured_resources = glob([
+        "Tests/LogicTests/Resources/BundleExample.app/**",
+        "Tests/LogicTests/Resources/DirectoryBundle/**",
+    ]),
+    deps = [
+        "@FMDB//:fmdb",
```
tburgin: @FMDB

tburgin:
Ack, makes sense.

PR #347, opened 2019-02-15
project format

tburgin:
* Use buildifier to format .bzl and BUILD files
* Rename Source/santa-driver -> Source/santa_driver
* Allow to kernel_tests binary to unload the driver after tests


BUILD
```diff
@@ -43,13 +43,13 @@ launchctl load /Library/LaunchAgents/com.google.santagui.plist
 
 run_command(
     name = "reload",
-    srcs = ["//Source/santa-driver"],
+    srcs = ["//Source/santa_driver:santa-driver"],
```
russellhancox: I'd suggest renaming the rule too, so you don't need to specify it.

Source/santa_driver/BUILD
```diff
@@ -121,11 +121,10 @@ function sigint() {
   sudo kextunload -b com.google.santa-driver >/dev/null
   exit 1
 }
-unzip -o $${BUILD_WORKSPACE_DIRECTORY}/bazel-bin/Source/santa-driver/santa-driver.zip >/dev/null
+unzip -o $${BUILD_WORKSPACE_DIRECTORY}/bazel-bin/Source/santa_driver/santa-driver.zip >/dev/null
 echo "Launching Kernel Tests as root. You may be prompted for your sudo password."
 trap sigint INT
-sudo $${BUILD_WORKSPACE_DIRECTORY}/bazel-bin/Source/santa-driver/kernel_tests_bin
-echo "Tests complete, unloading driver."
-sudo kextunload -b com.google.santa-driver >/dev/null
```
russellhancox: I had this as a fallback in case the tests crash or the user cancels out of them because they've hung.

BUILD
tburgin: Done.

Source/santa_driver/BUILD
tburgin: Ack

PR #355, opened 2019-02-22
santad/santactl: Remove transitive whitelisting preference.

russellhancox:
Infer whether transitive whitelisting should be enabled based on whether there are any compiler rules installed. If a sync server sent a compiler rule (or one was manually added because there is no sync server) it follows that transitive whitelisting should be enabled.

### DIFF
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -249,15 +249,6 @@
 ///
 @property BOOL enableBundles;
 
-#pragma mark Transitive Whitelisting Settings
-
-///
-///  If YES, binaries marked with SNTRuleStateWhitelistCompiler rules are allowed to transitively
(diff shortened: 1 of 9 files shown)


nguyen-phillip:
Seems fine to me, but do we not want a way to flip transitive whitelisting on/off independently of having compiler rules installed?  Is the idea that the compiler rules would be removed/replaced with regular rules by the sync server if transitive whitelisting is disabled (after having been enabled)?

tburgin:
Good point. What if we stored the hash of the compiler along with the generated transitive rule? We could then easily reap any transitive rules created by the compiler. For instance when a sync server sends down a REMOVE for a compiler hash, we would also cleanup the transitive rules.

russellhancox:
I'd forgotten about that and I don't like the idea of forcing the sync server to have to do that much work, so I'm retracting this.

PR #370, opened 2019-06-23
Use updated deps

tburgin:
* bazel 0.27.0 defaults to python 3 - added a flag to .bazelrc to force python 2


tburgin:
Yep, by default the build will fail if python 3 is not installed.

tburgin:
PTAL
Seems like the default 10.13 travis image has a stale version of homebrew. Updated our config to use their 10.14 image.

PR #387, opened 2019-09-06
* Added Xcode project + pods

tburgin:
* Cleaned up unused SNTXPCUnprivilegedControlInterface MachServices id.
* Change santad's MachServices id to be compatible with the default SystemExtension namespace template.


Source/common/SNTXPCControlInterface.m
```diff
@@ -15,19 +15,22 @@
 #import "Source/common/SNTXPCControlInterface.h"
 
 #import <MOLXPCConnection/MOLXPCConnection.h>
+#import <MOLCodesignChecker/MOLCodesignChecker.h>
```
russellhancox: Re-order

Source/common/SNTXPCControlInterface.m
```diff
@@ -15,19 +15,22 @@
...
 #import "Source/common/SNTRule.h"
 #import "Source/common/SNTStoredEvent.h"
 
 @implementation SNTXPCControlInterface
 
 + (NSString *)serviceId {
-  return @"SantaXPCControl";
+  MOLCodesignChecker *cs = [[MOLCodesignChecker alloc] initWithSelf];
+  // "teamid.com.google.santa.daemon.xpc" when signed.
+  // "com.google.santa.daemon.xpc" when not signed.
```
russellhancox: How does this work? santad will ~always be signed but the launchd.plist assumes it isn't? Shouldn't this logic be "if >10.15 return teamid.*, otherwise return *?"

Source/santa/Resources/santa-info.plist
```diff
@@ -2,31 +2,31 @@
...
-	<string>${SANTA_VERSION}</string>
+	<string>$(MARKETING_VERSION)</string>
+	<key>CFBundleVersion</key>
+	<string>$(CURRENT_PROJECT_VERSION)</string>
 	<key>LSMinimumSystemVersion</key>
 	<string>${MACOSX_VERSION_MIN}</string>
 	<key>LSUIElement</key>
 	<true/>
+	<key>NSHumanReadableCopyright</key>
+	<string>Google, Inc.</string>
```
russellhancox: Google LLC.

Source/santa_driver/Info.plist
```diff
@@ -26,16 +24,18 @@
...
 			<key>IOResourceMatch</key>
 			<string>IOKit</string>
 			<key>IOUserClientClass</key>
 			<string>com_google_SantaDriverClient</string>
-			<key>IOMatchCategory</key>
-			<string>com_google_SantaDriver</string>
 		</dict>
 	</dict>
+	<key>NSHumanReadableCopyright</key>
+	<string>Google, Inc.</string>
```
russellhancox: Google LLC

Source/common/SNTXPCControlInterface.m
tburgin: I was thinking we would update santad's launchd.plist at build time with the teamid. So all release builds would have a MachService of: `EQHXZ8M8AV.com.google.santa.daemon.xpc`. This way we can support running "legacy" santad or SystemExtension santad on 10.15.

We could keep santad's launchd.plist constant for "legacy" santad and dynamic for SystemExtension santad. If santad can detect if it is running as a SystemExtension that would be the best differentiator. I will look into it.

tburgin: What do you think about using the `XPC_SERVICE_NAME` env var?
`XPC_SERVICE_NAME=com.google.santa.daemon` when a started as a LaunchDaemon
`XPC_SERVICE_NAME=teamid.com.google.santa.daemon` when started as a SystemExtension.
I was hoping this would be the actual MachService name, instead it appears to the launchd label. Which should work, just to need to append `.xpc` to the names with a teamid at the beginning.

tburgin: Also SystemExtensions seem to be relocated before being executed. So that could also be a good indicator.

Output of:
```objc
    NSLog(@"%@", [NSProcessInfo processInfo].environment);
    NSLog(@"%@", [NSProcessInfo processInfo].arguments);
```

```
{
    PATH = "/usr/bin:/bin:/usr/sbin:/sbin";
    "XPC_FLAGS" = 0x0;
    "XPC_SERVICE_NAME" = "998Z9C32TC.com.google.santa.daemon";
}
(
    "/Library/SystemExtensions/99F756D2-DD7D-4C73-869C-4479B108017A/com.google.santa.daemon.systemextension/Contents/MacOS/santad"
)
```

tburgin: I couldn't think of a simpler way or coordinating the MachService name across components. So just back to checking macOS version.

tburgin:
PTAL

Source/santa/SNTAppDelegate.m
```diff
@@ -34,6 +37,15 @@ @implementation SNTAppDelegate
 #pragma mark App Delegate methods
 
 - (void)applicationDidFinishLaunching:(NSNotification *)aNotification {
+  if (@available(macOS 10.15, *)) {
+    LOGI(@"Requesting SystemExtension activation");
+    OSSystemExtensionRequest *req =
+        [OSSystemExtensionRequest activationRequestForExtension:@"com.google.santa.daemon"
```
russellhancox: I think this string should be a constant somewhere more central.

Source/santa/SNTAppDelegate.m
```diff
@@ -34,6 +37,15 @@ @implementation SNTAppDelegate
...
+                                                          queue:dispatch_get_main_queue()];
```
russellhancox: Does this block the thread?

tburgin: It did, yep. Much better on a non-main thread.

PR #388, opened 2019-09-09
add SNTEventProvider interface

tburgin:
* Initial EndpointSecurity integration

### DIFF
--- a/Source/santad/SNTEndpointSecurityManager.m
+++ b/Source/santad/SNTEndpointSecurityManager.m
@@ -40,8 +40,9 @@
       santa_message_t sm;
       sm.uid = audit_token_to_ruid(m->event.exec.target->audit_token);
       sm.gid = audit_token_to_rgid(m->event.exec.target->audit_token);
-      sm.pid =audit_token_to_pid(m->event.exec.target->audit_token);
-      sm.ppid = m->event.exec.target->ppid;
+      sm.pid = audit_token_to_pid(m->event.exec.target->audit_token);
+      // original_ppid stays constant even in the event a process is reparented
+      sm.ppid = m->event.exec.target->original_ppid;
       sm.es_message = (void *)es_copy_message(m);
 
       size_t l = m->event.exec.target->executable->path.length;
@@ -56,15 +57,15 @@
 
       switch (m->event_type) {
         case ES_EVENT_TYPE_AUTH_EXEC:
-          // TODO(bur/rah): Probably also want to do this for is_es_client.
-          // TODO(bur/rah): Since these events are not evaluated by Santa's pipline they are
-          //                missing bits of information such as SHA256 and REASON. Refactor the
-          //                logging cache.
-          if (m->event.exec.target->is_platform_binary) {
-            LOGD(@"platform binary: %s", sm.path);
-            [self postAction:ACTION_RESPOND_ALLOW forMessage:sm];
-            return;
-          }
+//          // TODO(bur/rah): Probably also want to do this for is_es_client.
+//          // TODO(bur/rah): Since these events are not evaluated by Santa's pipline they are
+//          //                missing bits of information such as SHA256 and REASON. Refactor the
+//          //                logging cache.
+//          if (m->event.exec.target->is_platform_binary) {
+//            LOGD(@"platform binary: %s", sm.path);
+//            [self postAction:ACTION_RESPOND_ALLOW forMessage:sm];
+//            return;
+//          }
           sm.action = ACTION_REQUEST_BINARY;
           if (self.decisionCallback) self.decisionCallback(sm);
           break;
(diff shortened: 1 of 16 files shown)


Source/santad/SNTEndpointSecurityManager.m
```diff
@@ -0,0 +1,159 @@
...
+  self = [super init];
+  if (self) {
+    es_client_t *client = NULL;
+    es_new_client_result_t ret = es_new_client(&client, ^(es_client_t *c, const es_message_t *m) {
+      // TODO(bur/rah): Currently this class only subscribes to exec events. Move this code
+      // somewhere exec specific if other types of events are added to this client.
+      santa_message_t sm;
+      sm.uid = audit_token_to_ruid(m->event.exec.target->audit_token);
+      sm.gid = audit_token_to_rgid(m->event.exec.target->audit_token);
+      sm.pid =audit_token_to_pid(m->event.exec.target->audit_token);
```
russellhancox: missing space after =

Source/santad/SNTEndpointSecurityManager.m
```diff
@@ -0,0 +1,159 @@
...
+      sm.ppid = m->event.exec.target->ppid;
```
russellhancox: Is there not a way to get this out of the audit_token?

Source/santad/SNTEndpointSecurityManager.m
```diff
@@ -0,0 +1,159 @@
...
+      strncpy(sm.path,m->event.exec.target->executable->path.data, l);
+      sm.path[l] = '\0';
+
+      switch (m->event_type) {
+        case ES_EVENT_TYPE_AUTH_EXEC:
+          // TODO(bur/rah): Probably also want to do this for is_es_client.
+          // TODO(bur/rah): Since these events are not evaluated by Santa's pipline they are
+          //                missing bits of information such as SHA256 and REASON. Refactor the
+          //                logging cache.
+          if (m->event.exec.target->is_platform_binary) {
```
russellhancox: I don't think we should ignore _all_ platform binaries - doesn't that include everything that's part of the base OS? We have some users who like to block things like iTunes, by hash, etc.

Source/santad/SNTEndpointSecurityManager.m
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  return self;
+}
+
+- (void)listenForDecisionRequests:(void (^)(santa_message_t))callback API_AVAILABLE(macos(10.15)) {
+  // Listen for exec auth messages.
+  self.decisionCallback = callback;
+  es_event_type_t events[] = { ES_EVENT_TYPE_AUTH_EXEC };
+  es_return_t sret = es_subscribe(self.client, events, 1);
```
russellhancox: What happens if this is called twice?

tburgin:
Unfortunately Travis is using an older Xcode 11 beta:
```
Source/santad/SNTEndpointSecurityManager.m:47:57: error: no member named 'length' in 'es_string_token_t'
      size_t l = m->event.exec.target->executable->path.length;
```

Version 11.0 beta 6 (11M392r)+:

```
typedef struct {
	// Length of the `data` field. Equivalent to strlen().
	size_t length;
	const char * data;
} es_string_token_t;
```

Source/santad/SNTEndpointSecurityManager.m
tburgin: It doesn't look to be easily exposed, no.
Though I did find "original_ppid" which "stays constant even in the event a process is reparented". I think this is exactly what we want, unless I am missing something.

Source/santad/SNTEndpointSecurityManager.m
tburgin: Yeah, I naively thought that "is_platform_binary" was just the "CoreOS" bits. That doesn't seem to be the case. /usr/bin/yes for example has "is_platform_binary" marked as true.

Commenting this out for now.

Source/santad/SNTEndpointSecurityManager.m
tburgin: Good question, I called it twice on the same client. Both succeeded. I didn't see duplicate auth exec messages. So it seems like an effective NOP.

tburgin:
PTAL

Source/santad/SNTEndpointSecurityManager.m
```diff
@@ -0,0 +1,159 @@
...
+///    distributed under the License is distributed on an "AS IS" BASIS,
+///    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+///    See the License for the specific language governing permissions and
+///    limitations under the License.
+
+#import "Source/santad/SNTEndpointSecurityManager.h"
+
+#import "Source/common/SNTLogging.h"
+
+#include <EndpointSecurity/EndpointSecurity.h>
```
russellhancox: import?

Source/santad/SNTEndpointSecurityManager.m
```diff
@@ -0,0 +1,159 @@
...
+
+@end
+
+@implementation SNTEndpointSecurityManager
+
+- (instancetype)init API_AVAILABLE(macos(10.15)) {
+  self = [super init];
+  if (self) {
+    es_client_t *client = NULL;
+    es_new_client_result_t ret = es_new_client(&client, ^(es_client_t *c, const es_message_t *m) {
```
russellhancox: I'm thinking that the content of this block should probably be moved into a method/function and the block just calls it.

tburgin:
PTAL

PR #401, opened 2019-10-20
handle all ACTION_RESPOND_*

tburgin:
* For large binaries, ACTION_RESPOND_ACK was double freeing the es message.
* Don't apply the prefix filter to exec events.

### DIFF
--- a/Source/santad/EventProviders/SNTEndpointSecurityManager.mm
+++ b/Source/santad/EventProviders/SNTEndpointSecurityManager.mm
@@ -199,7 +199,10 @@
       break;
   }
 
-  if (self.prefixTree->HasPrefix(sm.path)) {
+  // Filter events matching the prefix tree.
+  // Exec events are not filtered.
+  if (!(m->event_type == ES_EVENT_TYPE_AUTH_EXEC || m->event_type == ES_EVENT_TYPE_NOTIFY_EXEC) &&
+      self.prefixTree->HasPrefix(sm.path)) {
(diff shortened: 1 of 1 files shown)


Source/santad/EventProviders/SNTEndpointSecurityManager.mm
```diff
@@ -249,13 +252,18 @@ - (int)postAction:(santa_action_t)action forMessage:(santa_message_t)sm
...
       ret = es_respond_auth_result(self.client, (es_message_t *)sm.es_message,
                                    ES_AUTH_RESULT_ALLOW, true);
       break;
     case ACTION_RESPOND_DENY:
+    case ACTION_RESPOND_TOOLONG:
       ret = es_respond_auth_result(self.client, (es_message_t *)sm.es_message,
                                    ES_AUTH_RESULT_DENY, false);
       break;
+    case ACTION_RESPOND_ACK:
+      return ES_RESPOND_RESULT_SUCCESS;
```
russellhancox: Does this work? I thought you had to respond before returning?

tburgin: Yep, it works. We just need to ignore the ACKs:
https://github.com/google/santa/blob/master/Source/santad/SNTExecutionController.m#L115

The real response is still sent here:
https://github.com/google/santa/blob/master/Source/santad/SNTExecutionController.m#L160

PR #422, opened 2019-12-11
es event provider: support transitive whitelisting

tburgin:


Source/santad/EventProviders/SNTEndpointSecurityManager.mm
```diff
@@ -64,21 +72,77 @@ - (void)establishClient API_AVAILABLE(macos(10.15)) {
...
+            sm.action = ACTION_NOTIFY_WHITELIST;
+            sm.pid = pid;
+            LOGI(@"CLOSE: creating a transitive rule: path=%s pid=%d", sm.path, sm.pid);
+            self.decisionCallback(sm);
+          }
+
+          // Continue log this event
+          break;
+        }
+        case ES_EVENT_TYPE_NOTIFY_RENAME: {
```
russellhancox: I thought we determined catching renames wasn't necessary? Did I misremember or did you find otherwise?

Source/santad/EventProviders/SNTEndpointSecurityManager.mm
```diff
@@ -281,6 +328,17 @@ - (int)postAction:(santa_action_t)action forMessage:(santa_message_t)sm
   es_respond_result_t ret;
   switch (action) {
     case ACTION_RESPOND_ALLOW_COMPILER:
+      if (sm.pid >= PID_MAX) {
+        LOGE(@"Unable to watch compiler pid=%d >= pid_max=%d", sm.pid, PID_MAX);
+      } else {
+        LOGI(@"Watching compiler pid=%d path=%s", sm.pid, sm.path);
```
russellhancox: This might be quite noisy, perhaps LOGD?

Source/santad/EventProviders/SNTEndpointSecurityManager.mm
tburgin: We still need to catch renames for "compilers" that don't close the output before exiting. There is a workaround we could implement, but that should be in another pull.

PR #428, opened 2019-12-20
don't log TRUNCATE and don't log fileops from com.google.santa.daemon

tburgin:


Source/santad/EventProviders/SNTEndpointSecurityManager.mm
```diff
@@ -167,14 +170,18 @@ - (void)establishClient API_AVAILABLE(macos(10.15)) {
           break;
         }
         case ES_ACTION_TYPE_NOTIFY: {
+          // Don't log fileop events from com.google.santa.daemon
+          if (m->event_type != ES_EVENT_TYPE_NOTIFY_EXEC && self.selfPID == pid) return;
```
russellhancox: Swap these clauses as  the latter is presumably the less common

PR #433, opened 2020-02-07
santad: Move signature fetching into SNTPolicyProcessor

russellhancox:
This also removes an unnecessary hash and checks code signatures on non-MachO files (which is rare but possible).

It also fixes a rare crash in EndpointSecurityManager

### DIFF
--- a/Source/santad/SNTExecutionController.m
+++ b/Source/santad/SNTExecutionController.m
@@ -115,33 +115,22 @@ static size_t kLargeBinarySize = 30 * 1024 * 1024;
     [self.eventProvider postAction:ACTION_RESPOND_ACK forMessage:message];
   }
 
-  // If the binary is a critical system binary, don't check its signature.
-  // The binary was validated by santad at startup.
-  SNTCachedDecision *cd = self.ruleTable.criticalSystemBinaries[binInfo.SHA256];
-  MOLCodesignChecker *csInfo; // Needed further down in this scope.
-  if (!cd) {
-    // Get codesigning info about the file but only if it's a Mach-O.
-    if (binInfo.isMachO) {
-      NSError *csError;
-      csInfo = [[MOLCodesignChecker alloc] initWithBinaryPath:binInfo.path
-                                               fileDescriptor:binInfo.fileHandle.fileDescriptor
-                                                        error:&csError];
-      // Ignore codesigning if there are any errors with the signature.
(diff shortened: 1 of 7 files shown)


Source/santad/SNTExecutionController.m
```diff
@@ -115,66 +115,41 @@ - (void)validateBinaryWithMessage:(santa_message_t)message {
...
   // Send the decision to the kernel.
   [self.eventProvider postAction:action forMessage:message];
 
   // Log to database if necessary.
-  if (cd.decision != SNTEventStateAllowBinary &&
-      cd.decision != SNTEventStateAllowCompiler &&
-      cd.decision != SNTEventStateAllowTransitive &&
-      cd.decision != SNTEventStateAllowCertificate &&
-      cd.decision != SNTEventStateAllowScope) {
+  if (!(cd.decision & SNTEventStateAllow)) {
```
tburgin: We still want to add SNTEventStateAllowUnknown to the event database.

Source/santad/SNTPolicyProcessor.m
```diff
@@ -36,13 +38,27 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
...
+- (nonnull SNTCachedDecision *)decisionForFileInfo:(nonnull SNTFileInfo *)fileInfo
+                                        fileSHA256:(nullable NSString *)fileSHA256 {
   SNTCachedDecision *cd = [[SNTCachedDecision alloc] init];
   cd.sha256 = fileSHA256 ?: fileInfo.SHA256;
-  cd.certSHA256 = certificateSHA256;
+
+  // If the binary is a critical system binary, don't check its signature.
+  // The binary was validated at startup when the rule table was initialized.
+  SNTCachedDecision *systemCd = self.ruleTable.criticalSystemBinaries[cd.sha256];
+  if (systemCd) {
```
tburgin: nit: if (systemCd) return systemCd;

Source/santad/SNTExecutionController.m
russellhancox: Ahh, that's why we didn't use the mask

Source/santad/SNTPolicyProcessor.m
russellhancox: Done

PR #434, opened 2020-02-07
santad: Add config flag to block all binaries with bad signatures.

russellhancox:
Also log out a bad signature's error code.

### DIFF
--- a/Source/santad/SNTPolicyProcessor.m
+++ b/Source/santad/SNTPolicyProcessor.m
@@ -121,6 +121,14 @@
     }
   }
 
+  if (csInfoError && csInfoError.code != errSecCSUnsigned &&
+      [[SNTConfigurator configurator] enableBadSignatureProtection]) {
+    cd.decisionExtra =
+        [NSString stringWithFormat:@"Blocked due to signature error: %ld", (long)csInfoError.code];
+    cd.decision = SNTEventStateBlockCertificate;
+    return cd;
+  }
+
   NSString *msg = [self fileIsScopeBlacklisted:fileInfo];
   if (msg) {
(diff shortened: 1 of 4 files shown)


Source/santad/SNTPolicyProcessor.m
```diff
@@ -131,6 +135,14 @@ - (nonnull SNTCachedDecision *)decisionForFileInfo:(nonnull SNTFileInfo *)fileIn
     return cd;
   }
 
+  if (csInfoError && csInfoError.code != errSecCSUnsigned &&
```
tburgin: If we are considering this a "blacklisted scope", similar to the __PAGEZERO scope, shouldn't this go into the `fileIsScopeBlacklisted` method? Or at least before the `fileIsScopeWhitelisted` check?

russellhancox: Good point, I've moved it above both `fileIsScopeBlacklisted`/`fileIsScopeWhitelisted`.

Source/santad/SNTPolicyProcessor.m
```diff
@@ -117,6 +121,14 @@ - (nonnull SNTCachedDecision *)decisionForFileInfo:(nonnull SNTFileInfo *)fileIn
     }
   }
 
+  if (csInfoError && csInfoError.code != errSecCSUnsigned &&
+      [[SNTConfigurator configurator] enableBadSignatureProtection]) {
```
tburgin: nit: Since the default is NO, this saves a few steps for most folks.
if ([[SNTConfigurator configurator] enableBadSignatureProtection] &&
     csInfoError && csInfoError.code != errSecCSUnsigned)

PR #438, opened 2020-02-18
Config: EnableSystemExtension option

### LINKED ISSUE #437: Add UseSystemExtension Config
* Allow for Santa running on 10.15 to continue using the kext. This could prove useful for admins that want to control when Santa starts using the system extension. Especially during upgrades to 10.15, where system extension MDM policies may not yet be active.
* It should default to true

tburgin:
A (sort of) hack to support enterprise managed Macs updating to macOS 10.15+ from previous macOS versions with Santa installed.
 
Fixes #437

### DIFF
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -165,7 +165,7 @@
 @property(readonly, nonatomic) BOOL enableMachineIDDecoration;
 
 ///
-///  Use the bundled SystemExtension  on macOS 10.15+, defaults to YES.
+///  Use the bundled SystemExtension on macOS 10.15+, defaults to YES.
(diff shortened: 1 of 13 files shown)


Source/common/SNTConfigurator.h
```diff
@@ -164,6 +164,15 @@
 ///
 @property(readonly, nonatomic) BOOL enableMachineIDDecoration;
 
+///
+///  Use the bundled SystemExtension on macOS 10.15+, defaults to YES.
+///  Disable to continue using the bundled KEXT.
+///  This is a one way switch, if this is ever true on macOS 10.15+ the SystemExtension will be used.
```
russellhancox: s/the SystemExtension will be used/the kext will be deleted

PR #465, opened 2020-05-21
santad/sysx: Prevent unlinking databases

russellhancox:


Source/santad/EventProviders/SNTEndpointSecurityManager.mm
```diff
@@ -240,6 +240,32 @@ - (void)messageHandler:(es_message_t *)m API_AVAILABLE(macos(10.15)) {
...
     }
+    case ES_EVENT_TYPE_AUTH_UNLINK: {
+      es_string_token_t pathToken = m->event.unlink.target->path;
+      NSString *path = [[NSString alloc] initWithBytes:pathToken.data
+                                                length:pathToken.length
+                                              encoding:NSUTF8StringEncoding];
+      if ([path isEqualToString:@"/private/var/db/santa/rules.db"] ||
+          [path isEqualToString:@"/private/var/db/santa/events.db"]) {
+        LOGW(@"Preventing attempt to delete Santa databases!");
+        es_respond_auth_result(self.client, m, ES_AUTH_RESULT_DENY, false);
```
tburgin, line 251: I think we can cache this, we are never going to allow unlinking or renaming for these paths.

Source/santad/EventProviders/SNTEndpointSecurityManager.mm
```diff
@@ -240,6 +240,32 @@ - (void)messageHandler:(es_message_t *)m API_AVAILABLE(macos(10.15)) {
...
+      }
+      es_respond_auth_result(self.client, m, ES_AUTH_RESULT_ALLOW, false);
```
tburgin, line 253: Same for the allow.

Source/santad/EventProviders/SNTEndpointSecurityManager.mm
russellhancox: Good point

Source/santad/EventProviders/SNTEndpointSecurityManager.mm
russellhancox: Done - same for rename

russellhancox:
PTAL: also added a check to allow santad to unlink/rename if it needs to.

Source/santad/EventProviders/SNTEndpointSecurityManager.mm
```diff
@@ -240,6 +240,35 @@ - (void)messageHandler:(es_message_t *)m API_AVAILABLE(macos(10.15)) {
...
       break;
     }
+    case ES_EVENT_TYPE_AUTH_UNLINK: {
+      es_string_token_t pathToken = m->event.unlink.target->path;
+      NSString *path = [[NSString alloc] initWithBytes:pathToken.data
+                                                length:pathToken.length
+                                              encoding:NSUTF8StringEncoding];
+      if (([path isEqualToString:@"/private/var/db/santa/rules.db"] ||
+           [path isEqualToString:@"/private/var/db/santa/events.db"]) &&
+           audit_token_to_pid(m->process->audit_token) != getpid()) {
```
tburgin, line 250: nit: self.selfPID

russellhancox: Done

PR #466, opened 2020-05-21
santad: If database is locked don't attempt to unlink it

russellhancox:


Source/santad/DataLayer/SNTDatabaseTable.m
```diff
@@ -29,6 +31,10 @@ - (instancetype)initWithDatabaseQueue:(FMDatabaseQueue *)db {
   if (self) {
     [db inDatabase:^(FMDatabase *db) {
       if (![db goodConnection]) {
+        if ([db lastErrorCode] == SQLITE_LOCKED) {
+          LOGW(@"The database '%@' is locked by another process. Aborting.", [db databasePath]);
+          return nil;
```
tburgin, line 36: This won't propagate up to initWithDatabaseQueue: returning nil.

PR #477, opened 2020-07-08
syncservice: create stub for syncservice

tburgin:


Source/common/SNTXPCSyncServiceInterface.h
```diff
@@ -0,0 +1,51 @@
...
+#import "Source/common/SNTCommonEnums.h"
+
+@class SNTStoredEvent;
+
+///  Protocol implemented by syncservice and utilized by daemon and ctl for communication with a sync server.
+@protocol SNTSyncServiceXPC
+- (void)postEventsToSyncServer:(NSArray<SNTStoredEvent *> *)events isFromBundle:(BOOL)isFromBundle;
+- (void)postBundleEventToSyncServer:(SNTStoredEvent *)event
+                              reply:(void (^)(SNTBundleEventAction))reply;
+- (void)isFCMListening:(void (^)(BOOL))reply;
```
russellhancox, line 28: `- (void)performFullSyncWithProgress:(NSProgress *)progress`, maybe? or replace the NSProgress with a progress callback?

tburgin: NSProgress actually spans the XPC connection transparently. All we need is a reply block, similar to 
`hashBundleBinariesForEvent:reply`.

PR #493, opened 2020-08-30
Support for %hostname%, %uuid% and %serial% to eventDetailURLForEvent

hughneale:
Hello,

I've added support for `%hostname%`, `%uuid%` and `%serial%` to eventDetailURLForEvent to provide additional system information for blocked events. We're using the `%machine_id%` as our own unique identifier and it'd be useful to be able to provide some additional system information in event detail URLs for external systems that don't have access to the machine_id. 

Thanks

Best,

H

### DIFF
--- a/Source/common/SNTBlockMessage.m
+++ b/Source/common/SNTBlockMessage.m
@@ -126,15 +126,15 @@
     formatStr = [formatStr stringByReplacingOccurrencesOfString:@"%machine_id%"
                                                      withString:config.machineID];
   }
-  if (hostname) {
+  if (hostname.length) {
     formatStr = [formatStr stringByReplacingOccurrencesOfString:@"%hostname%"
                                                      withString:hostname];
   }
-  if (uuid) {
+  if (uuid.length) {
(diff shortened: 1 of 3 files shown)


hughneale:
@googlebot I signed it!

Source/common/SNTBlockMessage.m
```diff
@@ -122,6 +126,18 @@ + (NSURL *)eventDetailURLForEvent:(SNTStoredEvent *)event {
     formatStr = [formatStr stringByReplacingOccurrencesOfString:@"%machine_id%"
                                                      withString:config.machineID];
   }
+  if (hostname) {
```
russellhancox, line 129: Use `hostname.length` here - that will ensure the value is both non-nil and not zero-length. Same for uuid and serial below

hughneale: Thanks, I've amended this. 👍

hughneale:
Thank you!

russellhancox:
Thank you for the contribution! They'll be included in our next release, which will likely be in a few weeks as we're trying to get another long-awaited feature implemented.

PR #494, opened 2020-09-30
Make it possible to remotely set the FullSyncInterval

hughneale:
Hello,

The documentation seems to suggest the ability to remotely set the default sync interval from 10 minutes. Though maybe this was only explicitly meant for FCM notifications? Either way, I'd like to be able to remotely set the sync interval for non-FCM driven deployments. Happy to discuss if was done for a specific design rationale. However, I'd like to be able to set the sync interval remotely based on different client groups.

If you're happy I'll add a follow on commit with updates to the docs.

Many thanks,

H

### DIFF
--- a/Source/santactl/Commands/sync/SNTCommandSyncManager.m
+++ b/Source/santactl/Commands/sync/SNTCommandSyncManager.m
@@ -55,7 +55,7 @@ static NSString *const kFCMTargetHostIDKey = @"target_host_id";
 // allowlistNotificationQueue is used to serialize access to the allowlistNotifications dictionary.
 @property(nonatomic) NSOperationQueue *allowlistNotificationQueue;
 
-@property NSUInteger FullSyncInterval;
+@property NSUInteger fullSyncInterval;
 
 @property NSUInteger FCMFullSyncInterval;
 @property NSUInteger FCMGlobalRuleSyncDeadline;
@@ -124,7 +124,7 @@ static void reachabilityHandler(
     _allowlistNotificationQueue = [[NSOperationQueue alloc] init];
     _allowlistNotificationQueue.maxConcurrentOperationCount = 1;  // make this a serial queue
 
-    _FullSyncInterval = kDefaultFullSyncInterval;
+    _fullSyncInterval = kDefaultFullSyncInterval;
     _eventBatchSize = kDefaultEventBatchSize;
     _FCMFullSyncInterval = kDefaultFCMFullSyncInterval;
     _FCMGlobalRuleSyncDeadline = kDefaultFCMGlobalRuleSyncDeadline;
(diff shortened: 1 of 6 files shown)


Source/santactl/Commands/sync/SNTCommandSyncManager.m
```diff
@@ -55,6 +55,8 @@ @interface SNTCommandSyncManager () {
 // allowlistNotificationQueue is used to serialize access to the allowlistNotifications dictionary.
 @property(nonatomic) NSOperationQueue *allowlistNotificationQueue;
 
+@property NSUInteger FullSyncInterval;
```
russellhancox, line 58: Lowercase F (the FCM ones start upper because of being an acronym)

Source/santactl/Commands/sync/SNTCommandSyncPreflight.m
```diff
@@ -105,12 +105,17 @@ - (BOOL)sync {
   self.syncState.FCMToken = resp[kFCMToken];
 
   // Don't let these go too low
-  NSUInteger value = [resp[kFCMFullSyncInterval] unsignedIntegerValue];
+  NSUInteger FCMintervalValue = [resp[kFCMFullSyncInterval] unsignedIntegerValue];
   self.syncState.FCMFullSyncInterval =
-      (value < kDefaultFullSyncInterval) ? kDefaultFCMFullSyncInterval : value;
-  value = [resp[kFCMGlobalRuleSyncDeadline] unsignedIntegerValue];
+      (FCMintervalValue < kDefaultFullSyncInterval) ? kDefaultFCMFullSyncInterval : FCMintervalValue;
```
russellhancox, line 110: This should probably be (and always have been) `kDefaultFCMFullSyncInterval`.

Source/santactl/Commands/sync/SNTCommandSyncPreflight.m
```diff
@@ -105,12 +105,17 @@ - (BOOL)sync {
   self.syncState.FCMToken = resp[kFCMToken];
 
   // Don't let these go too low
-  NSUInteger value = [resp[kFCMFullSyncInterval] unsignedIntegerValue];
+  NSUInteger FCMintervalValue = [resp[kFCMFullSyncInterval] unsignedIntegerValue];
```
russellhancox, line 108: `FCMIntegerValue`

Though as this variable is being re-used for multiple distinct purposes, value is still fine.

Source/santactl/Commands/sync/SNTCommandSyncState.h
```diff
@@ -35,6 +35,9 @@
 /// An XSRF token to send in the headers with each request.
 @property NSString *xsrfToken;
 
+/// Full sync interval in seconds without FCM to update kDefaultFullSyncInterval
```
russellhancox, line 38: `without FCM to update kDefaultFullSyncInterval` => `when FCM is not used, defaults to 10m.`

russellhancox (changes requested):
Just a few style nits, generally LGTM.

If you don't mind also updating the docs, that would be very much appreciated! 👍

Source/santactl/Commands/sync/SNTCommandSyncPreflight.m
hughneale: I was going to leave it as value but I thought you might want them split - happy to do whatever you'd prefer 😄

hughneale:
I've added a single line of documentation. I feel there should be more which is why I put it off for the follow up PR. Is there anywhere else I should be looking? Thank you for the quick PR reviews.

russellhancox (approved):

russellhancox:
Thanks!

PR #498, opened 2020-10-16
Add the option to ignore actions from other ES clients

tburgin:
* Also bump version to 1.15

### DIFF
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -321,9 +321,9 @@
 @property(readonly, nonatomic) BOOL enableForkAndExitLogging;
 
 ///
-///  If true, skip processing any actions from other endpoint security clients. Defaults to false.
+///  If true, ignore actions from other endpoint security clients. Defaults to false. This only applies when running as a sysx.
 ///
-@property(readonly, nonatomic) BOOL skipOtherEndpointSecurityClients;
+@property(readonly, nonatomic) BOOL ignoreOtherEndpointSecurityClients;
 
 ///
 ///  Retrieve an initialized singleton configurator object using the default file path.
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -321,6 +321,11 @@
 @property(readonly, nonatomic) BOOL enableForkAndExitLogging;
 
 ///
+///  If true, skip processing any actions from other endpoint security clients. Defaults to false.
+///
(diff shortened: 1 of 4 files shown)


Source/common/SNTConfigurator.h
```diff
@@ -320,6 +320,11 @@
 ///
 @property(readonly, nonatomic) BOOL enableForkAndExitLogging;
 
+///
+///  If true, skip processing any actions from other endpoint security clients. Defaults to false.
```
russellhancox, line 324: Call out that this only applies when running as a sysx; _other_ implies it but let's make it clear.

Source/common/SNTConfigurator.m
```diff
@@ -79,6 +79,7 @@ @implementation SNTConfigurator
 static NSString *const kEnableSystemExtension = @"EnableSystemExtension";
 
 static NSString *const kEnableForkAndExitLogging = @"EnableForkAndExitLogging";
+static NSString *const kSkipOtherEndpointSecurityClients = @"SkipOtherEndpointSecurityClients";
```
russellhancox, line 82: _Skip_ sounds like we might be preventing them from doing processing. Maybe _ignore_?

Source/santad/EventProviders/SNTEndpointSecurityManager.mm
```diff
@@ -77,6 +77,14 @@ - (void)establishClient API_AVAILABLE(macos(10.15)) {
 
     es_client_t *client = NULL;
     es_new_client_result_t ret = es_new_client(&client, ^(es_client_t *c, const es_message_t *m) {
+      // If enabled, skip any action generated from another endpoint security client.
+      if (config.skipOtherEndpointSecurityClients && m->process->is_es_client) {
```
russellhancox, line 81: Reverse these conditions - I think `is_es_client` is going to be very rare and is a fairly simple field access vs a property get (with all the KVO goodness that comes with it).

russellhancox (changes requested):

Source/common/SNTConfigurator.h
tburgin: Done

Source/common/SNTConfigurator.m
tburgin: Good idea, done

Source/santad/EventProviders/SNTEndpointSecurityManager.mm

russellhancox (approved):

PR #501, opened 2020-10-21
project: add EnableDebugLogging option

tburgin:


Source/common/SNTConfigurator.m
```diff
@@ -581,6 +587,11 @@ - (BOOL)ignoreOtherEndpointSecurityClients {
   return number ? [number boolValue] : NO;
 }
 
+- (BOOL)enableDebugLogging {
+  NSNumber *number = self.configState[kEnableDebugLogging];
+  return number ? [number boolValue] : self.debugFlag;
```
russellhancox, line 592: Thinking about this some more, we probably want to take any true value so that EnableDebugLogging=false in the config doesn't prevent the --debug flag from working.

```objc
return [number boolValue] || self.debugFlag
```

tburgin: Good idea, done.

russellhancox (approved):

PR #507, opened 2020-10-30
Project: Add GitHub Actions CI workflow

russellhancox:
This workflow will do a release build and run all tests on every pull request on to the main branch and every push to other branches. If a push _only_ affects docs the workflow won't run


tburgin (approved):
Rock on

PR #520, opened 2020-12-18
release: split out the kext into a separate release label

tburgin:
* Allows for different `--macos_cpus` flags when building the kext or app
* Also fix `SNTLoggingKernel` BUILD rule
* Bump version to 1.16

### DIFF
--- a/Source/santa_driver/BUILD
+++ b/Source/santa_driver/BUILD
@@ -54,6 +54,7 @@ macos_kernel_extension(
     bundle_id = "com.google.santa-driver",
     bundle_name = "santa-driver",
     infoplists = ["Info.plist"],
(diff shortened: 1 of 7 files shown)


Source/santa_driver/BUILD
```diff
@@ -70,6 +70,11 @@ objc_library(
     deps = ["//Source/common:SNTKernelCommon"],
 )
 
+sh_binary(
+    name = "last_resort",
+    srcs = ["last_resort.sh"],
```
russellhancox, line 75: This doesn't exist?

tburgin: Oops, removed.

russellhancox (approved):

PR #524, opened 2021-01-08
santad: Add an optional internal cache for use with the sysx

### LINKED ISSUE #510: High Build times on Big Sur
Santa 1.15 on Big Sur is causing high local build times.

What logs would be useful to debug this?

```
> santactl status
>>> Daemon Info
  Driver Connected          | Yes
  Mode                      | Monitor
  File Logging              | Yes
  Watchdog CPU Events       | 72  (Peak: 665.36%)
  Watchdog RAM Events       | 0  (Peak: 216.54MB)
```

Screen Shot 2020-12-02 at 6 40 23 AM

russellhancox:
This bypasses the sysx provided cache for executions and uses SantaCache to implement our own version. Performance seems better, and should resolve the issues we've been seeing when 2 EndpointSecurity extensions are running on a machine.

Fixes #510


Source/santad/EventProviders/SNTEndpointSecurityManager.h
```diff
@@ -17,5 +17,14 @@
...
 
+#include <EndpointSecurity/EndpointSecurity.h>
+
+// Gleaned from https://opensource.apple.com/source/xnu/xnu-4903.241.1/bsd/sys/proc_internal.h
+const pid_t PID_MAX = 99999;
+
 @interface SNTEndpointSecurityManager : NSObject<SNTEventProvider>
+- (santa_vnode_id_t)vnodeIDForFile:(es_file_t *)file;
+
+@property(nonatomic) es_client_t *client;
```
tburgin, line 28: Can this be readonly here, then make it readwrite in SNTEndpointSecurityManager.mm?

Source/santad/EventProviders/SNTEndpointSecurityManager.mm
```diff
@@ -107,6 +103,9 @@ - (void)establishClient API_AVAILABLE(macos(10.15)) {
           // Ignore unmodified files
           if (!m->event.close.modified) return;
 
+          // Remove from decision cache in case this is invalidating a cached binary.
+          [self removeCacheEntryForVnodeID:[self vnodeIDForFile:m->event.close.target]];
```
tburgin, line 107: How is this callable from SNTEndpointSecurityManager.mm?

Source/santad/EventProviders/SNTCachingEndpointSecurityManager.mm
```diff
@@ -0,0 +1,192 @@
...
+      // Allow the exec, but don't cache the decision so subsequent execs of the compiler get
+      // marked appropriately.
+      ret = es_respond_auth_result(self.client, (es_message_t *)sm.es_message,
+                                   ES_AUTH_RESULT_ALLOW, false);
+      break;
+    case ACTION_RESPOND_ALLOW:
+    case ACTION_RESPOND_ALLOW_PENDING_TRANSITIVE:
+      [self addToCache:sm.vnode_id decision:ACTION_RESPOND_ALLOW timeout:GetCurrentUptime()];
+      ret = es_respond_auth_result(self.client, (es_message_t *)sm.es_message,
+                                   ES_AUTH_RESULT_ALLOW, false);
```
tburgin, line 97: I think we do want to cache here with ES. It means less work for us in general. If/when the ES cache gets cleared without our knowledge our cache will kick in.

Source/santad/EventProviders/SNTCachingEndpointSecurityManager.mm
```diff
@@ -0,0 +1,192 @@
...
+  while (true) {
+    // Check to see if item is in cache
+    auto return_action = [self checkCache:vnode_id];
+
+    // If item was in cache with a valid response, return it.
+    // If item is in cache but hasn't received a response yet, sleep for a bit.
+    // If item is not in cache, break out of loop and forward request to callback.
+    if (RESPONSE_VALID(return_action)) {
+      if (return_action == ACTION_RESPOND_ALLOW) {
+        es_respond_auth_result(self.client, m, ES_AUTH_RESULT_ALLOW, false);
```
tburgin, line 59: same here

Source/santad/EventProviders/SNTEndpointSecurityManager.h
russellhancox: Good call

Source/santad/EventProviders/SNTEndpointSecurityManager.mm
russellhancox: It's implemented further down (and is a no-op) and then overridden in the caching subclass

Source/santad/EventProviders/SNTCachingEndpointSecurityManager.mm
russellhancox: I was a little worried about how the 2 caches would interact, particularly as we don't really understand the behavior of the ES cache. But I think you're right, the less work we do the better, if we see issues we can revisit.

Source/santad/EventProviders/SNTCachingEndpointSecurityManager.mm
russellhancox: Done

Source/santad/EventProviders/SNTEndpointSecurityManager.mm
tburgin: Oops, I see - it's not new.

tburgin (approved):
Nice work

PR #526, opened 2021-01-26
santactl/sync: retry individual requests during a sync

russellhancox:
Each request is retried up to 3 times with gaps of 2s, 4s, 6s

### DIFF
--- a/Source/santactl/Commands/sync/SNTCommandSyncStage.m
+++ b/Source/santactl/Commands/sync/SNTCommandSyncStage.m
@@ -85,7 +85,12 @@
   NSError *error;
   NSData *data;
 
-  for (int attempt = 1; attempt < 4; ++attempt) {
+  for (int attempt = 1; attempt < 6; ++attempt) {
+    if (attempt > 1) {
+      struct timespec ts = {.tv_sec = (attempt * 2)};
+      nanosleep(&ts, NULL);
+    }
+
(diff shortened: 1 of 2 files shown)


Source/santactl/Commands/sync/SNTCommandSyncStage.m
```diff
@@ -83,20 +83,31 @@ - (NSMutableURLRequest *)requestWithDictionary:(NSDictionary *)dictionary {
...
+         error.code == NSURLErrorCannotParseResponse) &&
+        [self fetchXSRFToken]) {
+      NSMutableURLRequest *mutableRequest = [request mutableCopy];
+      [mutableRequest setValue:self.syncState.xsrfToken forHTTPHeaderField:kXSRFToken];
+      request = mutableRequest;
+      continue;
+    }
+
+    struct timespec ts = {.tv_sec = (attempt * 2)};
+    nanosleep(&ts, NULL);
```
tburgin, line 107: nit: on the last loop we will be sleeping needlessly

if (attempt + 1 < 4) nanosleep(&ts, NULL);

tburgin (approved):

Source/santactl/Commands/sync/SNTCommandSyncStage.m
russellhancox: Done, PTAL

tburgin (approved):

PR #540, opened 2021-02-25
santad: Fix transitive rules when using the sysx cache feature

### LINKED ISSUE #539: Santa 2021.02 SysxCache and transitive whitelisting: unable to run binaries marked as compilers.
Hi team,

We found a bug with Santa 2021.02 in Big Sur with the SysCache on. Any binary marked as a compiler can only be run once and then it will be killed.

The behavior is different in Big Sur vs Catalina using system extensions.

For Big Sur (11.2 and 11.3 Beta):

1) Install Santa 2012.02 with these config options enabled
```
    EnableSysxCache = 1;
    IgnoreOtherEndpointSecurityClients = 1;     # we're unsure yet if this has any effect or not
```

2) Mark a binary as a compiler (we mark Xcode's `ld` as a compiler).

3) Run it 2 times. The first time it will run, the second one will hang for about 60 seconds and then be killed.

```
$ /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld
ld: warning: platform not specified
ld: warning: -arch not specified
ld: warning: No platform min-version specified on command line
ld: no object files specified
$ /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld
Killed: 9
```

4) Check Console logs:

```
default	13:48:56.761076-0800	com.google.santa.daemon	E com.google.santa.daemon: Deadline reached: deny pid=28938 ret=0
```

---

In macOS 10.15 with system extensions and SysCache on the behavior is a bit different:

* When running the compiler it will run the first time immediately. The second time it will take about 60 seconds and finally run.
* Somehow, mac reports that santa is killed by the compiler.

```system.log
Feb 24 14:51:27 C02C71ESMD6V com.apple.xpc.launchd[1] (EQHXZ8M8AV.com.google.santa.daemon[327]): Service exited due to SIGKILL | sent by ld[52920]
```

We also saw in the Console `sysextd` complaining about `request contains no authorization` right before com.google.santa.daemon reported it was starting up again.

russellhancox:
There are a few related changes - the increase of the deadline timeout is because 2s seems just a fraction too short sometimes and adjusting the DEBUG log mapping was just to help debugging of this feature.

Fixes #539

### DIFF
--- a/Source/common/SNTLogging.m
+++ b/Source/common/SNTLogging.m
@@ -88,7 +88,10 @@ void logMessage(LogLevel level, FILE *destination, NSString *format, ...) {
         break;
       case LOG_LEVEL_DEBUG:
         levelName = "D";
-        syslogLevel = ASL_LEVEL_DEBUG;
+        // Log debug messages at the same ASL level as INFO.
+        // While it would make sense to use DEBUG, watching debug-level logs
+        // in Console means enabling all debug logs, which is absurdly noisy.
+        syslogLevel = ASL_LEVEL_NOTICE;
         break;
     }
 
--- a/Source/santad/EventProviders/SNTCachingEndpointSecurityManager.mm
+++ b/Source/santad/EventProviders/SNTCachingEndpointSecurityManager.mm
@@ -79,7 +79,7 @@ template<> uint64_t SantaCacheHasher<santa_vnode_id_t>(santa_vnode_id_t const& t
     }
   }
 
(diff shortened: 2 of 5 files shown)


tburgin (approved):
Nice! Small nit: `timeout:` from `[addToCache:decision:timeout:]` is misleading, mind renaming it to currentTicks or similar?

Issue #553, opened 2021-04-30
Santa not recognizing the client certificate in the system keychain


ameyah:
Hello Santa devs,

I have been seeing this issue with some Santa clients. The problem doesn't seem to be with the certificate - other apps (like Chrome) can access the same client certificate.

Sync log:

```
Preflight starting
Clean sync requested by user
Performing request, attempt 1
Server Trust: <server-trust>
Server asked for client authentication but no usable client certificate found.
Failed to decode JSON response: Error Domain=NSCocoaErrorDomain Code=3840 "JSON text did not start with array or object and option to allow fragments not set." UserInfo={NSDebugDescription=JSON text did not start with array or object and option to allow fragments not set.}
Preflight complete
Event upload starting
Performing request, attempt 1
Failed to decode JSON response: Error Domain=NSCocoaErrorDomain Code=3840 "JSON text did not start with array or object and option to allow fragments not set." UserInfo={NSDebugDescription=JSON text did not start with array or object and option to allow fragments not set.}
Uploaded 3 events
Event upload complete
Rule download starting
Performing request, attempt 1
Failed to decode JSON response: Error Domain=NSCocoaErrorDomain Code=3840 "JSON text did not start with array or object and option to allow fragments not set." UserInfo={NSDebugDescription=JSON text did not start with array or object and option to allow fragments not set.}
Rule download failed, aborting run
```

The cert CN matches with the `ClientAuthCertificateCN` value. `find-certificate` does locate that certificate

```
$ /usr/bin/security find-certificate -a -c <cn> -m -Z':
SHA-256 hash: B0319D238A72933EC5C3C2EE430F696F5DAD32E05CBC1C1575DE619ABCAF5097
<redacted>
    "labl"<blob>="<cn>"
    ...
```

Any other logs that might help the debugging?

Thanks!

ameyah:
FWIW adding a new certificate (and deleting the previous one) with the same CN sometimes works, but is unreliable.

russellhancox:
Does `sudo security find-identity -v` list the certificate?

ameyah:
Found the issue:

The previous certificates (expired at this point) with the same CN were seemingly deleted with `security delete-certificate`. The private key isn't deleted as per the documentation, but it seems the certificate isn't deleted as well, just hidden from the keychain UI. `security find-certificate` does show up the expired certificate.

My theory is Santa is picking up the older certificate, but doesn't use it because its expired.

[...]

PR #554, opened 2021-05-03
Project: Bump dependency versions, fix reload command

russellhancox:
The reload command would previously fail if you used multiple compilation modes for building as it would try extracting the versions from both comp modes.

The dependency bump includes a fix for #553.

Bump Santa version to 2021.5

### DIFF
--- a/version.bzl
+++ b/version.bzl
@@ -1,3 +1,3 @@
 """The version for all Santa components."""
 
-SANTA_VERSION = "2021.3"
+SANTA_VERSION = "2021.5"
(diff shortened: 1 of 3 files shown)


version.bzl
```diff
@@ -1,3 +1,3 @@
 """The version for all Santa components."""
 
-SANTA_VERSION = "2021.3"
+SANTA_VERSION = "2021.4"
```
tburgin, line 3: 2021.5

tburgin (approved):

version.bzl
russellhancox: Oops. Done

PR #569, opened 2021-08-12
Prevent Overwrite of Santa Databases

tnek:
SNTEndpointSecurityManager checks the source of an es_event_rename_t for tampering with the santa databases (events.db and rules.db) but it doesn't check destinations. A user can arbitrarily overwrite either Santa database by renaming a file into the databases to remove/modify rules and events.

### DIFF
--- a/Source/santad/EventProviders/SNTEndpointSecurityManagerTest.mm
+++ b/Source/santad/EventProviders/SNTEndpointSecurityManagerTest.mm
@@ -107,7 +107,7 @@ const NSString *const kRulesDBPath = @"/private/var/db/santa/rules.db";
     SNTEndpointSecurityManager *snt = [[SNTEndpointSecurityManager alloc] init];
 
     XCTestExpectation *expectation = [self expectationWithDescription:@"Wait for response from ES"];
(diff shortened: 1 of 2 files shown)


Source/santad/EventProviders/SNTEndpointSecurityManagerTest.mm
```diff
@@ -200,4 +200,128 @@ - (void)testSkipOtherESEvents {
...
 }
 
+- (void)testRenameOverwriteRulesDB {
+  for (const NSString *testPath in @[ kEventsDBPath, kRulesDBPath ]) {
+    MockEndpointSecurity *mockES = [MockEndpointSecurity mockEndpointSecurity];
+    [mockES reset];
+    SNTEndpointSecurityManager *snt = [[SNTEndpointSecurityManager alloc] init];
+
+    XCTestExpectation *expectation = [self expectationWithDescription:@"Wait for response from ES"];
+    __block ESResponse *got = nil;
```
tburgin, line 210: No need for the nil initialization, objc variables are nil initialized by default.

__block ESResponse *got;

tburgin (approved):

PR #570, opened 2021-08-12
Add an extra return on the rename case

tnek:
Missed handling the fallthrough from https://github.com/google/santa/pull/569.

This should've been caught by a linter, will add that to our CI in a followup PR


pmarkowsky (approved):
LGTM.

PR #579, opened 2021-08-17
Add actual test binaries to the repo

tnek:
I apologize in advance - forgot to actually add the binaries in https://github.com/google/santa/pull/578

This should've been caught faster by the CI - will have to look into how we can speed that up.


pmarkowsky (approved):
LGTM.

PR #582, opened 2021-08-17
Project: Update deps in WORKSPACE to be reproducible.

russellhancox:
This removes a lot of warnings when starting in a clean workspace


pmarkowsky (approved):
LGTM.

PR #584, opened 2021-08-18
Project: remove xcode projects, they're stale

russellhancox:


tnek (dismissed):
Should they be included in the .gitignore as well?

russellhancox:
Good call

tnek (approved):

Issue #585, opened 2021-08-18
Documentation Overhaul


russellhancox:
The docs are, in many ways, out-of-date and could be much clearer. We need to document how to configure many things, better docs on deployment, better docs on how to build the project, etc.

We could also do with tidying up the old ReadTheDocs config.

russellhancox:
cc @pmarkowsky

pmarkowsky:
Just listing things off the top of my head that we should add to docs:

- [ ] How to setup your own environment (bootstrapping with your own dev certs).
- [ ] State machines for rules (e.g. what takes precedence certificate vs. binary rules.)
- [ ] Setting up an easy dev cycle with configs.
- [ ] Updating coverage, etc.

pmarkowsky:
- [ ] Using ESF vs. the KEXT.

pmarkowsky:
We should also document the sync server protocol.

russellhancox:
- [ ] Better documentation of configuration options instead of a simple table with no explanation. Ideally we'd have some automation of these docs from the headers.

PR #586, opened 2021-08-18
Project: delete obsolete ReadTheDocs config

russellhancox:
We haven't used ReadTheDocs for >6m

Part of #585


tnek (approved):

russellhancox:
> FYI: https://github.com/google/santa#docs still points to https://santa.readthedocs.io/, but that's now broken.

Oops, mailed a CL to fix.

PR #588, opened 2021-08-19
Pin bazel rules_apple to https://github.com/bazelbuild/rules_apple/pull/1191

tnek:
Turns out 3.13.3 doesn't actually include this commit. 

Tested locally on an M1 mac this time.


keith (approved):
note there is a bit of discussion about how it should work if you want to test an intel binary on a m1 machine, but this is probably still an improvement for the default case https://github.com/bazelbuild/rules_apple/issues/1221

pmarkowsky (approved):
LGTM.

PR #596, opened 2021-09-08
Add tulsi project

tnek:
Create a tulsi project that correctly labels and marks our main artifacts (all of our binaries and unit tests) and update the docs


pmarkowsky (approved):
LGTM.  This is the first step towards autogenerating the XCode project from the given targets.

PR #602, opened 2021-09-20
Include an integration test setup and fixture with moroz

tnek:
This PR includes several scripts for building and setting up an integration test environment that doesn't require the Santa provisionprofile. This sets up an environment such that a build of santad with `--define=SANTA_BUILD_TYPE=ci` can actually load. 

I also tested whether the entitlements file change for santad affects our internal provisionfile - the entitlements in Santa_Daemon_Dev.provisionprofile will take priority over the `entitlements` build rule.

### DIFF
--- a/Testing/init_dev_certs.sh
+++ b/Testing/init_dev_certs.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+set -e
+GIT_ROOT=$(git rev-parse --show-toplevel)
+CNF_PATH=$GIT_ROOT/Testing/openssl.cnf
+
+openssl genrsa -out ./santa.key 2048
+openssl rsa -in ./santa.key -out ./santa.key
+openssl req -new -key ./santa.key -out ./santa.csr -config $CNF_PATH
+openssl x509 -req -days 10 -in ./santa.csr -signkey ./santa.key -out ./santa.crt -extfile $CNF_PATH -extensions codesign
+openssl pkcs12 -export -out santa.p12 -inkey santa.key -in santa.crt -password pass:santa
+
+KEYCHAIN="/Library/Keychains/System.keychain"
+sudo security import ./santa.p12 -k $KEYCHAIN -A -P santa
+sudo security add-trusted-cert -d -r trustRoot -k $KEYCHAIN santa.crt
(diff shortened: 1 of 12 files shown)


Testing/init_dev_certs.sh
```diff
@@ -0,0 +1,14 @@
...
+
+openssl genrsa -out ./santa.key 2048
+openssl rsa -in ./santa.key -out ./santa.key
+openssl req -new -key ./santa.key -out ./santa.csr -config $CNF_PATH
+openssl x509 -req -days 10 -in ./santa.csr -signkey ./santa.key -out ./santa.crt -extfile $CNF_PATH -extensions codesign
+openssl pkcs12 -export -out santa.p12 -inkey santa.key -in santa.crt -password pass:santa
+
+KEYCHAIN="/Library/Keychains/System.keychain"
+sudo security import ./santa.p12 -k $KEYCHAIN -A -P santa
+sudo security add-trusted-cert -d -r trustRoot -k $KEYCHAIN santa.crt
```
tburgin, line 14: Does add-trusted-cert work on Big Sur+?

tburgin (approved):

Testing/init_dev_certs.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+set -e
```
groob, line 2: Another option is to use the go tools. Example:

go run $(go env GOROOT)/src/crypto/tls/generate_cert.go --host=localhost:9000,127.0.0.1:9000 --ecdsa-curve=P256 --ca=true

pmarkowsky (approved):
LGTM.

Testing/init_dev_certs.sh
tnek: Yes, while working on this, I've tested this on 11.5.2 and 11.6 for Big Sur, and also on Beta 6 of Monterrey.

Testing/init_dev_certs.sh
tnek: For it to be a valid codesigning cert, you also specifically need the extended properties included in `openssl.cnf`.

Issue #563, opened 2021-08-09
Provide support for metrics


pmarkowsky:
For larger deployments we'd like Santa to be able to export metrics for use with monitoring solutions like prometheus,  or osquery.

arubdesu:
Can you be more specific what observability/metrics use cases you're thinking about? I can't think of any workstation-side agents I've ever encountered that have health/metrics endpoints for prometheus in specific, osquery has Docker/lxd-specific stuff, but besides CarbonBlack info the tables seem more applicable to servers. osquery can look at santa's local logs... prometheus is used sync server-side for graphing the rabbitMQ-alike event processing pipeline by products like zentral. 
Most of the things I care about managing a fleet of not-insignificant size is "what are my unknowns over x hours/days" which really only seem to be useful in aggregate. If the sync server is missing data you need to properly manage the service... wouldn't someone at Google point their focus at (internal, actually use-able) Upvote? Respectfully speaking.

pmarkowsky:
For background, at Google we're using Monarch to monitor our agents specifically CPU/memory usage, look for crash loops and other reliability metrics.  Having the Santa integrated helps us find issues like crashes / CPU spikes after rule syncs and other oddities. Additionally this lets us track statistics over time such as how much of the fleet is in lockdown mode and stays in lockdown mode or how often are folks updating rules.  Internally we're working to make it easy for Santa to work with our monitoring solution like the rest of our agents. 

That said I suggested prometheus and other solutions out there to try and make this a feature that isn't just for Google and others can use it if they'd like.

PR #604, opened 2021-09-20
Add initial configuration options for barebones metric service

pmarkowsky:
This PR adds the initial configuration options for the bare bones.

Specifically it adds the following three things:

| Option | Meaning |
|---|---|
| **MetricFormat** |  a string describing how metrics should be formatted. Supported options ("raw json" for a single JSON object containing all metrics, "json" for JSON metrics one metric per line.  |
|  **MetricURL** | a string URL describing  where to export the formatted metrics. |
| **exportMetrics** | a helper function in SNTConfigurator that checks if all necessary keys are set and metrics should be exported. | 

Part of #563.

### DIFF
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -382,7 +382,7 @@
 @property(readonly, nonatomic) SNTMetricFormatType metricFormat;
 
 ///
-/// URL describing where metrics are exported, defaults to "".
+/// URL describing where metrics are exported, defaults to nil.
 ///
 @property(readonly, nonatomic) NSURL *metricURL;
 
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -377,9 +377,9 @@
 @property(readonly, nonatomic) BOOL exportMetrics;
 
 ///
-/// String format for to export metrics as, defaults to "".
+/// Format to export Metrics as.
 ///
-@property(readonly, nonatomic) NSString *metricFormat;
+@property(readonly, nonatomic) SNTMetricFormatType metricFormat;
 
 ///
 /// URL describing where metrics are exported, defaults to "".
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -372,6 +372,11 @@
 @property(readonly, nonatomic) BOOL fcmEnabled;
 
 ///
+/// True if metricsExportFormat and metricsFile path are all set. Defaults to false.
+///
+@property(readonly, nonatomic) BOOL exportMetrics;
+
+///
 ///  Retrieve an initialized singleton configurator object using the default file path.
 ///
 + (instancetype)configurator;
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -372,9 +372,20 @@
 @property(readonly, nonatomic) BOOL fcmEnabled;
 
 ///
-/// True if metricsExportFormat and metricsFile path are all set. Defaults to false.
+/// True if metricsFormat and metricsURL are set. False otherwise.
 ///
 @property(readonly, nonatomic) BOOL exportMetrics;
+
+///
+/// String format for to export metrics as, defaults to "".
+///
+@property(readonly, nonatomic) NSString *metricFormat;
+
+///
+/// URL describing where metrics are exported, defaults to "".
+///
+@property(readonly, nonatomic) NSURL *metricURL;
+
 
 ///
 ///  Retrieve an initialized singleton configurator object using the default file path.
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -102,7 +102,7 @@ static NSString *const kBlockedPathRegexKey = @"BlockedPathRegex";
 static NSString *const kBlockedPathRegexKeyDeprecated = @"BlacklistRegex";
 
 // TODO(markowsky): move these to sync server only.
-static NSString *const kMetricsFormat = @"ExportMetricsFormat";
+static NSString *const kMetricsFormat = @"MetricsFormat";
 static NSString *const kMetricsFile = @"ExportMetricsFile";
 
 // The keys managed by a sync server.
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -102,6 +102,8 @@ static NSString *const kBlockedPathRegexKey = @"BlockedPathRegex";
 static NSString *const kBlockedPathRegexKeyDeprecated = @"BlacklistRegex";
 
 // TODO(markowsky): move these to sync server only.
+// Valid values for kMetricsFormat should be rawjson, and json. 
+// TODO(markowsky): add support for prometheus.
 static NSString *const kMetricsFormat = @"MetricsFormat";
 static NSString *const kMetricsURL = @"MetricsURL";
 
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -102,10 +102,8 @@ static NSString *const kBlockedPathRegexKey = @"BlockedPathRegex";
 static NSString *const kBlockedPathRegexKeyDeprecated = @"BlacklistRegex";
 
 // TODO(markowsky): move these to sync server only.
-// Valid values for kMetricsFormat should be rawjson, and json. 
-// TODO(markowsky): add support for prometheus.
-static NSString *const kMetricsFormat = @"MetricsFormat";
-static NSString *const kMetricsURL = @"MetricsURL";
+static NSString *const kMetricFormat = @"MetricFormat";
+static NSString *const kMetricURL = @"MetricURL";
 
 // The keys managed by a sync server.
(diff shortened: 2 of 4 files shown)


Source/common/SNTConfigurator.m
```diff
@@ -101,6 +101,10 @@ @implementation SNTConfigurator
 static NSString *const kBlockedPathRegexKey = @"BlockedPathRegex";
 static NSString *const kBlockedPathRegexKeyDeprecated = @"BlacklistRegex";
 
+// TODO(markowsky): move these to sync server only.
+static NSString *const kMetricsFormat = @"ExportMetricsFormat";
```
tnek, line 105: Add a comment listing the valid format types too

tnek (commented):
Add the configuration keys and key info from the PR description to the [configuration doc](https://github.com/google/santa/blob/main/docs/deployment/configuration.md) as well

Source/common/SNTConfigurator.m
```diff
@@ -101,6 +101,10 @@ @implementation SNTConfigurator
...
+static NSString *const kMetricsFile = @"ExportMetricsFile";
+
```
russellhancox, line 107: What do you think of making this a URL where file:// URLs are accepted? I could see other formats being added that export via HTTPS

Source/common/SNTConfigurator.m
tnek: Sorry, meant for valid values. I'm assuming it's something like "json" vs "proto"?

pmarkowsky: I was going to put list the values in a future PR. 

Sketched out I have the following in my notes:

- **rawJson**: a full blob of the metrics (basically the raw metrics
- **json**: metrics rewritten to one metric per line for our system.
- **prometheus**:  serialized to the prometheus format.

I'll update this.

Source/common/SNTConfigurator.m
pmarkowsky: That makes a lot of sense. My initial thought was to have one option per output type since  we have the issue of pull vs. push for different metric systems, e.g. prometheus expects that like borgmon of old we're serving HTTP.  

I'll update the metric service code to treat it as a URL instead.

Source/common/SNTConfigurator.m
pmarkowsky, line 105: MetricsFormat is correct.

```suggestion
static NSString *const kMetricsFormat = @"MetricsFormat";
```

Source/common/SNTConfigurator.m
```diff
@@ -101,6 +101,10 @@ @implementation SNTConfigurator
 static NSString *const kBlockedPathRegexKey = @"BlockedPathRegex";
 static NSString *const kBlockedPathRegexKeyDeprecated = @"BlacklistRegex";
 
+// TODO(markowsky): move these to sync server only.
+static NSString *const kMetricsFormat = @"ExportMetricsFormat";
+static NSString *const kMetricsFile = @"ExportMetricsFile";
```
pmarkowsky, line 106:
```suggestion
static NSString *const kMetricsURL = @"MetricsURL";
```

Source/common/SNTConfigurator.m
```diff
@@ -654,6 +660,22 @@ - (BOOL)fcmEnabled {
   return (self.fcmProject.length && self.fcmEntity.length && self.fcmAPIKey.length);
 }
 
+
+///
+/// Returns YES if all of the necessary options are set to export metrics, NO
+/// otherwise.
+///
+- (BOOL) exportMetrics {
+  NSArray<NSString *> *metricKeys = @[kMetricsFormat, kMetricsFile];
```
pmarkowsky, line 669:
```suggestion
  NSArray<NSString *> *metricKeys = @[kMetricsFormat, kMetricsURL];
```

Source/common/SNTConfigurator.m
pmarkowsky: @russellhancox I fixed the issue. Apparently I didn't push the last commit before I'd written the PR description.

Source/common/SNTConfigurator.m
```diff
@@ -101,6 +101,10 @@ @implementation SNTConfigurator
 static NSString *const kBlockedPathRegexKey = @"BlockedPathRegex";
 static NSString *const kBlockedPathRegexKeyDeprecated = @"BlacklistRegex";
 
+// TODO(markowsky): move these to sync server only.
```
pmarkowsky, line 104:
```suggestion
// TODO(markowsky): move these to sync server only.
// Valid values for kMetricsFormat should be rawjson, and json. 
// TODO(markowsky): add support for prometheus.
```

Source/common/SNTConfigurator.m
```diff
@@ -171,6 +177,8 @@ - (instancetype)init {
       kFCMProject : string,
       kFCMEntity : string,
       kFCMAPIKey : string,
+      kMetricsFormat : string,
+      kMetricsFile : string,
```
pmarkowsky, line 181:
```suggestion
      kMetricsURL : string,
```

tnek (approved):

Source/common/SNTConfigurator.h
```diff
@@ -371,6 +371,22 @@
...
 
+///
+/// True if metricsFormat and metricsURL are set. False otherwise.
+///
+@property(readonly, nonatomic) BOOL exportMetrics;
+
+///
+/// String format for to export metrics as, defaults to "".
+///
+@property(readonly, nonatomic) NSString *metricFormat;
```
russellhancox, line 382: It might be better for this to be an enum with the getter handling the conversion from a string. That way downstream code can switch on it.

Source/common/SNTConfigurator.h
```diff
@@ -371,6 +371,22 @@
...
+
+///
+/// URL describing where metrics are exported, defaults to "".
```
russellhancox, line 385: > defaults to ""

why not nil?

Source/common/SNTConfigurator.m
```diff
@@ -1,4 +1,4 @@
-/// Copyright 2015 Google Inc. All rights reserved.
+/// Copyright 2021 Google Inc. All rights reserved.
```
russellhancox, line 1: FYI: You don't need to update the copyright years - they should be correct when added and then left alone.

Source/common/SNTConfigurator.h
pmarkowsky: I switched it to an enum. Originally I didn't do this because I was thinking that its less descriptive in the config. 

However since we're already doing this for ClientMode I agree it makes more sense as an enum.

Source/common/SNTConfigurator.m
pmarkowsky: Good to know.  It's an old habit from other jobs. I'll stop going forward.

Source/common/SNTConfigurator.h
```diff
@@ -371,6 +371,22 @@
...
+///
+@property(readonly, nonatomic) BOOL exportMetrics;
+
+///
+/// Format to export Metrics as.
+///
+@property(readonly, nonatomic) SNTMetricFormatType metricFormat;
+
+///
+/// URL describing where metrics are exported, defaults to "".
```
pmarkowsky, line 385:
```suggestion
/// URL describing where metrics are exported, defaults to nil.
```

Source/common/SNTConfigurator.h
pmarkowsky: It's a typo. Fixing.

Source/common/SNTConfigurator.m
```diff
@@ -654,6 +662,32 @@ - (BOOL)fcmEnabled {
...
+/// Returns YES if all of the necessary options are set to export metrics, NO
+/// otherwise.
+///
+- (BOOL) exportMetrics {
+  return self.configState[kMetricsFormat] != SNTMetricFormatTypeUnknown &&
+    ![self.configState[kMetricsURL] isEqualToString:@""];
+}
+
+- (SNTMetricFormatType)metricsFormat {
+  switch ([self.configState[kMetricsFormat] longLongValue]) {
```
russellhancox, line 676: I'd keep the value that users set as a string (you're right, it's much clearer) and have this bit of code translating from that string to the enum value.

pmarkowsky: I've changed it back to a string and normalized the values entered by the operator.

Source/common/SNTConfigurator.m
```diff
@@ -654,6 +660,34 @@ - (BOOL)fcmEnabled {
...
+- (SNTMetricFormatType)metricFormat {
+  NSString *normalized = [self.configState[kMetricFormat] lowercaseString];
+  normalized = [normalized stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
+
+  if ([normalized isEqualToString: @"rawjson"]) {
+    return SNTMetricFormatTypeRawJSON;
+  } else if ([normalized isEqualToString: @"json"]) {
+    return SNTMetricFormatTypeJSON;
+  } else {
+      return SNTMetricFormatTypeUnknown;
```
russellhancox, line 682: Indent -2

PR #614, opened 2021-09-24
Deflake SNTApplicationTest by tracking subscriptions to specific event types

tnek:
EndpointSecurityTestUtil assumed only a single es_subscribe was needed to be called before serving all events. This resulted in listenForLogRequests and listenForDecisionRequests racing each other, which would result in auth events being dropped during tests if listenForLogRequests won the race.

This PR changes mock `es_subscribe` to note specific events we're ready for and switches us to `waitForExpectations` so that we can explicitly declare which expectations to wait for.


tnek:
This doesn't seem to have fully fixed it. Adding a sleep(1) afterwards is an ugly hack that deflakes it for now. 

Will go back to this afterwards, but this should at least stop it from blocking our releases.

russellhancox (approved):

PR #615, opened 2021-09-27
Remove use of entitlements field for the santad build rule

tnek:
Revert entitlements change and update the codesign behavior of  `Testing/build_and_sign.sh` to match that of the internal kokoro `sign.sh`.

I misunderstood how our release workflow worked when testing for if https://github.com/google/santa/pull/602 broke anything - as it turns out, our .provisionfile isn't actually used in the release build and we instead use separate keys _and_ the entitlements file, where the actual `codesign` tool fills in TeamIdentifierPrefix. The entitlements arg for the build rule can't nicely fill this in, so remove it for now and instead mirror the kokoro sign.sh behavior within external build_and_sign.sh.

### DIFF
--- a/Testing/build_and_sign.sh
+++ b/Testing/build_and_sign.sh
@@ -2,6 +2,7 @@
 set -e
 GIT_ROOT=$(git rev-parse --show-toplevel)
 
+KEYCHAIN="santa-dev-test.keychain"
 SANTAD_PATH=Santa.app/Contents/Library/SystemExtensions/com.google.santa.daemon.systemextension/Contents/MacOS/com.google.santa.daemon
 SANTAD_ENTITLEMENTS="$GIT_ROOT/Source/santad/com.google.santa.daemon.systemextension.entitlements"
 SANTA_BIN_PATH=Santa.app/Contents/MacOS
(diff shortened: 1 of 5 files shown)


pmarkowsky (approved):
LGTM, but seems like a good time to wait for @russellhancox or @tburgin to chime in, in case we've missed something with the greater build pipeline.

Testing/build_and_sign.sh
```diff
@@ -3,20 +3,25 @@ set -e
...
 
     echo "> Build complete, installing santa"
     TMP_DIR=$(mktemp -d)
     tar xvf $GIT_ROOT/bazel-bin/santa-*.tar.gz -C $TMP_DIR
+    CS_ARGS="--prefix=EQHXZ8M8AV -fs $SIGNING_IDENTITY --timestamp --options library,kill,runtime"
 
-    for bin in $TMP_DIR/binaries/$SANTA_BIN_PATH/* $TMP_DIR/binaries/$SANTAD_PATH; do
-        sudo codesign --prefix=EQHXZ8M8AV --preserve-metadata=entitlements -fs $SIGNING_IDENTITY --timestamp --options library,kill,runtime $bin
+    for bin in $TMP_DIR/binaries/$SANTA_BIN_PATH/*; do
+        sudo codesign --preserve-metadata=entitlements ${CS_ARGS} $bin
```
russellhancox, line 20: Can we drop `sudo` from all the codesign and bazel commands? I don't think it's needed.

tnek: Done

russellhancox (approved):

PR #616, opened 2021-09-30
Detect existence of a provisionprofile and use that instead in build_and_sign.sh

tnek:
We only create a self-signed cert to sign with if there's no provisionprofile. 

Also add the option to randomize the version string in build_and_sign.sh.


pmarkowsky (approved):
LGTM.

PR #622, opened 2021-10-04
README: sync servers; remove Upvote, sort the rest alphabetically.

russellhancox:
The Upvote project has been archived so we should stop pointing people at it.


pmarkowsky (approved):
LGTM.

PR #623, opened 2021-10-04
Tests: ensure SNTPrefixTree test finishes executing at appropriate time

russellhancox:
Also switch to using XCTAssert* instead of XCTFail

### DIFF
--- a/Source/common/SNTPrefixTreeTest.mm
+++ b/Source/common/SNTPrefixTreeTest.mm
@@ -50,7 +50,9 @@
   // Create a bunch of background noise.
   dispatch_async(dispatch_get_global_queue(0, 0), ^{
     for (uint64_t i = 0; i < UINT64_MAX; ++i) {
-      t->HasPrefix([UUIDs[i % count] UTF8String]);
+      dispatch_async(dispatch_get_global_queue(0, 0), ^{
+        t->HasPrefix([UUIDs[i % count] UTF8String]);
+      });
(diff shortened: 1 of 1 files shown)


Source/common/SNTPrefixTreeTest.mm
```diff
@@ -45,26 +45,27 @@ - (void)testThreading {
     [UUIDs addObject:[NSUUID UUID].UUIDString];
   }
 
+  __block BOOL stop = NO;
+
   // Create a bunch of background noise.
   dispatch_async(dispatch_get_global_queue(0, 0), ^{
-    dispatch_apply(UINT64_MAX, dispatch_get_global_queue(0, 0), ^(size_t i) {
```
tburgin, line 50: Replacing dispatch_apply with a loop removes the concurrent noise generation and replaces it with lots of serial calls to HasPrefix. I think the intention was to create noise on lots of threads, but it may not be necessary?

russellhancox: Good point, we can't use dispatch_apply because there's no way to stop it (we could have the block called by dispatch_apply check the bail-out variable but we're still calling that block millions of times needlessly). We can use dispatch_async ourselves to push the HasPrefix call onto a queue but still bail the loop early once we're done.

tburgin (approved):

tnek (approved):

PR #668, opened 2021-11-08
Change SNTEventLog to be a singleton emit a singleton Logger object

tnek:
Presently, SNTEventLog is already managed as effectively a singleton by the nature of all of the logging mechanisms themselves within SNTApplication anyways. 

Making our logging layer a (stub-able) singleton simplifies this part of adding other EventProviders, and should make writing "did this attempt to log?" tests easier.


tnek:
Doesn't work quite as expected - will re-send PR later

PR #670, opened 2021-11-08
Change SNTEventLog to be a singleton emit a singleton Logger object

tnek:
Presently, SNTEventLog is already managed as effectively a singleton by the nature of all of the logging mechanisms themselves within SNTApplication anyways.

Making our logging layer a (stub-able) singleton simplifies this part of adding other EventProviders, and should make writing "did this attempt to log?" tests easier.

(prev: https://github.com/google/santa/pull/668, realized the issues were unrelated but cleaned up this PR a bit more)


Source/santad/Logs/SNTEventLog.m
```diff
@@ -418,4 +420,23 @@ - (NSString *)originalPathForTranslocation:(santa_message_t)message {
   return [origURL path];  // this will be nil if there was an error
 }
 
++ (instancetype)logger {
+  static SNTEventLog *logger;
+  static dispatch_once_t onceToken;
+  SNTConfigurator *configurator = [SNTConfigurator configurator];
```
mlw, line 426: Recommend removing the `configurator` variable. It's only used once inside the ``dispatch_once` and doesn't need to be grabbed on every call.

tnek: Done

Source/santad/Logs/SNTEventLog.m
```diff
@@ -418,4 +420,23 @@ - (NSString *)originalPathForTranslocation:(santa_message_t)message {
   return [origURL path];  // this will be nil if there was an error
 }
 
++ (instancetype)logger {
+  static SNTEventLog *logger;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    SNTConfigurator *configurator = [SNTConfigurator configurator];
+    switch ([configurator eventLogType]) {
```
russellhancox, line 428:
```suggestion
    switch ([[SNTConfigurator configurator] eventLogType]) {
```

russellhancox (changes requested):

russellhancox (approved):

PR #673, opened 2021-11-10
Tests: Update SNTMetricHTTPWriterTest

russellhancox:


Source/santametricservice/Writers/SNTMetricHTTPWriterTest.m
```diff
@@ -28,15 +28,15 @@ - (void)setUp {
   self.httpWriter = [[SNTMetricHTTPWriter alloc] init];
   self.mockResponses = [[NSMutableArray alloc] init];
```
tnek, line 30: Can you add a comment here with context for why this is necessary?

russellhancox: This is the best I could come up with, does it make sense?

tnek (approved):

PR #683, opened 2021-11-19
Add a configuration option for users to add their own root labels

pmarkowsky:
This PR adds the ability for users to add their own root labels for metrics. 

This can be useful if you want to add identifiers for observability or if your existing monitoring solution needs explicit values to be set.

Related-issue: #563


Source/common/SNTConfigurator.m
```diff
@@ -181,6 +183,7 @@ - (instancetype)init {
       kMetricFormat : string,
       kMetricURL : string,
       kMetricExportInterval : number,
+      kMetricExtraLabels : string,
```
russellhancox, line 186: You could declare this as a dictionary and avoid any parsing. You'll need to add an option for that on line 125 but then the getter on line 719 can just return the value directly.

Source/santad/SNTApplicationCoreMetrics.m
```diff
@@ -85,9 +85,24 @@ static void RegisterMemoryAndCPUMetrics(SNTMetricSet *metricSet) {
...
+
+  // get extra root labels from configuration
+  SNTConfigurator *config = [SNTConfigurator configurator];
+
+  NSDictionary *extraLabels = [config extraMetricLabels];
+
+  if ([extraLabels count] >= 0) {
+    NSLog(@"Extra labels %@", extraLabels);
+    for (NSString *key in extraLabels) {
+      [metricSet addRootLabel:key value:(NSString *)extraLabels[key]];
```
russellhancox, line 103: Will this do-the-right-thing if the config contains a label already declared, like host_name? What happens if the "value" is empty? Can that be used to un-set a default label?

russellhancox (changes requested):

Source/santad/SNTApplicationCoreMetrics.m
```diff
@@ -85,9 +85,34 @@ static void RegisterMemoryAndCPUMetrics(SNTMetricSet *metricSet) {
...
+      if ([@"" isEqualToString:(NSString *)extraLabels[key]]) {
+        [metricSet removeRootLabel:key];
+        continue;
+      }
+
+      // Set or override the value.
+      [metricSet addRootLabel:key value:(NSString *)extraLabels[key]];
+    }
+  }
 }
```
russellhancox, line 117:
```suggestion
  if (extraLabels.count == 0) return;
    
  for (NSString *key in extraLabels) {
    // remove the root label if the value is empty.
    if ([@"" isEqualToString:(NSString *)extraLabels[key]]) {
      [metricSet removeRootLabel:key];
      continue;
    }

    // Set or override the value.
    [metricSet addRootLabel:key value:(NSString *)extraLabels[key]];
  }
} 
```

russellhancox (dismissed):

russellhancox (approved):

PR #685, opened 2021-11-23
USB mass storage blocking and remounting

tnek:
This is a fairly large PR adding extra USB authorization functionality to Santa's existing USB mounting logging. For Googlers, this shouldn't come as a surprise; see the go/santa-usb-blocking design doc which will also have the below details:

The following two config keys have been added:
* `BlockUSBMount` - toggling authorization of USB mounting on/off. 
* `RemountUSBMode` - configuration for forced USB remounting to other permissions.

These are the following combinations of options and their respective behaviors:

1. if `BlockUSBMount` is enabled & `RemountUSBMode` is unset:All removable USB storage mounting will be rejected.
2. if `BlockUSBMount` is enabled & `RemountUSBMode` is set: The initial mount request will be rejected if it doesn't comply with the flags set in RemountUSBMode, and a followup mount request will arrive with the mount flags set in RemountUSBMode.
3. If `BlockUSBMount` is disabled: Santa's event log will continue to log but not require any authorization of disks that appear and/or are mounted, as it has always done.

For the actual mount flags supported, see the options documented for the `-o` flag in `$ man mount`

PR Implementation details:
* USB Mount listening & diskarbitration code previously in Santa has been decoupled from SNTApplication and moved into its own standalone SNTDeviceManager.
* Add the actual remounting and blocking mechanisms This is accomplished by a mix of 1) ES' ES_EVENT_AUTH_MOUNT to reject mounts with unauthorized flags and 2) diskarbitration's DADiskMountWithArguments
* Configuration docs updated with the new config keys.
* Add the sync service and config key for enabling mass USB storage blocking
* Update EndpointSecurityTestUtil and existing tests that use it to properly handle multiple ES clients

For followup PRs:
* Dedicated docs for this feature - most likely will be a cleaned up version (removing Google-specific details) of the design doc in go/santa-usb-blocking
* Functionality to notify the user when this is occurring & configuration options for the actual message sent to users. 
* Create a DiskArbitrationTestUtil to fully mock out a fake disk. This will be necessary later if we'd like to do detailed structured logging of all DADIskRef DiskInfo data.


Source/santad/EventProviders/SNTDeviceManager.mm
```diff
@@ -0,0 +1,292 @@
...
+  // gracefully handle it with a deny-by-default in the worst-case before it can do that.
+  // This isn't an issue for notify events, so we're in no rush for those.
+  std::shared_ptr<std::atomic<bool>> responded;
+  if (m->action_type == ES_ACTION_TYPE_AUTH) {
+    responded = std::make_shared<std::atomic<bool>>(false);
+    dispatch_after(timeout, self.esAuthQueue, ^(void) {
+      if (responded->load()) return;
+      LOGE(@"SNTDeviceManager: deadline reached: deny pid=%d ret=%d",
+           audit_token_to_pid(m->process->audit_token),
+           es_respond_auth_result(c, m, ES_AUTH_RESULT_DENY, false));
```
mlw, line 256: We have similar logic in `SNTEndpointSecurityManager.mm`. As a future TODO, it would be nice to consider having centralized logic for handling timeouts of messages from arbitrary clients

Source/santad/EventProviders/SNTDeviceManager.mm
```diff
@@ -0,0 +1,292 @@
...
+- (void)initES API_AVAILABLE(macos(10.15)) {
+  while (!self.client) {
+    es_client_t *client = NULL;
+    es_new_client_result_t ret = es_new_client(&client, ^(es_client_t *c, const es_message_t *m) {
+      // Set timeout to 5 seconds before the ES deadline.
+      [self handleESMessageWithTimeout:m
+                            withClient:c
+                               timeout:dispatch_time(m->deadline, NSEC_PER_SEC * -5)];
+    });
+    [self flushCache];
```
mlw, line 141: Can remove this call. ES automatically clears caches when new clients connect (or existing clients disconnect)

Source/santad/EventProviders/SNTDeviceManager.h
```diff
@@ -0,0 +1,45 @@
...
+@interface SNTDeviceManager : NSObject
+
+@property BOOL subscribed;
+@property BOOL blockUSBMount;
+@property NSArray<NSString *> *remountArgs;
+
+- (instancetype)init;
+
+- (void)listen;
+- (void)listenDA;
```
mlw, line 28: Nit: would recommend not exposing `listenDA`/`listenES` externally

Source/santad/SNTApplication.m
```diff
@@ -415,6 +395,27 @@ - (void)observeValueForKeyPath:(NSString *)keyPath
...
     [self stopMetricsPoll];
     [self startMetricsPoll];
+  } else if ([keyPath isEqualToString:NSStringFromSelector(@selector(blockUSBMount))]) {
+    BOOL new = [ change[newKey] boolValue ];
+    BOOL old = [change[oldKey] boolValue];
+
+    if (new != old) {
+      LOGI(@"BlockUSBMount changed: %d -> %d", old, new);
+      self.deviceManager.blockUSBMount = new;
+      if (@available(macos 10.15, *)) [self.deviceManager flushCache];
```
mlw, line 405: Might be worth considering to never cache `AUTH MOUNT` responses, and then we never have to flush all caches on rule updates. Mount operations are already slow, and likely infrequent. Having the ES client handle every one is unlikely to add enough overhead to be noticed. However, flushing the ES cache can definitely have an impact on the system and should be avoided.

Would recommend making `flushCache` an internal-only interface on `SNTDeviceManager.mm`. (maybe even remove the interface altogether since, as currently designed, we subscribe to `AUTH MOUNT` immediately after creating the client which already flushed the cache anyways.

Source/santad/EventProviders/SNTDeviceManager.h
```diff
@@ -0,0 +1,45 @@
...
+- (void)flushCache API_AVAILABLE(macos(10.15));
+- (void)remount:(DADiskRef)disk mountMode:(long)remountMask;
+
+// handleESMessage handles an ES message synchronously. This will block all incoming ES events
+// until either we serve a response or we hit the auth deadline. Prefer [SNTDeviceManager
+// handleESMessageWithTimeout]
+- (void)handleESMessage:(const es_message_t *)m
+             withClient:(es_client_t *)c API_AVAILABLE(macos(10.15));
+
+- (void)handleESMessageWithTimeout:(const es_message_t *)m
```
mlw, line 40: It doesn't look like the `handleESMessage*` methods are called outside of this class. Recommend not exposing externally.

Source/santad/EventProviders/SNTDeviceManager.mm
tnek: Added it as to-do for now. This PR in this form actually originated from when I was trying to make the SNTEndpointSecurityManager's messageHandler actually handle all of the messages

Source/santad/EventProviders/SNTDeviceManager.mm
tnek: Done.

Source/santad/EventProviders/SNTDeviceManager.h
tnek: Done

Source/santad/SNTApplication.m
tnek: Done, removed `flushCache` entirely.

Source/santad/EventProviders/SNTDeviceManager.h
tnek: Done

Source/santad/EventProviders/SNTDeviceManagerTest.mm
```diff
@@ -0,0 +1,95 @@
...
+  if (!deviceManager.subscribed) {
+    XCTestExpectation *initExpectation =
+      [self expectationWithDescription:@"Wait for SNTDeviceManager to instantiate"];
+
+    dispatch_async(dispatch_get_global_queue(QOS_CLASS_USER_INTERACTIVE, 0), ^{
+      [deviceManager listen];
+    });
+
+    dispatch_async(dispatch_get_global_queue(QOS_CLASS_USER_INTERACTIVE, 0), ^{
+      while (!deviceManager.subscribed)
```
mlw, line 46: I don't think `[deviceManager listen]` is asynchronous at all, so this `dispatch_async` and loop aren't really necessary. (Could maybe just assert `subscribed` is set?)

Source/santad/EventProviders/SNTDeviceManager.h
```diff
@@ -0,0 +1,45 @@
...
+@property NSArray<NSString *> *remountArgs;
+
+- (instancetype)init;
+
+- (void)listen;
+- (void)listenDA;
+- (void)listenES API_AVAILABLE(macos(10.15));
+
+- (void)flushCache API_AVAILABLE(macos(10.15));
+- (void)remount:(DADiskRef)disk mountMode:(long)remountMask;
```
mlw, line 28: Nit: This interface also doesn't appear to be needed externally.

Source/santad/EventProviders/SNTDeviceManagerTest.mm
tnek: This is to enforce a hard timeout incase some future code change causes our client to never initialize correctly. I can add a comment explaining this.

Source/santad/EventProviders/SNTDeviceManager.h
tnek: Removed.

mlw (approved):

Source/santad/EventProviders/SNTDeviceManager.mm
```diff
@@ -0,0 +1,288 @@
...
+      return;
+    }
+
+    long newMode = mountMode | remountOpts;
+    LOGI(@"SNTDeviceManager: remounting device '%s'->'%s', flags (%lu) -> (%lu)",
+         m->event.mount.statfs->f_mntfromname, m->event.mount.statfs->f_mntonname, mountMode,
+         newMode);
+    [self remount:disk mountMode:newMode];
+  }
+  es_respond_auth_result(self.client, m, ES_AUTH_RESULT_DENY, false);
```
pmarkowsky, line 226: Move this before the if on line 213.

Not sure we need to worry about the timeout if we do this.

pmarkowsky (commented):
Overall looks good. 

We should normalize and sanitize the data from the config (e.g. lowercase everything and explicitly filter semi-colons etc.).

Source/common/SNTConfigurator.m
```diff
@@ -145,6 +149,8 @@ - (instancetype)init {
       kAllowedPathRegexKeyDeprecated : re,
       kBlockedPathRegexKey : re,
       kBlockedPathRegexKeyDeprecated : re,
+      kBlockUSBMountKey : number,
+      kRemountUSBModeKey : array,
```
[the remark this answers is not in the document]
tnek: Args to mount from remountArgs are already sanitized by `mountArgsToMask`.

Source/santad/EventProviders/SNTDeviceManager.mm
tnek: Done

Source/santad/EventProviders/SNTDeviceManager.h
```diff
@@ -0,0 +1,29 @@
...
+#import <DiskArbitration/DiskArbitration.h>
+#import <Foundation/Foundation.h>
+
+#include <EndpointSecurity/EndpointSecurity.h>
+
+@interface SNTDeviceManager : NSObject
+
+@property BOOL subscribed;
+@property BOOL blockUSBMount;
+@property NSArray<NSString *> *remountArgs;
```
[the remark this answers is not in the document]
tnek: Done

Source/santad/EventProviders/SNTDeviceManager.h
```diff
@@ -0,0 +1,29 @@
...
+///    distributed under the License is distributed on an "AS IS" BASIS,
+///    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+///    See the License for the specific language governing permissions and
+///    limitations under the License.
+#import <DiskArbitration/DiskArbitration.h>
+#import <Foundation/Foundation.h>
+
+#include <EndpointSecurity/EndpointSecurity.h>
+
+@interface SNTDeviceManager : NSObject
```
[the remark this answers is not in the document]
tnek: Done

Source/common/SNTConfigurator.m
```diff
@@ -399,6 +405,14 @@ + (NSSet *)keyPathsForValuesAffectingEnableBadSignatureProtection {
   return [self configStateSet];
 }
 
++ (NSSet *)keyPathsForValuesAffectingBlockUSBMount {
```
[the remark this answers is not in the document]
tnek: If I understand it correctly, keyPathsForValuesAffecting:string methods used by KVO observers, which we register in SNTApplication for config keys.

Source/common/SNTConfigurator.m
```diff
@@ -399,6 +405,14 @@ + (NSSet *)keyPathsForValuesAffectingEnableBadSignatureProtection {
...
+  return [self configStateSet];
+}
+
++ (NSSet *)keyPathsForValuesAffectingRemountUSBMode {
```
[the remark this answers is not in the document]
tnek: Same as above comment

pmarkowsky (dismissed):
LGTM.

pmarkowsky (approved):
Approving again. Now that tests are passing.

PR #687, opened 2021-11-30
Add a metrics command to santactl

### LINKED ISSUE #657: Add a metrics command to santactl
We should add a metrics command to santactl. The command should get a copy of the SNTMetricSet via XPC and then format it for users as both text and JSON. 

# Steps to Completion

- [ ] Add a new XPC interface
- [ ] Add a new command to santactl called metrics
- [ ] Support json formatting via a flag

pmarkowsky:
Adds a metrics command to santactl that displays the current status of metrics in santad.

Fixes #657 

Related to: #563. 

Example Output:

```
>>> Metrics Info
  Metrics Server            | http://localhost:2444/submit
  Metrics Format            | monarchjson
  Export Interval (seconds) | 30

>>> Root Labels
  host_name                 | markowsky-macbookpro.roam.corp.google.com
  username                  | root
  job_name                  | santad
  service_name              | santa

>>> Metrics 
  Metric Name               | /santa/mode
  Description               | Santa's operating mode
  Type                      | SNTMetricTypeGaugeString 6
  Created                   | 2021-11-30T19:03:16.928Z
  Last Updated              | 2021-11-30T20:24:52.817Z
  Data                      | monitor

  Metric Name               | /proc/memory/resident_size
  Description               | The resident set size of this process
  Type                      | SNTMetricTypeGaugeInt64 7
  Created                   | 2021-11-30T19:03:16.928Z
  Last Updated              | 2021-11-30T20:24:52.817Z
  Data                      | 50081792

  Metric Name               | /proc/os/version
  Description               | Short operating System version
  Type                      | SNTMetricTypeConstantString 2
  Created                   | 2021-11-30T19:03:16.928Z
  Last Updated              | 2021-11-30T19:03:16.928Z
  Data                      | 12.1

  Metric Name               | /build/label
  Description               | Version of the binary
  Type                      | SNTMetricTypeConstantString 2
  Created                   | 2021-11-30T19:03:16.928Z
  Last Updated              | 2021-11-30T19:03:16.928Z
  Data                      | 2021.21

  Metric Name               | /proc/birth_timestamp
  Description               | Start time of Santad, in microseconds since epoch
  Type                      | SNTMetricTypeConstantInt64 3
  Created                   | 2021-11-30T19:03:16.928Z
  Last Updated              | 2021-11-30T19:03:16.928Z
  Data                      | 1638298996927623

  Metric Name               | /proc/memory/virtual_size
  Description               | The virtual memory size of this process
  Type                      | SNTMetricTypeGaugeInt64 7
  Created                   | 2021-11-30T19:03:16.928Z
  Last Updated              | 2021-11-30T20:24:52.817Z
  Data                      | 36701704192

  Metric Name               | /proc/cpu_usage
  Description               | CPU time consumed by this process, in seconds
  Type                      | SNTMetricTypeGaugeDouble 8
  Field                     | mode=user
  Created                   | 2021-11-30T19:03:16.928Z
  Last Updated              | 2021-11-30T20:24:52.817Z
  Data                      | 74.86385799999999
  Field                     | mode=system
  Created                   | 2021-11-30T19:03:16.928Z
  Last Updated              | 2021-11-30T20:24:52.817Z
  Data                      | 53.404325

```


Source/santactl/Commands/SNTCommandMetrics.m
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    printf("\n");
+    printf(">>> Metrics \n");
+
+    for (NSString *metricName in normalizedMetrics[@"metrics"]) {
+      NSDictionary *metric = normalizedMetrics[@"metrics"][metricName];
+      const char *metricNameStr = [metricName UTF8String];
+      const char *description = [metric[@"description"] UTF8String];
+      // NSString * metricType =  makeSNTMetricTypeString(metric[@"type"]);
```
tnek, line 149: Should this be uncommented?

Source/santactl/Commands/SNTCommandMetrics.m
```diff
@@ -0,0 +1,180 @@
...
+    NSString *metricStr = [[NSString alloc] initWithData:metricData encoding:NSUTF8StringEncoding];
+    printf("%s\n", [metricStr UTF8String]);
+  } else {
+    if (exportMetrics) {
+      printf(">>> Metrics Info\n");
+      printf("  %-25s | %s\n", "Metrics Server", [metricsURLStr.absoluteString UTF8String]);
+      printf("  %-25s | %s\n", "Metrics Format",
+             [[self stringFromMetricFormat:metricFormat] UTF8String]);
+      printf("  %-25s | %lu\n", "Export Interval (seconds)", metricExportInterval);
+    } else {
```
tnek, line 128: From a quick glance, it looks like adding `--json` to args skips this check. I didn't trace too deeply, since there's an easier fix:

Move the check for exportMetrics being enabled to an earlier section as a guard clause, so you don't have to reason about what happens to the `nil` metrics object returned later on in the case that metrics are disabled.

Source/santad/SNTDaemonControlController.m
```diff
@@ -242,6 +243,18 @@ - (void)setEnableTransitiveRules:(BOOL)enabled reply:(void (^)(void))reply {
   reply();
 }
 
+#pragma mark Metrics Ops
+
+- (void)metrics:(void (^)(NSDictionary *))reply {
+  // If metrics are not enabled send nil back
+  if (![[SNTConfigurator configurator] exportMetrics]) {
```
tnek, line 250: Maybe also LOGI/LOGE here?

Source/santactl/Commands/SNTCommandMetrics.m
```diff
@@ -0,0 +1,180 @@
...
+  return mutableMetrics;
+}
+
+- (void)runWithArguments:(NSArray *)arguments {
+  __block NSDictionary *metrics;
+
+  dispatch_group_t group = dispatch_group_create();
+  dispatch_group_enter(group);
+
+  [[self.daemonConn remoteObjectProxy] metrics:^(NSDictionary *exportedMetrics) {
```
tnek, line 97: Add a unit test to verify that output work as intended (similar to SNTCommandFileInfoTest)

Source/santactl/Commands/SNTCommandMetrics.m
pmarkowsky: Yeah I had a minor tool failure with sourcetree should be fixed.

Source/santactl/Commands/SNTCommandMetrics.m
```diff
@@ -0,0 +1,181 @@
...
+          const char *lastUpdatedStr = [field[@"last_updated"] UTF8String];
+          const char *data = [[NSString stringWithFormat:@"%@", field[@"data"]] UTF8String];
+
+          if (strlen(fieldNameStr) > 0) {
+            printf("  %-25s | %s=%s\n", "Field", fieldNameStr, fieldValueStr);
+          }
+
+          printf("  %-25s | %s\n", "Created", createdStr);
+          printf("  %-25s | %s\n", "Last Updated", lastUpdatedStr);
+          printf("  %-25s | %s\n", "Data", data);
```
russellhancox, line 171: nit: I'd 'label' this Value rather than Data.

Source/santactl/Commands/SNTCommandMetrics.m
```diff
@@ -0,0 +1,181 @@
...
++ (NSString *)shortHelpText {
+  return @"Show Santa metric information.";
+}
+
++ (NSString *)longHelpText {
+  return (@"Provides metrics about Santa's operation while it's running.\n"
+          @"  Use --json to output in JSON format");
+}
+
+- (NSString *)stringFromMetricFormat:(SNTMetricFormatType)format {
```
russellhancox, line 49: I think this method should live with the definition of SNTMetricFormatType, so we don't add another type and forget to include it here.

Source/santactl/Commands/SNTCommandMetrics.m
```diff
@@ -0,0 +1,181 @@
...
+  switch (format) {
+    case SNTMetricFormatTypeMonarchJSON: return @"monarchjson";
+    case SNTMetricFormatTypeRawJSON: return @"rawjson";
+    default: return @"Unknown Metric Format";
+  }
+}
+
+- (NSDictionary *)normalize:(NSDictionary *)metrics {
```
russellhancox, line 57: Don't we have a nearly identical method somewhere already?

Source/santad/SNTDaemonControlController.m
pmarkowsky: I could see LOGD, but if you don't have metrics configured but run the metrics command then this should just return.

Source/santactl/Commands/SNTCommandMetrics.m
pmarkowsky: so we're using the `Value` key as the field value and data as the contents in `SNTMetricSet` so I was trying to be consistent with the existing code.

Source/santactl/Commands/SNTCommandMetrics.m
pmarkowsky: We do. I rewrote it here in case we wanted to have normalization change for human consumption and I'm not sure we'd hit the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) yet.

Source/santactl/Commands/SNTCommandMetrics.m
pmarkowsky: Added tests in the latest commit.

Source/santactl/Commands/SNTCommandMetrics.m
pmarkowsky: This is done in the latest commit.

Source/santactl/Commands/SNTCommandMetrics.m
pmarkowsky: I've moved this function to `common/SNTMetricSet.m`

russellhancox (approved):

Source/santactl/Commands/SNTCommandMetricsTest.m
```diff
@@ -32,12 +36,24 @@ - (void)setUp {
...
+  OCMStub([self.mockConfigurator metricFormat])
+      .andReturn(SNTMetricFormatTypeMonarchJSON);
+  OCMStub([self.mockConfigurator metricURL])
+      .andReturn([NSURL URLWithString:@"http://localhost:2444/submit"]);
+  OCMStub([self.mockConfigurator metricExportInterval])
+      .andReturn((NSUInteger)30);
 }
 
 - (void)tearDown {
+  [self.mockConfigurator stopMocking];
```
russellhancox, line 56: You don't need to (and shouldn't) call stopMocking unless you're doing so mid-test.

russellhancox (changes requested):

tnek (approved):

Source/santactl/Commands/SNTCommandMetricsTest.m
pmarkowsky: Yup. Removed.

russellhancox (approved):

PR #697, opened 2021-12-13
Fix: santactl metrics command behavior

pmarkowsky:
Removed the check for export metrics in santad. Metrics are always collected but only exported to a monitoring system when all of the necessary config options are set. Since they're always collected `santactl metrics` should always return metrics data.

Related issues: #563


russellhancox (approved):

pmarkowsky:
Remove the check for export metrics in santad

Metrics are always collected but only exported to a monitoring system when all of the necessary config options are set. Since they're always collected santactl metrics should always return metrics data.

PR #698, opened 2021-12-13
santad: Fix PrinterProxy workaround for Monterey+

russellhancox:
I also ran the linter and it fixed a few BUILD files.


pmarkowsky (approved):
LGTM.

PR #704, opened 2021-12-25
Update hedron_compile_commands

cpsauer:
Hey guys, hedron_compile_commands creator here. Smiled (as an ex-Googler) seeing that you'd set it up in #690. Thanks for giving it a try!

Figured I'd PR an update to the latest while I was here--esp since I just fixed an issue someone else was having.
I'd love it if you'd let me know how things go!

Chris

P.S. Couple pieces of unsolicited advice: Renovate is pretty sweet for automating keeping things up to date. And things might work even better with `refresh_compile_commands`.


russellhancox:
Thank you Chris for the PR and for the project.

I was planning to look at Renovate early in the new year, for various reasons it might not be simple for us to adopt but I'd like to try. As for things working better with `refresh_compile_commands`, is that referring to adding a rule in our top-level BUILD file? I initially added that but using the fallback seemed to work equally well but I probably missed something?

cpsauer:
Oh great! If it's working equally well, then not a problem :)
[I'm realizing now, belatedly, that the benefit is mostly when you're cross compiling, but here the host and target platform are the same.]

russellhancox (approved):

PR #706, opened 2022-01-06
Project: Explicitly set calendar on ISO8601 dates

russellhancox:
I left the `NSDateFormatter`'s from `santactl status` and `santactl fileinfo` as they were as the dates they produce are user-facing.


pmarkowsky (approved):
LGTM.

PR #707, opened 2022-01-11
santad: Don't use proc_pidpath when using ES

### LINKED ISSUE #642: Remove unnecessary call to `proc_pidpath`
When Santa receives its events from the EndpointSecurity framework, the path is provided as part of the `es_message_t`. Doing this extra lookup in Santa is unnecessary and prone to failure as `proc_pidpath` will not be able to return info for short-lived processes that may have already exited by the time the call is made. Note: This information still needs to be looked up when the Santa driver is in use as the kext doesn't grab this information before sending data up to the daemon.

russellhancox:
Fixes #642


pmarkowsky (approved):
LGTM.

PR #708, opened 2022-01-11
Sync: Allow configuring proxies

### LINKED ISSUE #672: Santa Proxy Support
Santa does not seem to support HTTP Connect or SOCKS proxies.  Is this on the roadmap or can it be added?

According to NSURLSession (https://developer.apple.com/documentation/foundation/nsurlsession) it is supported as a system wide setting, but for our use case it has to be a Santa specific setting.

russellhancox:
Fixes #672

### DIFF
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -255,6 +255,14 @@
 @property(readonly, nonatomic) NSURL *syncBaseURL;
 
 ///
+///  Proxy settings for syncing.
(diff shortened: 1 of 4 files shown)


pmarkowsky (commented):
Looks good. I think we just need to add update the .md to document the feature.

russellhancox:
> Looks good. I think we just need to add update the .md to document the feature.

Good call, done.

pmarkowsky (approved):
LGTM.

PR #718, opened 2022-01-21
rule download: return early on daemon timeout

tburgin:
* This fixes a potential bug where the postflight could run even though the rules didn't apply.

### DIFF
--- a/Source/santasyncservice/SNTSyncRuleDownload.m
+++ b/Source/santasyncservice/SNTSyncRuleDownload.m
@@ -46,7 +46,7 @@
                                                         dispatch_semaphore_signal(sema);
                                                       }];
   if (dispatch_semaphore_wait(sema, dispatch_time(DISPATCH_TIME_NOW, 300 * NSEC_PER_SEC))) {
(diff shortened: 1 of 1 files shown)


Source/santasyncservice/SNTSyncRuleDownload.m
```diff
@@ -45,7 +45,10 @@ - (BOOL)sync {
                                                         error = e;
                                                         dispatch_semaphore_signal(sema);
                                                       }];
-  dispatch_semaphore_wait(sema, dispatch_time(DISPATCH_TIME_NOW, 300 * NSEC_PER_SEC));
+  if (dispatch_semaphore_wait(sema, dispatch_time(DISPATCH_TIME_NOW, 300 * NSEC_PER_SEC))) {
+    LOGE(@"Failed to add rule(s) to database: timeout connecting to daemon");
```
russellhancox, line 49: nit: `connecting to` should really be `sending rules to daemon` or something

tburgin: Done!

russellhancox (approved):

PR #723, opened 2022-02-02
Fix additional strlcpy issue, simplify call paths

mlw:
This PR streamlines the call paths the utilize `strlcpy`, removing an unnecessary `length` parameter and making use of an existing helper function. This also fixes an off-by-one issue in an existing call to `strlcpy`.

### DIFF
--- a/Source/santad/EventProviders/SNTEndpointSecurityManager.h
+++ b/Source/santad/EventProviders/SNTEndpointSecurityManager.h
@@ -30,10 +30,6 @@
 // The populated buffer will be NUL terminated.
 + (BOOL)populateBufferFromESFile:(es_file_t *)file buffer:(char *)buffer size:(size_t)size;
 
-// Returns YES if the path was truncated.
-// The populated buffer will be NUL terminated.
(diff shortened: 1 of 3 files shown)


Source/santad/EventProviders/SNTEndpointSecurityManager.h
```diff
@@ -19,16 +19,21 @@
...
 - (void)setIsCompilerPID:(pid_t)pid;
 - (void)setNotCompilerPID:(pid_t)pid;
 
+// Returns YES if the path was truncated.
+// The populated buffer will be NUL terminated.
++ (BOOL)populateBufferFromESFile:(es_file_t *)file buffer:(char *)buffer size:(size_t)size;
+
+// Returns YES if the path was truncated.
+// The populated buffer will be NUL terminated.
++ (BOOL)populateBufferFromString:(const char *)string buffer:(char *)buffer size:(size_t)size;
```
tnek, line 35: No need to export this since it's only used by SNTEndpointSecurityManager

tnek (dismissed):

russellhancox (approved):

PR #728, opened 2022-02-10
Add a USB device blocking popup.

tnek:
* Refactor SNTNotificationManager and SNTMessageWindowController for generalized notification logic
* Add the following popup:

image

### DIFF
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -569,10 +569,17 @@ static NSString *const kSyncCleanRequired = @"SyncCleanRequired";
 }
 
 - (NSString *)bannedUSBBlockMessage {
+  if (!self.configState[kBannedUSBBlockMessage]) {
+    return @"The following device has been blocked from mounting.";
+  }
+
   return self.configState[kBannedUSBBlockMessage];
 }
 
 - (NSString *)remountUSBBlockMessage {
+  if (!self.configState[kRemountUSBBlockMessage]) {
+    return @"The following device has been remounted with reduced permissions.";
+  }
   return self.configState[kRemountUSBBlockMessage];
 }
 
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -66,6 +66,9 @@ static NSString *const kEventDetailURLKey = @"EventDetailURL";
 static NSString *const kEventDetailTextKey = @"EventDetailText";
 static NSString *const kUnknownBlockMessage = @"UnknownBlockMessage";
 static NSString *const kBannedBlockMessage = @"BannedBlockMessage";
+static NSString *const kBannedUSBBlockMessage = @"BannedUSBBlockMessage";
+static NSString *const kRemountUSBBlockMessage = @"RemountUSBBlockMessage";
+
 static NSString *const kModeNotificationMonitor = @"ModeNotificationMonitor";
 static NSString *const kModeNotificationLockdown = @"ModeNotificationLockdown";
 
@@ -162,6 +165,8 @@ static NSString *const kSyncCleanRequired = @"SyncCleanRequired";
       kEventDetailTextKey : string,
       kUnknownBlockMessage : string,
       kBannedBlockMessage : string,
+      kBannedUSBBlockMessage : string,
+      kRemountUSBBlockMessage : string,
       kModeNotificationMonitor : string,
       kModeNotificationLockdown : string,
       kSyncBaseURLKey : string,
@@ -559,6 +564,14 @@ static NSString *const kSyncCleanRequired = @"SyncCleanRequired";
 
 - (NSString *)bannedBlockMessage {
   return self.configState[kBannedBlockMessage];
+}
+
+- (NSString *)bannedUSBBlockMessage {
+  return self.configState[kBannedUSBBlockMessage];
+}
+
+- (NSString *)remountUSBBlockMessage {
+  return self.configState[kRemountUSBBlockMessage];
 }
 
 - (NSString *)modeNotificationMonitor {
--- a/Source/common/SNTDeviceEvent.h
+++ b/Source/common/SNTDeviceEvent.h
@@ -16,6 +16,8 @@
 
 @interface SNTDeviceEvent : NSObject <NSSecureCoding>
 
+- (instancetype)initWithOnName:(NSString *)mntonname fromName:(NSString *)mntfromname;
+
 @property NSString *mntonname;
 @property NSString *mntfromname;
 @property NSArray<NSString *> *remountArgs;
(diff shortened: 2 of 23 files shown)


tnek:
Done. They'll be configurable with the `BannedUSBBlockMessage` and `RemountUSBBlockMessage` configurator keys. Here's what these prompts look like now:

I'm struggling with phrasing for human-readable conversions of the remount permissions, if you have ideas for what's clearest for users. 

Screen Shot 2022-02-15 at 3 22 33 PM
Screen Shot 2022-02-15 at 2 58 55 PM

Source/common/SNTConfigurator.m
```diff
@@ -563,6 +568,14 @@ - (NSString *)bannedBlockMessage {
   return self.configState[kBannedBlockMessage];
 }
 
+- (NSString *)bannedUSBBlockMessage {
+  return self.configState[kBannedUSBBlockMessage];
```
mlw, line 572: Comment in `SNTConfigurator.h` says a reasonable default is provided if not configured, but no default is provided.

Source/common/SNTConfigurator.m
```diff
@@ -563,6 +568,14 @@ - (NSString *)bannedBlockMessage {
...
+}
+
+- (NSString *)remountUSBBlockMessage {
+  return self.configState[kRemountUSBBlockMessage];
```
mlw, line 576: Same as above - Comment in `SNTConfigurator.h` says a reasonable default is provided if not configured, but no default is provided.

Source/common/SNTDeviceEvent.m
```diff
@@ -0,0 +1,54 @@
+#import "Source/common/SNTDeviceEvent.h"
+
+@implementation SNTDeviceEvent
```
mlw, line 3: Would strongly recommend adding a custom init function that takes all of the properties as arguments so as to avoid partially initialized objects. Looks like the only current allocation of this object is down in `SNTDeviceManager.mm` and there isn't doesn't appear to be a reason for partial initialization.

Source/common/SNTConfigurator.m
tnek: A little bit of indirection here that also holds true for the other parameters and block messages like this: the actual reasonable default is visible from [SNTDeviceMessageWindowController attributedCustomMessage]. 

I'm open to moving the default here and updating all of the other cases of this with other block messages all at once in a followup PR if that sounds better.

Source/santa/SNTBinaryMessageWindowController.m
```diff
@@ -0,0 +1,175 @@
...
+      NSProgress *progress = object;
+      if (progress.fractionCompleted != 0.0) {
+        self.hashingIndicator.indeterminate = NO;
+      }
+      self.hashingIndicator.doubleValue = progress.fractionCompleted;
+    });
+  }
+}
+
+- (void)loadWindow {
```
mlw, line 76: Should there be a call to `[super loadWindow];` here? One exists in `SNTDeviceMessageWindowController` as well as the base class `SNTMessageWindowController`.

Source/common/SNTConfigurator.m
mlw: I see that now that I got lower in the PR :)

I'd probably lean slightly more in favor of the default being provided by the configurator class just so its a common place to look for defaults.

As a slight tangent - I wonder if it makes sense for this message to be completely controlled. Meaning what if the current "default" messages are _**always**_ displayed. And then the configuration value is additional information. Could be useful if some organization wants to provide additional information to users such as a person to contact or policy information. Thoughts?

Source/santa/SNTDeviceMessageWindowController.m
```diff
@@ -0,0 +1,65 @@
...
+  if (self.event.remountArgs && [self.event.remountArgs count] > 0) {
+    return [SNTBlockMessage
+      formatMessage:@"The following device has been remounted with reduced permissions."];
+  }
+
+  return [SNTBlockMessage formatMessage:@"The following device has been blocked from mounting."];
+}
+
+- (NSString *)messageHash {
+  return self.event.mntonname;
```
mlw, line 60: We should chat about best hashing strategies here.

Source/santad/SNTApplication.m
```diff
@@ -92,12 +93,27 @@ - (instancetype)init {
...
-      self.deviceManager.remountArgs = [configurator remountUSBMode];
+      deviceManager.remountArgs = [configurator remountUSBMode];
     }
 
+    NSString *deviceBlockMsg = deviceManager.remountArgs != nil
+                                 ? [configurator remountUSBBlockMessage]
+                                 : [configurator bannedUSBBlockMessage];
+
+    deviceManager.deviceBlockCallback = ^(SNTDeviceEvent *event) {
+      NSLog(@"wtf '%@' '%@'", event.mntfromname, event.mntonname);
```
mlw, line 109: Prolly should remove :)

Source/santa/SNTBinaryMessageWindowController.m
tnek: Done.

Source/santa/SNTDeviceMessageWindowController.m
tnek: Fixed it :)

Source/santad/SNTApplication.m
tnek: Oops, good catch. Done

Source/common/SNTConfigurator.m
tnek: The one case where I think this might fail is if the notification isn't in English, and the custom message needs to convey the same information but in another language.

I've moved this default to the configurator class to these modified keys. I'll do the same for all the other messages with defaults in a separate PR.

Source/common/SNTDeviceEvent.m
tnek: Done.

Source/common/SNTDeviceEvent.h
```diff
@@ -0,0 +1,27 @@
...
+///    distributed under the License is distributed on an "AS IS" BASIS,
+///    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+///    See the License for the specific language governing permissions and
+///    limitations under the License.
+
+#import <Foundation/Foundation.h>
+
+@interface SNTDeviceEvent : NSObject <NSSecureCoding>
+
+- (instancetype)initWithOnName:(NSString *)mntonname fromName:(NSString *)mntfromname;
```
mlw, line 19: Would recommend also going ahead and taking the `remountArgs` as well unless you have a good reason to leave them off.

mlw (approved):

PR #753, opened 2022-03-15
Move from ASL logging to using os_log directly

mlw:


Source/common/SNTLogging.h
```diff
@@ -19,38 +19,12 @@
...
-/// Simple logging macros
-#define LOGD(logFormat, ...) logMessage(LOG_LEVEL_DEBUG, stdout, logFormat, ##__VA_ARGS__)
-#define LOGI(logFormat, ...) logMessage(LOG_LEVEL_INFO, stdout, logFormat, ##__VA_ARGS__)
-#define LOGW(logFormat, ...) logMessage(LOG_LEVEL_WARN, stderr, logFormat, ##__VA_ARGS__)
-#define LOGE(logFormat, ...) logMessage(LOG_LEVEL_ERROR, stderr, logFormat, ##__VA_ARGS__)
-
-#ifdef __cplusplus
-}  // extern C
-#endif
+#define LOG_WITH_TYPE(type, fmt, ...) os_log_with_type(OS_LOG_DEFAULT, type, fmt, ##__VA_ARGS__)
```
russellhancox, line 24: Does `os_log_with_type()` print to stdout/stderr? All the output from `santactl sync` is written using the log macros which currently print to stdout/stderr (depending on the log level) unless the `--syslog` flag is passed. This PR is removing the flag, which is fine as long as it also prints to stdout/stderr, we'll just end up with the logs in syslog too but if it doesn't the command will output nothing and we'll have no way to debug sync issues.

mlw: So I made some apparently wrong assumptions about logging in Santa - namely that logging != printing, but that isn’t true. To support the current set of features, we’’d need to keep a couple of extra allocations and deal with `va_args`, minimizing the overall perf win this PR presents (though still would be a win to skip the ASL-->ULS conversion). There are some other oddities with the exiting code that we should probably address (e.g. why is only the santad daemon forced into syslog but not other daemons? Also log level filtering in code rarely makes sense using ULS, these should be filtered by whatever is monitoring the logs).

We should move away from ASL as it’s deprecated, but we should probably chat about how we might clean things up here long term. I’m going to close this PR for now while we work on a better long term plan.

Note: It is possible to have `os_log` mirror output to `stderr` by setting the environment variable `OS_ACTIVITY_DT_MODE`. However it looks like the current code wants to sometimes distinguish between `stdout` and `stderr` so this wouldn’t be good enough.

PR #761, opened 2022-03-18
SantaCache: Use new/delete instead of malloc/free, use std::atomic, fix warnings

russellhancox:


Source/common/SNTPrefixTree.cc
```diff
@@ -85,7 +85,7 @@ IOReturn SNTPrefixTree::AddPrefix(const char *prefix, uint64_t *node_count) {
       }
 
       // Is there enough room for the rest of the prefix?
-      if ((node_count_ + (len - i)) > max_nodes_) {
+      if ((node_count_ + (len - (size_t)i)) > max_nodes_) {
```
tnek, line 88: Is there a reason for adding this cast?

Source/common/SantaCache.h
```diff
@@ -79,34 +70,33 @@ class SantaCache {
         (1 << (32 -
                __builtin_clz((((uint32_t)max_size_ / per_bucket) - 1) ?: 1)));
     if (unlikely(bucket_count_ > UINT32_MAX)) bucket_count_ = UINT32_MAX;
-    buckets_ = (struct bucket *)malloc(bucket_count_ * sizeof(struct bucket));
-    bzero(buckets_, bucket_count_ * sizeof(struct bucket));
+    buckets_ = new struct bucket[bucket_count_];
```
tnek, line 73: How about making this a unique_ptr instead of dealing with new/delete?

Source/common/SNTPrefixTree.cc
russellhancox: This is reviving an old branch, I'm not sure what that was for. Reverted.

tburgin (commented):
Now that we are in user space is there a concurrent unordered map implementation we could use instead?

I wonder how much of a performance hit we would take if we just serialize a std:: unordered_map?

Source/common/SantaCache.h
russellhancox: Done

russellhancox:
> Now that we are in user space is there a concurrent unordered map implementation we could use instead?

I'm not aware of any that are open-source, freely licensed and easily integrated.

> I wonder how much of a performance hit we would take if we just serialize a std:: unordered_map?

I think we should thoroughly benchmark this, it _might_ be fine to switch but SantaCache has been working fine and isn't much burden to maintain so we should only switch if std::unordered_map (or an equally simple alternative) offers equal or better performance.

pmarkowsky:
I'd second, using `std::unorderd_map`, if it benchmarks as comparable or better. That said do we have a good benchmark?

>I'm not aware of any that are open-source, freely licensed and easily integrated.

I've heard good things about https://github.com/greg7mdp/parallel-hashmap, but not used it before. 

It's supposed to be a header only, a drop-in replacement for `std::unordered_map`.

There's also the [abseil hashmaps](https://abseil.io/docs/cpp/guides/container)

russellhancox:
I'm gonna drop this PR: I intended it as a simple cleanup now that we don't need to support kernel-space but the tests are failing 1-in-1000 times and I'm not certain why. I don't think spending much more time on this is a worthwhile investment and we should instead write a benchmark to consider replacing SantaCache with `phmap::parallel_flat_hash_map` or `abseil::flat_hash_map`

PR #763, opened 2022-03-21
santactl/metrics: Allow filtering metrics

russellhancox:
And don't print metric type number if the name is known.


pmarkowsky (approved):
LGTM.

PR #765, opened 2022-03-22
santad: Split ES cache into root/non-root varieties

russellhancox:


Source/santad/EventProviders/SNTCachingEndpointSecurityManager.mm
```diff
@@ -29,21 +29,30 @@ uint64_t GetCurrentUptime() {
...
 
 - (instancetype)init {
   self = [super init];
   if (self) {
-    // TODO(rah): Consider splitting into root/non-root cache
-    _decisionCache = new SantaCache<santa_vnode_id_t, uint64_t>();
+    _rootDecisionCache = new SantaCache<santa_vnode_id_t, uint64_t>();
+    _nonRootDecisionCache = new SantaCache<santa_vnode_id_t, uint64_t>();
+
+    // If we fail to retrieve the root vnode ID we'll put everything in the root cache.
```
tburgin, line 43: The comment is a bit hard to make sense of since the code below it is checking the success branch not the failure the comment is referring to. Maybe expand on this a bit. Something like: "Because `_rootVnodeID` is default initialized to zero and we check against a zero `_rootVnodeID` in `-[cacheForVnodeID:]`"

Aside...does this happen in practice? `com.google.santa.daemon` should always have full disk access when running as an ES client.

tburgin (approved):

tburgin (approved):

Source/santad/EventProviders/SNTCachingEndpointSecurityManager.mm
russellhancox: I've added a couple comments to clarify this.

> Aside...does this happen in practice? 

I don't expect this to ever happen but `stat()` can _in theory_ fail.

PR #772, opened 2022-03-25
Modified build target names for santa proto

mlw:
Modified target names to better match style of other projects


pmarkowsky (approved):
LGTM.

PR #781, opened 2022-04-12
Disable layering check for Objective-C

googlewalt:
The Objective-C targets in these packages are currently not layering
check clean.  Disabling layering check ensures that they will continue
to build even when we enable layering check for Objective-C within
Google.

We also need to reenable layering_check on proto_library to avoid
internal breakges, due to 1. proto_library still implying cpp for some
code paths, and 2. disabling cpp layering check requires being on an
allowlist.


russellhancox (approved):

russellhancox:
Thanks!

PR #783, opened 2022-04-14
Enhanced Proto Draft

mlw:


Source/common/new.santa.proto
```diff
@@ -0,0 +1,283 @@
...
+  string name = 2;
+}
+
+message ProcessID {
+  int32 pid = 1;
+  int32 pidversion = 2;
+}
+
+message NanoTime {
+    int64 sec;
```
russellhancox, line 25: remove 2 spaces, same for nsec and same for the MicroTime message

Source/common/new.santa.proto
```diff
@@ -0,0 +1,283 @@
...
+  uint32 mode = 2;
+  uint32 nlink = 3;
+  uint64 ino = 4;
+  UserInfo user = 5;
+  GroupInfo group = 6;
+  int32 rdev = 7;
+  NanoTime a_time = 8;
+  NanoTime m_time = 9;
+  NanoTime c_time = 10;
+  NanoTime b_time = 11;
```
russellhancox, line 51: Expand out the field names. 

Also, are all 4 of these useful to record? I didn't think macOS ever recorded access times and I'm unsure whether there's value in recording both modify and change (the latter seems more useful, to me). Each one we record is up to 17 bytes, which adds up with the number of events we generate.

Source/common/new.santa.proto
```diff
@@ -0,0 +1,283 @@
...
+  int64 size = 12;
+  int64 blocks = 13;
+  int32 blksize = 14;
+  uint32 flags = 15;
+  int32 gen = 16;
```
russellhancox, line 56: Is this field likely to be useful?

Source/common/new.santa.proto
```diff
@@ -0,0 +1,283 @@
...
+  int32 session_id = 6;
+  UserInfo effective_user = 7;
+  GroupInfo effective_group = 8;
+  UserInfo real_user = 9;
+  GroupInfo real_group = 10;
+
+  bool is_platform = 11;
+  bool is_es_client = 12;
+  CodeSignature code_signature = 13;
+  uint32 cs_flags = 14;
```
russellhancox, line 100: Could this be included in the CodeSignature message?

Source/common/new.santa.proto
```diff
@@ -0,0 +1,283 @@
...
+    DECISION_ALLOW = 1;
+    DECISION_DENY = 2;
+  }
+
+  enum Reason {
+    REASON_UNKNOWN = 0;
+    REASON_BINARY = 1;
+    REASON_CERT = 2;
+    REASON_COMPILER = 3;
+    REASON_NOT_RUNNING = 4;
```
russellhancox, line 125: I think we can scrap this reason, I don't think it's possible under ES - it was used with the kext to signify an execution we were logging had happened before the client was connected.

Source/common/new.santa.proto
```diff
@@ -0,0 +1,283 @@
...
+    REASON_PENDING_TRANSITIVE = 5;
+    REASON_SCOPE = 6;
+    REASON_TEAM_ID = 7;
+    REASON_TRANSITIVE = 8;
+    // TODO: Support this
+    REASON_OTHER_ES_CLIENT = 9;
```
russellhancox, line 131: Is this a thing? We ignore fileops from other es clients but I don't know that we'd want to do the same for exec.

Source/common/new.santa.proto
```diff
@@ -0,0 +1,283 @@
...
+  enum Mode {
+    MODE_UNKNOWN = 0;
+    MODE_LOCKDOWN = 1;
+    MODE_MONITOR = 2;
+  }
+
+  ProcessInfo instigator = 1;
+  ProcessInfo target = 2;
+
+  File script = 3;
```
russellhancox, line 143: Where does this come from?

Source/common/new.santa.proto
```diff
@@ -0,0 +1,283 @@
...
+  File workingDirectory = 4;
+
+  repeated string args = 5;
+  repeated string envs = 6;
+  repeated FileDescriptor fds = 7;
```
russellhancox, line 148: I'm struggling to understand the value of this - we'll be able to see that every exec has a number of file descriptors at exec time but we won't know what files those descriptors are for, only the type. Am I missing something?

Source/common/new.santa.proto
```diff
@@ -0,0 +1,283 @@
...
+  MicroTime start_time = 17;
+}
+
+message CertificateInfo {
+  Hash hash = 1;
+  string common_name = 2;
+}
+
+message Execution {
+  enum Decision {
```
russellhancox, line 114: nit: I generally find it easier to comprehend embedded enums when they're directly above the field that uses them.

Source/common/new.santa.proto
```diff
@@ -0,0 +1,283 @@
...
+  string quarantine_url = 13;
+  string original_path = 14;
+}
+
+message Fork {
+  ProcessInfo instigator = 1;
+  ProcessInfo child = 2;
+}
+
+// If ExitType is EXITED, `value` is the exit status
```
russellhancox, line 165: Consider using a oneof to document this:

```proto
message Exited {
  int32 exit_status = 1;
}
message Signaled {
  int32 signal = 1;
}
oneof ExitType {
  Exited exited = 1;
  Signaled signaled = 2;
  Signaled stopped = 3;
}
```

Source/common/new.santa.proto
```diff
@@ -0,0 +1,283 @@
...
+  int32 pid = 1;
+  int32 pidversion = 2;
+  string path = 3;
+  Hash hash = 4;
+}
+
+message SantaMessage {
+  // Unique, per-event generated.
+  // Can allow client to relate events across time
+  string uuid = 1;
```
russellhancox, line 254: Are we going to generate these or do they come from ES?

Source/common/new.santa.proto
mlw: Depends on the FS. I think it's more useful with NFS (?). I don't think it's used by APFS (?).

Source/common/new.santa.proto
mlw: It can be useful, really depends on the context and the type of signal someone is looking for.

This proto is definitely more verbose that what Santa has historically logged. The assumption is that when deployed, it's on system with reasonable upload bandwidth and proper log rotation in place. Personally, I'd hesitate to try and eliminate specific fields from things like `stat` to save a few bytes  given it could be deployed in an environment where that signal was important. But I understand the argument and happy to chat more offline.

Source/common/new.santa.proto
mlw: I think it makes more sense to keep here. The flags are a property of a process not code signature. E.g., in the future we could add a `CodeSignature` field to the `File` message, but `cs_flags` wouldn't apply.

Source/common/new.santa.proto
mlw: May or may not be good to add. I was thinking about what happened if we got a NOTIFY EXEC event for a process that was denied and how that would be represented in Santa. Haven't tracked it down, but will look more into it before committing this.

Source/common/new.santa.proto
mlw: It's part of `es_event_exec_t`. It is provided if the shell image activator is used in the kernel (basically, if someone does `./foo.sh`, ES can report the process (e.g. `/bin/sh` or whatever was in the `#!` line) and the script info).

Source/common/new.santa.proto
mlw: We chatted a bit offline. There could be interesting signal from the types of descriptors a process has at proc start time. Will continue to evaluate if we think it's useful to include for v1 of this proto or maybe hold off.

Source/common/new.santa.proto
mlw: I like it

Source/common/new.santa.proto
mlw: We would generate these. Can chat if we think it's useful or not. Could be used to tie multiple events together in the future (as a sort of FK). Also for a search index on a backend.

Source/common/new.santa.proto
russellhancox: SG

Source/common/new.santa.proto
russellhancox: Fair enough, then just expand the field names.

Source/common/new.santa.proto
russellhancox: Ah I forgot that was a thing, thanks!

PR #784, opened 2022-04-15
Fix event team ID decision value

np5:
We are getting `"decision": "UNKNOWN"` in the events ATM. This is an attempt at fixing it. I am still having issues trying to test a custom Santa build, so I can only say that it compiles…


russellhancox (approved):

russellhancox:
Thanks!

PR #794, opened 2022-04-29
Created a profiles package so provisioning profiles only need to be in one place.

pmarkowsky:
This PR puts all of the provisioning profiles in a top level `profiles` package.

⚠️ This will break your build until you copy the `provisioningprofile`'s to the `/profiles` directory.

### DIFF
--- a/Source/santa/BUILD
+++ b/Source/santa/BUILD
@@ -70,7 +70,7 @@ macos_application(
     minimum_os_version = "10.15",
     provisioning_profile = select({
         "//:ci_build": None,
-        "//conditions:default": "Santa_Dev.provisionprofile",
+        "//conditions:default": "//profiles:santa_dev",
     }),
     version = "//:version",
     visibility = ["//:santa_package_group"],
--- a/profiles/BUILD
+++ b/profiles/BUILD
@@ -1,0 +1,16 @@
+package(
+    default_visibility = ["//:santa_package_group"],
+    features = ["-layering_check"],
+)
+
+licenses(["notice"])
+
+filegroup(
+  name="santa_dev",
+  srcs=["Santa_Dev.provisionprofile"]
+)
+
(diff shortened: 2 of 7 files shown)


Source/santa/BUILD
```diff
@@ -70,7 +70,7 @@ macos_application(
     minimum_os_version = "10.15",
     provisioning_profile = select({
         "//:ci_build": None,
-        "//conditions:default": "Santa_Dev.provisionprofile",
+        "//conditions:default": "//profiles:santa_dev",
```
russellhancox, line 73: Is there any way we could move this select() into the profiles package?

profiles/BUILD
```diff
@@ -0,0 +1,16 @@
+package(
```
russellhancox, line 1: Any reason not to put these rules in the root BUILD file? We're not going to distribute the profiles so this will be the only file ever in this dir.

pmarkowsky: No real reason,  top level seemed a little cluttered and the package made the rules look clear e.g. `//profiles:santa_dev`.  

The other thought I had  was that we could just write `copy your provisioningprofiles in the profiles directory as Santa_dev and Santa_Daemon_Dev`. in future docs.

russellhancox: Hmm, that latter reason is good enough for me

Source/santa/BUILD
pmarkowsky: Tried this and it's not great since you can't have the select apply as the result of the label. The issue here is that we need either `None` for the rule or a file. Filegroup can't return an empty list and satisfy the `provisioning_profile`.

russellhancox: Ack, was worth asking.

russellhancox (approved):

PR #799, opened 2022-05-04
santad: Add 'null' event logger. Fixes #754

### LINKED ISSUE #754: Flag to disable Logging
What are the options available to disable logging completely. Considering i have already whitelisted what is required. There is no point to write to disk or even syslog it just to monitor the executions which i already know about.

I dont see how i can disable the logging completely in EventLogType or EventLogPath keys in config

Is there a workaround to achieve this

russellhancox:


Source/santad/Logs/SNTEventLog.m
```diff
@@ -447,7 +447,10 @@ + (instancetype)logger {
         logger = [[SNTProtobufEventLog alloc] init];
         break;
       }
-      default: logger = nil;
+      case SNTEventLogTypeNull: {
+        logger = nil;
```
tburgin, line 451: Maybe add a note here explaining why using nil is okay. My understanding is that we are relying on `objc_msgSend` to efficiently ignore the logging messages that will be sent to a nil receiver. This is much cleaner and probably just as efficient as doing a nil check on the return value of `[SNTEventLog logger]` before sending the objc message.

https://github.com/apple-oss-distributions/objc4/blob/488ca1fa9db825a411ba002a7e159bd4aaafc3c4/runtime/Messengers.subproj/objc-msg-x86_64.s#L730

Source/common/SNTConfigurator.h
```diff
@@ -151,9 +151,10 @@
 
 ///
 ///  Defines how event logs are stored. Options are:
-///    SNTEventLogTypeSyslog: Sent to ASL or ULS (if built with the 10.12 SDK or later).
-///    SNTEventLogTypeFilelog: Sent to a file on disk. Use eventLogPath to specify a path.
-///    SNTEventLogTypeProtobuf: (BETA) Sent to a file on disk, using maildir format. Use
+///    SNTEventLogTypeSyslog "syslog": Sent to ASL or ULS (if built with the 10.12 SDK or later).
+///    SNTEventLogTypeFilelog "": Sent to a file on disk. Use eventLogPath to specify a path.
```
tburgin, line 155: The empty string is a bit confusing. Maybe add a note that  `SNTEventLogTypeFilelog` is the default option and doesn't have an associated value in the config. Or maybe we should make a value folks can use, just to be explicit?

russellhancox: Made "file" an option

tburgin (approved):

Source/common/SNTConfigurator.m
```diff
@@ -751,7 +755,7 @@ - (BOOL)enableAllEventUpload {
   NSNumber *n = self.syncState[kEnableAllEventUploadKey];
   if (n) return [n boolValue];
 
-  return [self.configState[kEnableAllEventUploadKey];
+  return [self.configState[kEnableAllEventUploadKey]];
```
tburgin, line 758:
```objc
return self.configState[kEnableAllEventUploadKey];
```

tburgin (approved):

PR #800, opened 2022-05-04
Add config to allow uploading all events

### LINKED ISSUE #689: Monitor-specific agent feature request - ship all_the_events
Pairing with #688 to _stop_ the firehose, turning the firehose on so all events (including allowed via rules, disk mounting, file integrity monitoring) being sent would be of help. I am aware and wanting to be cognisant of this potentially being assumed to turn Santa into less 'jolly old soul' and more... 'nannyware', along the lines of things that could be used as basic application usage monitoring, but the point is that instead of adding endpoint security framework agents this could do the job across the wide range of utility Santa provides. A straightforward benefit would be the ability to make troubleshooting mis-applied rules easier without needing a shell to run `santactl fileinfo /binarypath`.

A potentially more obviously useful application of the data we'd gather is for rule maintenance: you can tell when a version of an app has actually become 'extinct' usage-wise so e.g. I can trigger its uninstallation or remove the rule allowing it in the case of arbitrary unsigned binaries (shoutout to my homie homebrew!)

russellhancox:
This config can be enabled by profile or by the sync server and causes the client to upload all events, including those that are explicitly allowed.

Fixes #689

### DIFF
--- a/Source/santad/SNTExecutionController.m
+++ b/Source/santad/SNTExecutionController.m
@@ -191,9 +191,11 @@ static NSString *const kPrinterProxyPostMonterey =
   [self incrementEventCounters:cd.decision];
 
   // Log to database if necessary.
-  if (cd.decision != SNTEventStateAllowBinary && cd.decision != SNTEventStateAllowCompiler &&
-      cd.decision != SNTEventStateAllowTransitive && cd.decision != SNTEventStateAllowCertificate &&
-      cd.decision != SNTEventStateAllowTeamID && cd.decision != SNTEventStateAllowScope) {
+  if ([[SNTCnfigurator configurator] enableAllEventUpload] ||
+      (cd.decision != SNTEventStateAllowBinary && cd.decision != SNTEventStateAllowCompiler &&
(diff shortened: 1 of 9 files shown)


Source/santad/SNTExecutionController.m
```diff
@@ -191,9 +191,11 @@ - (void)validateBinaryWithMessage:(santa_message_t)message {
   [self incrementEventCounters:cd.decision];
 
   // Log to database if necessary.
-  if (cd.decision != SNTEventStateAllowBinary && cd.decision != SNTEventStateAllowCompiler &&
-      cd.decision != SNTEventStateAllowTransitive && cd.decision != SNTEventStateAllowCertificate &&
-      cd.decision != SNTEventStateAllowTeamID && cd.decision != SNTEventStateAllowScope) {
+  if ([[SNTCnfigurator configurator] enableAllEventUpload] ||
```
tburgin, line 194: `[SNTConfigurator configurator].enableAllEventUpload`

tburgin (approved):

tburgin (approved):

PR #801, opened 2022-05-05
Fix ES Mock Client Subscription issues

pmarkowsky:
This fixes an issue with the ES mock where it was deleting all clients on an unsubscribe.

### DIFF
--- a/Source/santad/EventProviders/EndpointSecurityTestUtil.h
+++ b/Source/santad/EventProviders/EndpointSecurityTestUtil.h
@@ -73,11 +73,6 @@ API_UNAVAILABLE(ios, tvos, watchos)
 es_return_t es_muted_paths_events(es_client_t * _Nonnull client, 
                       es_muted_paths_t * _Nonnull * _Nullable muted_paths);
 
-API_AVAILABLE(macos(12.0))
-API_UNAVAILABLE(ios, tvos, watchos)
-es_return_t es_muted_paths_events(es_client_t * _Nonnull client, 
-                      es_muted_paths_t * _Nonnull * _Nullable muted_paths);
-
 API_AVAILABLE(macos(10.15))
 API_UNAVAILABLE(ios, tvos, watchos)
 es_respond_result_t es_respond_auth_result(es_client_t *_Nonnull client,
--- a/Source/santad/EventProviders/EndpointSecurityTestUtil.h
+++ b/Source/santad/EventProviders/EndpointSecurityTestUtil.h
@@ -78,10 +78,6 @@ API_UNAVAILABLE(ios, tvos, watchos)
 es_return_t es_muted_paths_events(es_client_t * _Nonnull client, 
                       es_muted_paths_t * _Nonnull * _Nullable muted_paths);
 
-API_AVAILABLE(macos(12.0)) 
-API_UNAVAILABLE(ios, tvos, watchos)
-void es_release_muted_paths(es_muted_paths_t * _Nonnull muted_paths);
-
 API_AVAILABLE(macos(10.15))
 API_UNAVAILABLE(ios, tvos, watchos)
 es_respond_result_t es_respond_auth_result(es_client_t *_Nonnull client,
--- a/Source/santad/EventProviders/EndpointSecurityTestUtil.h
+++ b/Source/santad/EventProviders/EndpointSecurityTestUtil.h
@@ -21,7 +21,8 @@ es_string_token_t MakeStringToken(const NSString *_Nonnull s);
 
 es_file_t MakeESFile(const char *_Nonnull path);
 es_process_t MakeESProcess(es_file_t *_Nonnull esFile);
-es_message_t MakeESMessage(es_event_type_t eventType, es_process_t *_Nonnull instigator, struct timespec ts);
+es_message_t MakeESMessage(es_event_type_t eventType, es_process_t *_Nonnull instigator,
+                           struct timespec ts);
 CF_EXTERN_C_END
 
 @class ESMessage;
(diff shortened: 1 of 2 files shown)


Source/santad/EventProviders/EndpointSecurityTestUtil.h
```diff
@@ -68,6 +68,20 @@ API_UNAVAILABLE(ios, tvos, watchos)
...
                                      es_handler_block_t _Nonnull handler);
 
+API_AVAILABLE(macos(12.0))
+API_UNAVAILABLE(ios, tvos, watchos)
+es_return_t es_muted_paths_events(es_client_t * _Nonnull client, 
+                      es_muted_paths_t * _Nonnull * _Nullable muted_paths);
+
+API_AVAILABLE(macos(12.0))
+API_UNAVAILABLE(ios, tvos, watchos)
+es_return_t es_muted_paths_events(es_client_t * _Nonnull client,
```
mlw, line 78: duplicate declaration

mlw (changes requested):

mlw (approved):

Source/santad/EventProviders/EndpointSecurityTestUtil.h
```diff
@@ -68,6 +68,11 @@ API_UNAVAILABLE(ios, tvos, watchos)
 es_new_client_result_t es_new_client(es_client_t *_Nullable *_Nonnull client,
                                      es_handler_block_t _Nonnull handler);
 
+API_AVAILABLE(macos(12.0))
+API_UNAVAILABLE(ios, tvos, watchos)
+es_return_t es_muted_paths_events(es_client_t * _Nonnull client, 
+                      es_muted_paths_t * _Nonnull * _Nullable muted_paths);
```
russellhancox, line 74: line up the indentation

Source/santad/EventProviders/EndpointSecurityTestUtil.mm
```diff
@@ -205,6 +205,24 @@ - (void)newClient:(es_client_t *_Nullable *_Nonnull)client
   [self.clients addObject:mockClient];
 }
 
+- (BOOL)removeClient:(es_client_t *_Nonnull) client {
+  MockESClient *clientToRemove = nil;
```
russellhancox, line 209: nil init of objc objects isn't necessary

Source/santad/EventProviders/EndpointSecurityTestUtil.mm
```diff
@@ -205,6 +205,24 @@ - (void)newClient:(es_client_t *_Nullable *_Nonnull)client
...
+  for (MockESClient *c in self.clients) {
+      if (client == (__bridge es_client_t *)c) {
```
russellhancox, line 211: Use isEqual: instead of ==. While the latter _might_ work OK in this case, the default NSObject impl. of isEqual: definitely will and it can be overridden.

Source/santad/EventProviders/EndpointSecurityTestUtil.mm
```diff
@@ -205,6 +205,24 @@ - (void)newClient:(es_client_t *_Nullable *_Nonnull)client
...
+        break;
+      }
+    }
+
+  if (!clientToRemove) {
+    NSLog(@"Attempted to remove unknown mock es client.");
+    return NO;
+  }
+
+  [self.clients removeObject: clientToRemove];
```
russellhancox, line 222: No space after `:`

Source/santad/EventProviders/EndpointSecurityTestUtil.mm
```diff
@@ -281,9 +299,30 @@ es_new_client_result_t es_new_client(es_client_t *_Nullable *_Nonnull client,
   return ES_NEW_CLIENT_RESULT_SUCCESS;
 };
 
+API_AVAILABLE(macos(12.0))
+API_UNAVAILABLE(ios, tvos, watchos)
+es_return_t es_muted_paths_events(es_client_t * _Nonnull client, 
+                      es_muted_paths_t * _Nonnull * _Nullable muted_paths) {
```
russellhancox, line 305: line up indentation

Source/santad/EventProviders/EndpointSecurityTestUtil.mm
```diff
@@ -281,9 +299,30 @@ es_new_client_result_t es_new_client(es_client_t *_Nullable *_Nonnull client,
...
+API_AVAILABLE(macos(12.0)) 
+API_UNAVAILABLE(ios, tvos, watchos)
+void es_release_muted_paths(es_muted_paths_t * _Nonnull muted_paths) {
+  free(muted_paths);
+}
+
 API_AVAILABLE(macos(10.15))
 API_UNAVAILABLE(ios, tvos, watchos) es_return_t es_delete_client(es_client_t *_Nullable client) {
-  [[MockEndpointSecurity mockEndpointSecurity] reset];
+  if (![[MockEndpointSecurity mockEndpointSecurity] removeClient: client]) {
```
russellhancox, line 323: no space after `:`

Source/santad/EventProviders/EndpointSecurityTestUtil.mm
pmarkowsky: In this case I was explicitly looking for pointer equality and wasn't sure if isEqual would be more inclusive.

russellhancox: The default implementation is "equal identity". If there's a reason to care that the object in question is at the same address as before then == is appropriate but warrants a comment about _why_ that's what's desired.

mlw (dismissed):

russellhancox (approved):

Source/santad/EventProviders/EndpointSecurityTestUtil.mm
pmarkowsky: Comment added.

PR #804, opened 2022-05-10
santasyncservice: handle loading and unloading of the service in the pkg

tburgin:
Thanks @bfreezy for notifying us about this issue!


pmarkowsky (approved):
LGTM.

PR #806, opened 2022-05-11
GUI: Switch to UserNotification.framework notifications

russellhancox:


Source/santa/SNTNotificationManager.m
```diff
@@ -207,36 +208,52 @@ - (void)hashBundleBinariesForEvent:(SNTStoredEvent *)event
...
+
+  UNMutableNotificationContent *content = [[UNMutableNotificationContent alloc] init];
+  content.title = @"Santa";
+
   switch (clientmode) {
-    case SNTClientModeMonitor:
-      un.informativeText = @"Switching into Monitor mode";
-      customMsg = [[SNTConfigurator configurator] modeNotificationMonitor];
-      if (!customMsg) break;
-      if (!customMsg.length) return;
```
tburgin, line 219: > if (!customMsg.length) return;

This seems like a bug and I think it's right to fix it the way you did. But just to be sure, do you think anyone is replying on this behavior? If someone puts an empty string in the custom message config they won't get a popup. Seems like a bug.

tburgin (approved):

tburgin (dismissed):

tburgin (approved):

PR #807, opened 2022-05-11
SNTConfigurator: remove mutability from sync state dict

tburgin:
The property doesn't need to be mutable, updating the type.


pmarkowsky (approved):
LGTM.

PR #810, opened 2022-05-11
syncservice: Fix SNTSyncTest

russellhancox:
Failing preflight early if the daemonConn doesn't return a response the tests. This fix is a bit awkward, I tried to add the defaults in setUp but then you can't overwrite the stubs in methods that need to do it


tburgin (approved):
Thanks

PR #816, opened 2022-05-26
Remove unused testing scripts

mlw:
~Fix the build script to look at new preferred location for provisioning profiles~

Update: After discussion - these test scripts are not widely used and do not offer a common way to build/test. Removing them to ensure how everyone builds/tests is synchronized.


russellhancox (commented):
I wasn't aware this script existed and I don't think it's needed - it re-creates a lot of logic that's in the top-level BUILD but is now out of date (there are refs to version.bzl for example). The `:reload` rule does the same thing.

russellhancox (approved):

PR #818, opened 2022-05-27
Added handling for Remount events to USB mass storage blocking

pmarkowsky:
This PR adds logic to handle Remount events from ES for USB mass storage blocking.  Previously a user could simply  remount storage.


Source/santad/EventProviders/SNTDeviceManager.mm
```diff
@@ -199,43 +200,80 @@ - (void)handleAuthMount:(const es_message_t *)m
...
+
+  //TODO: check kind and protocol for banned things (e.g. MTP).
+  LOGD(@"SNTDeviceManager: DiskInfo Protocol: %@", protocol);
+  LOGD(@"SNTDeviceManager: DiskInfo Kind: %@", kind);
+  LOGD(@"SNTDeviceManager: DiskInfo isInternal: %d", isInternal);
+  LOGD(@"SNTDeviceManager: DiskInfo isRemovable: %d", isRemovable);
+  LOGD(@"SNTDeviceManager: DiskInfo isEjectable: %d", isEjectable);
+
+  // If the device isn't a DMG or removable we're ok with the operation.
+  if (isInternal || isDMG || (!isRemovable && !isEjectable && !isUSB)) {
```
russellhancox, line 250: What is the new condition for?

Source/santad/EventProviders/SNTDeviceManager.mm
```diff
@@ -199,43 +200,80 @@ - (void)handleAuthMount:(const es_message_t *)m
...
   }
 
   SNTDeviceEvent *event = [[SNTDeviceEvent alloc]
-    initWithOnName:[NSString stringWithUTF8String:m->event.mount.statfs->f_mntonname]
-          fromName:[NSString stringWithUTF8String:m->event.mount.statfs->f_mntfromname]];
+    initWithOnName:[NSString stringWithUTF8String:eventStatFS->f_mntonname]
+          fromName:[NSString stringWithUTF8String:eventStatFS->f_mntfromname]];
 
-  BOOL shouldRemount = self.remountArgs != nil && [self.remountArgs count] > 0;
+  BOOL shouldRemount = self.remountArgs != nil && [self.remountArgs count] > 0;
```
russellhancox, line 259: Trailing space

Source/santad/EventProviders/SNTDeviceManager.mm
```diff
@@ -297,6 +335,9 @@ - (void)handleESMessageWithTimeout:(const es_message_t *)m
 - (void)handleESMessage:(const es_message_t *)m
              withClient:(es_client_t *)c API_AVAILABLE(macos(10.15)) {
   switch (m->event_type) {
+    case ES_EVENT_TYPE_AUTH_REMOUNT: {
+        [[fallthrough]];
```
russellhancox, line 339: Indent -2 spaces

Source/santad/EventProviders/SNTDeviceManager.mm
```diff
@@ -199,43 +200,80 @@ - (void)handleAuthMount:(const es_message_t *)m
...
 
-  long mountMode = m->event.mount.statfs->f_flags;
+  struct statfs *eventStatFS;
+  BOOL isRemount = NO;
+
+  switch (m->event_type) {
+    case  ES_EVENT_TYPE_AUTH_MOUNT:
+      eventStatFS = m->event.mount.statfs;
+      break;
+    case ES_EVENT_TYPE_AUTH_REMOUNT:
```
russellhancox, line 210: Will the remount that santad initiates come through here? Can we avoid that, if so?

Source/santad/EventProviders/SNTDeviceManager.mm
```diff
@@ -199,43 +200,80 @@ - (void)handleAuthMount:(const es_message_t *)m
...
   // TODO(tnek): Log all of the other attributes available in diskInfo into a structured log format.
   NSDictionary *diskInfo = CFBridgingRelease(DADiskCopyDescription(disk));
+  BOOL isInternal =  [diskInfo[(__bridge NSString *)kDADiskDescriptionDeviceInternalKey] boolValue];
   BOOL isRemovable = [diskInfo[(__bridge NSString *)kDADiskDescriptionMediaRemovableKey] boolValue];
-  BOOL isUSB =
-    [diskInfo[(__bridge NSString *)kDADiskDescriptionDeviceProtocolKey] isEqualTo:@"USB"];
-
-  if (!isRemovable || !isUSB) {
+  BOOL isEjectable = [diskInfo[(__bridge NSString *)kDADiskDescriptionMediaEjectableKey] boolValue];
+  NSString* protocol = diskInfo[(__bridge NSString *)kDADiskDescriptionDeviceProtocolKey];
```
tburgin, line 236: Do we know if other values besides USB and Virtual Interface are possible? 
Are we sure Virtual Interface means DMG in all cases?

pmarkowsky: This gets set in the disk description via the IOKit registry in `kIOPropertyPhysicalInterconnectTypeKey` to`kIOPropertyPhysicalInterconnectTypeVirtual` for DMGs.  I've been testing manually with a few different DMGs. including the Santa-2022.5 release and they've all shown up as `Virtual Interface`.  

Searching through other tools and mailing lists is confirming this check. I'll make it tighter by checking that both `kDADiskDescriptionDeviceModelKey` is equal to "Disk Image" and `kDADiskDescriptionMediaNameKey` is equal to "disk image".  I'll move this check into a method.

Source/santad/EventProviders/SNTDeviceManager.mm
pmarkowsky: Wanted to make sure we weren't popping up for DMGs explicitly or on internal mounts.

Source/santad/EventProviders/SNTDeviceManager.mm
pmarkowsky: One issue I realized in testing the tighter check is that we'd break using RAM disks. The `Virtual Interface` check also covered ram disks. I've updated comments.

Source/santad/EventProviders/SNTDeviceManager.mm
```diff
@@ -199,44 +200,76 @@ - (void)handleAuthMount:(const es_message_t *)m
...
+
+  NSString *kind = diskInfo[(__bridge NSString *)kDADiskDescriptionMediaKindKey];
+
+  // TODO: check kind and protocol for banned things (e.g. MTP).
+  LOGD(@"SNTDeviceManager: DiskInfo Protocol: %@ Kind: %@ isInternal: %d isRemovable: %d "
+       @"isEjectable: %d",
+       protocol, kind, isInternal, isRemovable, isEjectable);
 
-  if (!isRemovable || !isUSB) {
+  // If the device isn't a Virtual device (DMG/ramdisk etc.) or removable we're
```
tburgin, line 244: The comment is hard to parse. Could you update?

Looking at the current logic, maybe something like this?

"If the device is internal or virtual we are okay with the operation. We also are okay with operations for devices that are non-removal as long as they are NOT a USB device."

Source/santad/EventProviders/SNTDeviceManager.mm
pmarkowsky: The remount will come through because we're initiating it via diskarbitrationd. 

https://github.com/google/santa/blob/c7e309ccb149e0b7fba1979d73464a38735b516b/Source/santad/EventProviders/SNTDeviceManager.mm#L256-L257

We could prevent the remount message by doing the mount call in SNTDeviceManager and muting santad like in #820 .  That said we already have mocks for this.

Source/santad/EventProviders/SNTDeviceManager.mm
tburgin: Can we not break RAM disks? They can be useful and as I understand it they are not in scope for USB blocking.

Source/santad/EventProviders/SNTDeviceManager.mm
```diff
@@ -199,43 +200,80 @@ - (void)handleAuthMount:(const es_message_t *)m
...
+  switch (m->event_type) {
+    case  ES_EVENT_TYPE_AUTH_MOUNT:
+      eventStatFS = m->event.mount.statfs;
+      break;
+    case ES_EVENT_TYPE_AUTH_REMOUNT:
+      eventStatFS = m->event.remount.statfs;
+      isRemount = YES;
+      break;
+    default:
+      LOGE(@"Unexpected Event Type passed to DeviceManager handleAuthMount: %d", m->event_type);
```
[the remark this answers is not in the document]
pmarkowsky: Do we have a preference for `assert` vs. `NSAssert`?

Source/santad/EventProviders/SNTDeviceManager.mm
pmarkowsky: Good call, prose updated.

Source/santad/EventProviders/SNTDeviceManager.mm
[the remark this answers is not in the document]
pmarkowsky: I've put an `assert` in below.

Source/santad/EventProviders/SNTDeviceManager.mm
pmarkowsky: I'd say if we're going to redo the mounts let's do that in a future PR.

PR #820, opened 2022-05-27
Mute self to reduce message volume. Remove noisy log message.

mlw:
Grabs the santad audit token and then sends that to `es_mute_process`. This will ensure Santa does not receive notify messages for itself.

Also removed a noisy log message when multiple ES clients are enabled.

### DIFF
--- a/Source/santad/EventProviders/SNTEndpointSecurityManager.mm
+++ b/Source/santad/EventProviders/SNTEndpointSecurityManager.mm
@@ -48,6 +48,7 @@ static const pid_t PID_MAX = 99999;
     _decisionCallback = ^(santa_message_t) {};
     _logCallback = ^(santa_message_t) {};
     [self establishClient];
+    [self muteSelf];
     _prefixTree = new SNTPrefixTree();
     _esAuthQueue =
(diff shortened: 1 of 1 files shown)


Source/santad/EventProviders/SNTEndpointSecurityManager.mm
```diff
@@ -84,9 +98,7 @@ - (void)establishClient API_AVAILABLE(macos(10.15)) {
         if (m->action_type == ES_ACTION_TYPE_AUTH) {
           es_respond_auth_result(self.client, m, ES_AUTH_RESULT_ALLOW, false);
         }
-        if (self.selfPID != pid) {
```
russellhancox, line 87: Can we remove selfPID now? It looks like it's still referred to but those cases might also be removable?

tburgin: +1 for removing these. However, we would need to bail if `-[muteSelf:]` fails.

mlw: I'm all for bailing in cases like these instead of trying to carry on. Simplifies a bunch of logic.

russellhancox (approved):

PR #822, opened 2022-06-06
santad: Copy/retain ES message for use in deadline handler.

### LINKED ISSUE #821: ES message deadline handler bug
I was doing a code review and noticed that the ES AUTH deadline handler does not copy the ES message. According to the ES docs the original message is only guaranteed to be valid within the ES callback scope. The deadline handler executes outside this scope. So this could potentially lead to a missed auth response and crash if the deadline is reached (assuming ES actually free's the scoped message):

`
  es_new_client_result_t ret = es_new_client(&client, ^(es_client_t *c, const es_message_t *m) {
  // ...
          dispatch_after(dispatch_time(m->deadline, NSEC_PER_SEC * -5), self.esAuthQueue, ^(void) {
            if (responded->load()) return;
            LOGE(@"Deadline reached: deny pid=%d ret=%d", pid,
                 es_respond_auth_result(self.client, m, ES_AUTH_RESULT_DENY, false));
          });
          // Copy the message and return control back to ES
          es_message_t *mc = es_copy_message(m);
  //...
  }
`

russellhancox:
Fixes #821

### DIFF
--- a/Source/santad/EventProviders/SNTDeviceManager.mm
+++ b/Source/santad/EventProviders/SNTDeviceManager.mm
@@ -272,6 +272,7 @@ NS_ASSUME_NONNULL_BEGIN
   // This isn't an issue for notify events, so we're in no rush for those.
   es_message_t *mc = es_copy_message(m);
 
+  std::shared_ptr<std::atomic<bool>> responded;
   dispatch_semaphore_t processingSema = dispatch_semaphore_create(1);
   dispatch_semaphore_t deadlineExpiredSema = dispatch_semaphore_create(0);
 
@@ -281,6 +282,7 @@ NS_ASSUME_NONNULL_BEGIN
         // Handler already responded, nothing to do.
         return;
       }
+      if (responded->load()) return;
       LOGE(@"SNTDeviceManager: deadline reached: deny pid=%d ret=%d",
            audit_token_to_pid(m->process->audit_token),
            es_respond_auth_result(c, m, ES_AUTH_RESULT_DENY, false));
@@ -290,6 +292,7 @@ NS_ASSUME_NONNULL_BEGIN
 
   dispatch_async(self.esAuthQueue, ^{
     [self handleESMessage:m withClient:c];
+    responded->store(true);
     if (dispatch_semaphore_wait(processingSema, DISPATCH_TIME_NOW) != 0) {
       // Deadline expired, wait for deadline block to finish.
       dispatch_semaphore_wait(deadlineExpiredSema, DISPATCH_TIME_FOREVER);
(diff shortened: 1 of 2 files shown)


uptycs-brian:
Still a possible race, if the deadline handler executes while the message handler is active (by definition a missed deadline), then the message handler could finish while the deadline handler is active and release the message causing a potential crash in the ES library.

mlw (approved):

uptycs-brian:
@russellhancox You could simplify the sync of the handler and deadline blocks by just using the existing `responded` atomic. No need for a dispatch semaphore. Sorry about doing this in comments, I'm just not setup currently to build Santa.

// deadline
if (responded->exchange(true)) {
return;
}
...
es_free_message(mc);

// handler
[messageHandler:m];
if (!responded->exchange(true)) {
es_free_message(m);
}

russellhancox:
Thanks @uptycs-brian, I completely blanked on that as an option, that's much simpler.

mlw:
There is an issue with the proposal using `responded->exchange` - if the "deadline" block executes while the "processing" block is still down in `messageHandler:`, the "deadline" block frees the message while still being operated on by the "processing" block.

Regarding the dual semaphore approach - another solution instead of using the `responded` atomic, would be to only signal the `processingSema` from the block that "lost" the race waiting on it.

E.g.
```
          dispatch_after(dispatch_time(m->deadline, NSEC_PER_SEC * -5), self.esAuthQueue, ^(void) {
            if (dispatch_semaphore_wait(processingSema, DISPATCH_TIME_NOW) != 0) {
              // Handler has already responded
              // Signal the sema to keep it balanced.
              dispatch_semaphore_signal(processingSema);
              return;
            }

            LOGE(@"SNTEndpointSecurityManager: deadline reached: deny pid=%d ret=%d", pid,
                 es_respond_auth_result(self.client, mc, ES_AUTH_RESULT_DENY, false));
            dispatch_semaphore_signal(deadlineExpiredSema);
          });

          // Dispatch off to the handler and return control to ES.
          dispatch_async(self.esAuthQueue, ^{
            [self messageHandler:mc];
            if (dispatch_semaphore_wait(processingSema, DISPATCH_TIME_NOW) != 0) {
              // Deadline expired, wait for deadline block to finish.
              dispatch_semaphore_wait(deadlineExpiredSema, DISPATCH_TIME_FOREVER);

              // Signal the sema to keep it balanced.
              dispatch_semaphore_signal(processingSema);
            }
            es_free_message(mc);
          });
```

uptycs-brian:
Ugh, yep, good catch. Really need a separate retain for each block. But that's a separate copy on 10.15, which is expensive.

uptycs-brian:
Think I have a solution using `responded`. It involves polling in the pathological deadline case though:

```
auto responded = std::make_shared<std::atomic<int>>(0);
// deadline
if (responded->fetch_add(1) == 0) {
   return; // handled
}
...
// we own the message now; if necessary, wait for the handler to finish
// could dispatch this in a block to the notify queue
while (responded->load() > 1) {
     std::this_thread::sleep_for(std::chrono::milliseconds{50});
 }
es_free_message(mc);

// handler
responded->fetch_add(1);
[messageHandler:m];
if (responded->fetch_sub(1) == 1) {
  es_free_message(mc);
}
// deadline is active, it will release the message

mlw (approved):

PR #827, opened 2022-06-24
Add team ID to synced events

np5:
The team ID cannot be extracted from the signing certificate information, when the binary comes from the App Store. With this PR, we are trying to include it in the event information.

### DIFF
--- a/Source/common/SNTStoredEvent.h
+++ b/Source/common/SNTStoredEvent.h
@@ -95,6 +95,12 @@
 ///
 @property NSArray *signingChain;
 
+
+///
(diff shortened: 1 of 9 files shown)


np5:
Not too happy with the last commit. How would you like me to proceed? Update the fixture to include a Team ID? I also took the liberty to put the extra test after the cert test, for consistency with the other additions in this PR.

pmarkowsky:
@np5 yes please update the fixture to include a Team ID.

np5:
I didn't know really how to proceed, so I simply manually edited the fixture and added a Team ID to the `yes` event.

russellhancox (approved):

PR #840, opened 2022-07-07
adhoc build and run santa

tburgin:
This CL adds a new `SANTA_BUILD_TYPE` named `adhoc`. When building with this type no provisioning profiles or code signing certificates are used. This will be useful for anyone developing Santa who doesn't have access to Google's dev provisioning profiles or code signing certificates.

Add the `--define=SANTA_BUILD_TYPE=adhoc` flag to bazel commands to use this build mode. It works well for the `:reload` command for example:

```
bazel run :reload --define=SANTA_BUILD_TYPE=adhoc
```

Note: if you are currently running a release or non-adhoc dev build of Santa, this new adhoc build will show up as a second instance of Santa. Remove the non-adhoc instance like so:
```
systemextensionsctl uninstall EQHXZ8M8AV com.google.santa.daemon
```

Once CL is landed docs will follow.


tburgin:
*SIP must be disabled to run the adhoc build*

Source/common/SNTXPCControlInterface.m
```diff
@@ -27,10 +27,16 @@
 @implementation SNTXPCControlInterface
 
 + (NSString *)serviceID {
+#ifdef SANTAADHOC
+  // The mach service for an adhoc signed ES sysx uses the "endpoint-security" prefix instead of
+  // the teamid. In Santa's case it will be endpoint-security.com.google.santa.daemon.xpc.
+  return [NSString stringWithFormat:@"endpoint-security.%@.xpc", kBundleID];
```
mlw, line 33: We may want to consider setting `NSEndpointSecurityMachServiceName` (see man page for `EndpointSecurity(7)`. This should mitigate the need for different mach service names.

docs/development/building.md
```diff
@@ -37,123 +37,75 @@ git tag --sort=-creatordate
...
+Non-Google developers can use an adhoc build to run development builds of Santa.
+System Integrity Protection (SIP) will need to be disabled in order to run an
+adhoc build.
 
-```sh
-bazel run :reload -c opt
-```
+**This is only to be done a machine that is actively developing Santa.**
+
+1.  Boot into Recovery Mode: Reboot and hold down `command+r`
```
russellhancox, line 73: Should we document the Apple Silicon version?

Source/common/SNTXPCControlInterface.m
mlw: Something to note tho - this key didn't exist in 10.15, it came a bit later (10.15.4 maybe?)

russellhancox (approved):

Source/common/SNTXPCControlInterface.m
tburgin: Thanks! Let's wait to adopt `NSEndpointSecurityMachServiceName` until we drop support for 10.15.

docs/development/building.md
russellhancox: Oh, it's just power off and then keep holding power for 10s until it goes into recovery

russellhancox: https://support.apple.com/guide/mac-help/macos-recovery-a-mac-apple-silicon-mchl82829c17/mac

Issue #285, opened 2018-07-02
Allow managing ruleset via `.mobileconfig`


jesseendahl:
We manage Santa centrally and we want to ensure that the only rules loaded are **only** the ones that are currently defined in the config. Phrased differently—we want to ensure there no locally set rules: the Santa rules should match the desired state as defined in config management. Currently have to do some crazy stuff to achieve that goal. Specifically, we load a LaunchDaemon that runs a script every five minutes. That script:

- Copies & validates Santa's `rules.db ` to a temporary directory (the only way to get a list of all the configured rules)
- Creates a comparison table containing the configured rules with mandatory (Apple and Santa cert whitelisting) rules
- Compare the two rules tables and use `santactl` to add, remove, or update rules as needed.

All this happens without interrupting Santa, and is fairly lightweight, especially in the event that no differences are found—which is most the time. But we're concerned that it's pretty brittle, since it will break if the database structure changes.

It would be really ideal if we could just set the rules via the `.mobileconfig` and then Santa handled making sure they were loaded/reloaded every time the config changes.

What would also be *really* ideal is if local rule changes were blocked whenever the ruleset is being managed by the `.mobileconfig`.

Curious to hear what y'all think of this idea!

aidanpm:
@jesseendahl  💯 👍 

@tburgin, In the case of a `.mobilesonfig` used to configure Santa with a sync server, it would be nice if rules supplied in that same profile would be used until Santa has completed it's first sync with the configured server. Basically allowing the admin to guarantee a basic list of rules upon install until a synced rule set takes precedence. 

Though I'm not sure how valuable that truly is 🤷‍♂️

russellhancox:
I have a few concerns about this, mostly because our experience with mobileconfig files has been terrible - a change to any installed profile will cause a drop and refresh of all settings for all profiles. We handle this by waiting 5s after noticing a config change before doing anything (and resetting that timer if another change happens before it fires) but currently 'doing anything' is just reading the list of forced values so it's not particularly time consuming if we repeat the process a few times. This change would make 'doing anything' also resetting all rules in the database, which could take a lot longer, especially if the list of rules is large.

An alternative approach:

1. Add a key to the `.mobileconfig` called `RulesFile` which points at a root-owned file on disk with appropriate permissions (probably `0600`). Having this key set will have the same effect as having SyncBaseURL set, insofar as the `rule` command will be disabled.

[...]

PR #846, opened 2022-07-13
santad: Allow configuring a static set of rules via configuration profile

### LINKED ISSUE #777: Permanent (local) whitelist exemptions
Would it be possible to have a small and select number of applications permanently granted exemptions from Santa blocking, perhaps by a local configuration installed with Santa?

When rule sync does not work as expected blocking Mobile Device Management binaries, and being unable to connect from remote because the remote interactive connection application is blocked, are show stoppers.

russellhancox:
This PR adds a new top-level configuration key `StaticRules` that allows setting a static set of rules utilized _before_ any rules in the database. This allows for 2 things:

1. Adding a set of critical rules for management tools that should always be allowed to run even if a sync has not occurred or for some reason the rules database has been cleared.
2. For organizations that do not have a sync server, this can be used for configuring the ruleset.

Fixes #777 and #285


Source/common/SNTRule.m
```diff
@@ -48,6 +49,57 @@ - (instancetype)initWithIdentifier:(NSString *)identifier
...
+// Converts rule information downloaded from the server into a SNTRule.  Because any information
+// not recorded by SNTRule is thrown away here, this method is also responsible for dealing with
+// the extra bundle rule information (bundle_hash & rule_count).
+- (instancetype)initWithDictionary:(NSDictionary *)dict {
+  if (![dict isKindOfClass:[NSDictionary class]]) return nil;
+
+  self = [super init];
+  if (self) {
+    _identifier = dict[kRuleIdentifier];
+    if (_identifier == nil) {
```
pmarkowsky, line 61: I think we should be strict here and require that the dictionary has the correct keys, returning nil otherwise.

Assuming a SHA256 identifier seems problematic since we also take non-hash value identifiers like Team IDs.

Source/common/SNTConfigurator.m
```diff
@@ -549,6 +560,21 @@ - (void)setRemountUSBMode:(NSArray<NSString *> *)args {
   return args;
 }
 
+- (NSDictionary<NSString *, SNTRule *> *)staticRules {
+  NSArray *currentRules = self.configState[kStaticRules];
+  if (currentRules.hash != self.cachedStaticRulesHash) {
+    NSMutableDictionary *rules = [NSMutableDictionary dictionaryWithCapacity:currentRules.count];
+    for (id rule in currentRules) {
+      if (![rule isKindOfClass:[NSDictionary class]]) return self.cachedStaticRules;
+      SNTRule *r = [[SNTRule alloc] initWithDictionary:rule];
```
pmarkowsky, line 569: I think we should log a warning here if the dictionary is malformed, so admins can figure out if they've typo-ed a config.

Source/common/SNTRule.m
```diff
@@ -48,6 +49,57 @@ - (instancetype)initWithIdentifier:(NSString *)identifier
   return self;
 }
 
+// Converts rule information downloaded from the server into a SNTRule.  Because any information
+// not recorded by SNTRule is thrown away here, this method is also responsible for dealing with
+// the extra bundle rule information (bundle_hash & rule_count).
+- (instancetype)initWithDictionary:(NSDictionary *)dict {
```
pmarkowsky, line 55: We should write tests for this since we're kinda relying on it to filter out malformed config data.

pmarkowsky (changes requested):
Overall looks good, some minor things.

Source/common/SNTConfigurator.m
russellhancox: We can't log inside SNTConfigurator as that would create a circular dependency between SNTConfigurator and SNTLogging.

Source/common/SNTRule.m
russellhancox: > I think we should be strict here and require that the dictionary has the correct keys, returning nil otherwise.

I've added a second check that identifier is valid, this was otherwise already the case.

> Assuming a SHA256 identifier seems problematic since we also take non-hash value identifiers like Team IDs.

There are 2 acceptable keys when creating rules: `identifier` and `sha256` - the latter was originally the only valid key, we added `identifier` when team ID rules were added. This condition is for backward compatibility with the old key.

Source/common/SNTConfigurator.m
pmarkowsky: Right, we're using `SNTConfigurator` to set the logging level. Probably a good idea to split this up in the future so that `SNTConfigurator` can set the log level and log without the circular dependency, but outside the scope of this PR.

Source/common/SNTConfigurator.m
```diff
@@ -549,6 +560,21 @@ - (void)setRemountUSBMode:(NSArray<NSString *> *)args {
   return args;
 }
 
+- (NSDictionary<NSString *, SNTRule *> *)staticRules {
+  NSArray *currentRules = self.configState[kStaticRules];
```
tburgin, line 564: We need to make sure this is indeed an NSArray.

Source/common/SNTConfigurator.m
```diff
@@ -549,6 +560,21 @@ - (void)setRemountUSBMode:(NSArray<NSString *> *)args {
...
+  if (currentRules.hash != self.cachedStaticRulesHash) {
```
tburgin, line 565: How expensive is this hashing? For a really large set I can see the cost adding up since we need to hash for every exec. Instead of initializing lazily here we should move the parsing closer to when the rules are added to the profile. We could add a new property for the unprocessed rules and KVO observe changes. We could do the parsing there. This method turns into a fast getter for the current set of rules.

Source/common/SNTRule.m
```diff
@@ -48,6 +49,58 @@ - (instancetype)initWithIdentifier:(NSString *)identifier
...
+  if (![dict isKindOfClass:[NSDictionary class]]) return nil;
+
+  self = [super init];
+  if (self) {
+    _identifier = dict[kRuleIdentifier];
```
tburgin, line 60: Anything we get out of the dictionary we can’t trust. We need to check the types are what we expect.

PR #847, opened 2022-07-14
santad: Improve caching of static rules

russellhancox:
In #846 I forgot that  is only a count of the entries so if the config changes but the number of rules remains the same we would never update the cache. This PR moves the processing of the raw config into the KVO handler code so it is not at all in the hot-path. It also does more checking of types in the rule dictionary processing code.

### DIFF
--- a/Source/common/SNTRule.m
+++ b/Source/common/SNTRule.m
@@ -58,8 +58,9 @@
   self = [super init];
   if (self) {
     _identifier = dict[kRuleIdentifier];
-    if (![_identifier isKindOfClass:[NSString class]] || !_identifier.length)
+    if (![_identifier isKindOfClass:[NSString class]] || !_identifier.length) {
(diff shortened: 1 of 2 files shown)


tburgin (dismissed):

Source/common/SNTRule.m
```diff
@@ -58,7 +58,8 @@ - (instancetype)initWithDictionary:(NSDictionary *)dict {
   self = [super init];
   if (self) {
     _identifier = dict[kRuleIdentifier];
-    if (![_identifier isKindOfClass:[NSString class]] || !_identifier.length) _identifier = dict[kRuleSHA256];
+    if (![_identifier isKindOfClass:[NSString class]] || !_identifier.length)
```
tburgin, line 61: Mind adding {}

tburgin (approved):

PR #844, opened 2022-07-08
[DRAFT] ES/Logging Components Redesign

mlw:
This is a work-in-progress, being released in draft form in order to solicit early feedback.

## Building / Running
Build using the following command:
```
bazel build --apple_generate_dsym -c opt --verbose_failures --sandbox_debug //Source/santad:com.google.santad.sysext
```

You can then run the binary directly via:
```
sudo /path/to/compiled/binary [--quick [timeout_seconds]]
```

Using the optional `--quick` flag will cause the program to exit after 5 seconds. An optional number of seconds can also be passed.

Currently, logged events are echoed to the console via the `LOGE` macro. (This is temporary for during development and won't stay around.)

#### `EndpointSecurity`  Interfaces/Classes/Wrappers
* `Source/santad/EventProviders/EndpointSecurity`
	* `EndpointSecurityAPI` - Wraps the `EndpointSecurity` framework functions. Bridges between the wrapped types below and the ES types.
	* `Client` - This object wraps an `es_client_t`
	* `Message` - This object wraps an `es_message_t`
	* `EnrichedTypes` - Defines all the various per-event-type enriched variants for a `Message`
	* `Enricher` - Takes a `Message` and produces an appropriate `EnrichedMessage`

#### `EndpointSecurity` Clients
* `Source/santad/EventProviders`
	* `SNTEndpointSecurityClientBase` - Protocol declaring the base class interfaces available to ES clients
	* `SNTEndpointSecurityClient` - The base class for ES clients
	* `SNTEndpointSecurityAuthorizer` - `AUTH EXEC` control. Also does caching.
	* `SNTEndpointSecurityRecorder` - Subscribes to events that will get enriched and logged
	* `SNTEndpointSecurityDeviceManager` - `MOUNT`/`REMOUNT`/`UNMOUNT` actions
	* `SNTEndpointSecurityTamperResistance` - Provides functionality to help protect Santa and its resources

#### Resources
* `AuthResultCache` - Wrapper for the "root" and "nonroot" caches. Provided to the various clients that need to interact with the caches
* `SNTDecisionCache` - Wrapped for what was previously in the `SNTEventLog`. It stores `SNTCachedDecisions`.

#### Logging
* `Source/santad/LogsEndpointSecurity`
	* `Logger` - Provides the primary logging interfaces. Composed of a serializer and a writer.
	* `Serializers/`
		* `Serializer` - Abstract base class for all serializers
		* `BasicString` - Functionality previously provided by `SNTSyslogEventLog`
[...]

PR #850, opened 2022-07-14
santad: Log team ID in execution logs, where available

### LINKED ISSUE #776: Log Team IDs
After #634 we need to ensure that Team IDs are logged in all the right places.

## Steps to Complete This

- [ ] Update `SNTSyslogEventLog` and `SNTProtoBufEventLog` classes to output the teamID from the `es_message_t` included in the `santa_message_t`.

russellhancox:
Fixes #776

### DIFF
--- a/Source/common/santa.proto
+++ b/Source/common/santa.proto
@@ -79,6 +79,7 @@ message Execution {
   optional string original_path = 11;
   repeated string args = 12;
   optional string machine_id = 13;
+  optional string team_id = 14;
 }
 
 message DiskAppeared {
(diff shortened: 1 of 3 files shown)


russellhancox:
I should note that this is re-using the team ID field from the cached decision, which comes from MOLCodesignChecker rather than the one provided directly by ES. The reason for this is that it matches what we're using for decisions and using the ES version would require adding more fields to `santa_message_t`, which is going to be updated (or removed) in #844; moving over to using the ES-provided team ID will be part of that PR.

tburgin (approved):

pmarkowsky (approved):
LGTM.

PR #853, opened 2022-07-15
Ensure KVO works for USB config options

pmarkowsky:
This PR adds methods to SNTConfigurator to ensure KVO works for those options through out Santa.

### DIFF
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -475,7 +475,6 @@ static NSString *const kSyncCleanRequired = @"SyncCleanRequired";
   return [self syncAndConfigStateSet];
 }
 
-
 #pragma mark Public Interface
 
 - (SNTClientMode)clientMode {
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -475,6 +475,10 @@ static NSString *const kSyncCleanRequired = @"SyncCleanRequired";
   return [self syncAndConfigStateSet];
 }
 
++ (NSSet *)keyPathsForValuesAffectingUSBBlockMessage {
+  return [self syncAndConfigStateSet];
+}
+
 #pragma mark Public Interface
 
 - (SNTClientMode)clientMode {
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -468,7 +468,7 @@ static NSString *const kSyncCleanRequired = @"SyncCleanRequired";
 }
 
 + (NSSet *)keyPathsForValuesAffectingRemountUSBMode {
-  return [self syncAndConfigStateSet];
+  return [self configStateSet];
 }
 
 + (NSSet *)keyPathsForValuesAffectingRemountUSBBlockMessage {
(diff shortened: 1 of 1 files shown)


Source/common/SNTConfigurator.m
```diff
@@ -459,6 +459,23 @@ + (NSSet *)keyPathsForValuesAffectingEnableBadSignatureProtection {
...
 
++ (NSSet *)keyPathsForValuesAffectingBlockUSBMount {
+  return [self syncAndConfigStateSet];
+}
+
++ (NSSet *)keyPathsForValuesAffectingBannedUSBBlockMessage {
+  return [self syncAndConfigStateSet];
+}
+
++ (NSSet *)keyPathsForValuesAffectingRemountUSBBlockMode {
```
russellhancox, line 470: This should be `keyPathsForValuesAffectingRemountUSBMode`

Source/common/SNTConfigurator.m
```diff
@@ -459,6 +459,23 @@ + (NSSet *)keyPathsForValuesAffectingEnableBadSignatureProtection {
...
+  return [self syncAndConfigStateSet];
+}
+
++ (NSSet *)keyPathsForValuesAffectingRemountUSBBlockMessage {
+  return [self syncAndConfigStateSet];
+}
+
+
```
russellhancox, line 478: I also see a property `usbBlockMessage`, we should probably have a `keyPathsForValuesAffectingUsbBlockMessage` too

Source/common/SNTConfigurator.m
tburgin, line 470: RemountUSBMode ?

Source/common/SNTConfigurator.m
```diff
@@ -459,6 +459,23 @@ + (NSSet *)keyPathsForValuesAffectingEnableBadSignatureProtection {
   return [self configStateSet];
 }
 
++ (NSSet *)keyPathsForValuesAffectingBlockUSBMount {
+  return [self syncAndConfigStateSet];
+}
+
++ (NSSet *)keyPathsForValuesAffectingBannedUSBBlockMessage {
+  return [self syncAndConfigStateSet];
```
russellhancox, line 467: This one only comes from the config

Source/common/SNTConfigurator.m
```diff
@@ -459,6 +459,23 @@ + (NSSet *)keyPathsForValuesAffectingEnableBadSignatureProtection {
...
+}
+
++ (NSSet *)keyPathsForValuesAffectingRemountUSBBlockMode {
+  return [self syncAndConfigStateSet];
```
russellhancox, line 471: This one only comes from the config

russellhancox (changes requested):

Source/common/SNTConfigurator.m
pmarkowsky: Fixed.

pmarkowsky: Fixed.

Source/common/SNTConfigurator.m
russellhancox: It needs to be "AffectingUsbBlockMessage" - the property is `usbBlockMessage`

russellhancox (approved):

PR #858, opened 2022-07-18
Sync state plist | only allow santad read+write permissions

bfreezy:
Nit change here.

An insider or attacker can read the /var/db/santa/sync-state.plist file to determine allowlisted (or denylisted) locations that bypass Santa restrictions.

Example snippet:
```
❯ sudo cat /var/db/santa/sync-state.plist

<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>AllowedPathRegex</key>
	<string>^/Users/[\w-]+/(?:workspace|go/bin|go/src|/opt/homebrew/|/usr/local/(?:Caskroom|Cellar|Homebrew|bin)/.*</string>
	<key>BlockedPathRegex</key>
	<string>^/var/.*/AppTranslocation/.*</string>
</dict>
</plist>
```

Access to this file should require sudo to write/read it.


russellhancox:
Thanks!

PR #859, opened 2022-07-20
Project: Generate SNTConfigurator to reduce boilerplate

russellhancox:
Looking for feedback on the general approach - I'm not convinced myself whether this is a good idea or not.

If we agree this is something we should do:

1. The generator code is a bit messy and could do with a some tidying up.
2. It would be nice to extend to add automated documentation of keys, by converting the existing comments into a structured format. We could use decorators for this, e.g.:

```python
@description("""The operating mode. Defaults to MONITOR.""")
c.RegisterPropertyCustom("ClientMode", SNTClientMode, ...)
```


russellhancox:
We discussed this and decided not to move forward with it; the downsides (in particular the fact that IDEs have no visibility into generated properties) outweigh the benefits. We'll explore other options (macros, checklists for new properties).

PR #870, opened 2022-08-11
Santa: Post distributed notification when showing block UI

### LINKED ISSUE #869: Listening to Events
How do I subscribe to events generated by Santa. 

For example, I have an app that would like to listen in when Santa blocks a binary and would like to trigger additional flow. I assume the Santa Notification is not customizable; would like to provide the user with additional info and next steps when a binary is blocked.

Is there a better way to do it that does not involve the SyncServer and all done locally?

russellhancox:
This allows users to supplement or replace the Santa UI, as well as trigger post-block tools if needed.

Fixes #869

### DIFF
--- a/Source/santa/SNTNotificationManager.m
+++ b/Source/santa/SNTNotificationManager.m
@@ -14,6 +14,7 @@
 
 #import "Source/santa/SNTNotificationManager.h"
 
+#import <MOLCertificate/MOLCertificate.h>
 #import <MOLXPCConnection/MOLXPCConnection.h>
 #import <UserNotifications/UserNotifications.h>
 
@@ -23,6 +24,7 @@
 #import "Source/common/SNTLogging.h"
 #import "Source/common/SNTStoredEvent.h"
 #import "Source/common/SNTStrengthify.h"
+#import "Source/common/SNTSyncConstants.h"
 #import "Source/common/SNTXPCControlInterface.h"
 #import "Source/santa/SNTMessageWindowController.h"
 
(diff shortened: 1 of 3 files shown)


Source/santa/SNTNotificationManager.m
```diff
@@ -112,12 +114,49 @@ - (void)queueMessage:(SNTMessageWindowController *)pendingMsg {
 
   pendingMsg.delegate = self;
   [self.pendingNotifications addObject:pendingMsg];
+  [self postDistributedNotification:pendingMsg];
```
pmarkowsky, line 117: Should whether or not we send notifications be configurable?

Source/santa/SNTNotificationManager.m
```diff
@@ -112,12 +114,49 @@ - (void)queueMessage:(SNTMessageWindowController *)pendingMsg {
...
 
   if (!self.currentWindowController) {
     [self showQueuedWindow];
   }
 }
 
+- (void)postDistributedNotification:(SNTMessageWindowController *)pendingMsg {
```
pmarkowsky, line 124: nit: Add a comment here describing what a distributed notification is.

pmarkowsky (changes requested):
Can we add a test for this?

Source/santa/SNTNotificationManager.m
russellhancox: Done

Source/santa/SNTNotificationManager.m
russellhancox: I don't think so, the notifications are relatively cheap (esp. in the context of only sending them for de-duped blocked executions) and there's nothing sensitive in the notifications that can't easily be gleaned from other sources.

mlw (approved):

PR #872, opened 2022-08-11
GUI: For App Store published apps, include team ID.

### LINKED ISSUE #758: For AppStore Apps notifications show only Apple Signed Publisher instead of TeamID
For apps downloaded from the web the publisher field shows the 10 char ID for the team while for app store apps it its says Apple signed publisher. I have to run fileinfo to get the details

russellhancox:
With this change, the publisher field for an App Store published app will be `App Store (Team
ID: K36BKF7T3D` instead of `Apple Mac OS Application Signing`.

Fixes #758


pmarkowsky (approved):
LGTM.

PR #879, opened 2022-08-16
Fix up endTimestamp to be Monarch compliant

pmarkowsky:
This PR fixes an issue with our Monarch JSON that causes some metrics to be wonky.

Specifically our `endTimestamp` field needs to be the time of the export not the `last_updated` time.

### DIFF
--- a/Source/santametricservice/Formats/SNTMetricMonarchJSONFormat.m
+++ b/Source/santametricservice/Formats/SNTMetricMonarchJSONFormat.m
@@ -113,7 +113,7 @@ const NSString *kKey = @"key";
       }
 
       monarchDataEntry[kStartTimestamp] = [self->_dateFormatter stringFromDate:entry[@"created"]];
-      // monarch wants all the end timestamp to be updated, even if the value does not change.
+      // Monarch wants all the end timestamp to be updated, even if the value does not change.
       monarchDataEntry[kEndTimestamp] = [self->_dateFormatter stringFromDate:endTimestamp];
 
       if (!metric[@"type"]) {
--- a/Source/santametricservice/Formats/SNTMetricMonarchJSONFormat.m
+++ b/Source/santametricservice/Formats/SNTMetricMonarchJSONFormat.m
@@ -100,7 +100,8 @@ const NSString *kKey = @"key";
   }
 }
 
-- (NSArray<NSDictionary *> *)encodeDataForMetric:(NSDictionary *)metric {
+- (NSArray<NSDictionary *> *)encodeDataForMetric:(NSDictionary *)metric
+                                withEndTimestamp:(NSDate *)endTimestamp {
   NSMutableArray<NSDictionary *> *monarchMetricData = [[NSMutableArray alloc] init];
 
   for (NSString *fieldName in metric[@"fields"]) {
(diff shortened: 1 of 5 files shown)


Source/santametricservice/Formats/SNTMetricMonarchJSONFormat.m
```diff
@@ -160,7 +161,9 @@ - (void)encodeValueAndStreamKindFor:(NSString *)metricName
  * formatMetric translates the SNTMetricSet metric entries into those consumable by Monarch.
  **/
 
-- (NSDictionary *)formatMetric:(NSString *)name withMetric:(NSDictionary *)metric {
+- (NSDictionary *)formatMetric:(NSString *)name
```
russellhancox, line 164: The name of this method is confusing because of the differing uses of metric, perhaps `formatMetric:withValue:andEndtimestamp:` or `formatMetricNamed:withMetric:endTimestamp:`?

Source/santametricservice/Formats/SNTMetricMonarchJSONFormatTest.m
```diff
@@ -6,6 +6,17 @@
 @interface SNTMetricMonarchJSONFormatTest : XCTestCase
 @end
 
+// Stub out NSDate's date method
```
russellhancox, line 9: It would be preferable to use OCMock for this, it's shorter and handles a lot of edge cases.

russellhancox (changes requested):

mlw:
How I'm reading this - it appears like this will make the `last_updated` values reported by `SNTCommandMetrics` not match what is sent to the metrics server - is that desireable?

pmarkowsky:
@mlw Short answer is yes. Monarch requires the time series to be reported constantly so as long as we have the value we have to record a new `endTimestamp`. 

If we want `last_updated` to be the last time the metric changed then we need to do this differently if we just want everything to be the last time we exported then we can simplify the `SNTMetricSet` to just always add the timestamp on export.

Source/santametricservice/Formats/SNTMetricMonarchJSONFormatTest.m
pmarkowsky: Fair enough. We'd used this trick elsewhere. I'll move it to OCMock.

Source/santametricservice/Formats/SNTMetricMonarchJSONFormat.m
russellhancox: It wasn't the end timestamp argument I was concerned with, this still has `formatMetric:withMetric:`, which doesn't read well. I recognize that wasn't added in this change, but changing the signature highlighted it.

Source/santametricservice/Formats/SNTMetricMonarchJSONFormatTest.m
```diff
@@ -9,6 +11,13 @@ @interface SNTMetricMonarchJSONFormatTest : XCTestCase
 @implementation SNTMetricMonarchJSONFormatTest
 
 - (void)testMetricsConversionToJSON {
+  id classMock = OCMClassMock([NSDate class]);
+  NSDateFormatter *testDateFormatter = NSDateFormatter.new;
+  [testDateFormatter setDateFormat:@"yyyy-MM-dd HH:mm:ssZZZ"];
+  NSDate *mockedEndTimestamp = [testDateFormatter dateFromString:@"2021-09-16 21:08:10+0000"];
+
+  OCMStub([classMock date]).andReturn(mockedEndTimestamp);
+
```
russellhancox, line 20: You could simplify this to:

```objc
id classMock = OCClassMock([NSDate class]);
OCMStub([classMock date]).andReturn([NSDate dateWithTimeIntervalSince1970:1631826490]); // 2021-09-16 21:08:10Z
```

pmarkowsky: Changed.

russellhancox (approved):

PR #890, opened 2022-09-21
Fix USB config sync

np5:
- get `BlockUSBMount` and `RemountUSBMode` from the `configState` and from the `syncState`.
- disable `BlockUSBMount` if set to `false` in the preflight response.
- only set `BlockUSBMount` in `syncState` if present in the preflight response.
- Fix `keyPathsForValuesAffecting*` for `RemountUSBMode`, `RemountUSBBlockMessage` and `USBBlockMessage`.
- Add `BlockUSBMount` and `RemountUSBMode` to the docs.
- Fix `block_usb_mount` docs.


russellhancox (changes requested):
The changes under santasyncservice/ should be reverted - the mode is retrieved during preflight but it shouldn't be applied until postflight time as the majority of the other keys do. The changes in SNTConfigurator should be enough to fix the bug, I think

np5:
I think this change is necessary if we want to be able to unblock the USB device mounts and if we want to not update the SNTConfigurator syncState if the server is not sending the key.

liamn:
Hey folks 👋 

We're looking at rolling this out internally and appear to have hit the same issue - is there an update on this, or a workaround? The Santa documentation seems to suggest that it'll work with a sync server. Happy to help get this over the line if you need!

pmarkowsky:
@liamn the main issue with this PR is that the settings are being changed in the `Preflight` stage rather than at the end in the `Postflight` stage.  This breaks how the sync protocol is supposed to work. 

As @russellhancox calls out reverting the changes in the syncservice should enough to fix this. If the PR can be changed to do that I think it'll good to merge.

arubdesu:
If @np5 doesn't chime in soon I'll remind him that this is pending his response.

np5:
Sorry for the delay. I made the change in the preflight  because it was easier for me to manage the true/false/missing values in the server response. I think that "missing" or "absent" is important, because if a sync server is not yet implementing this, you do not want to overwrite the value set in the config profile. I am not really proficient with objective C, but maybe by setting the value as a pointer to a boolean in the preflight, it would be possible to maintain the "absent" state separate from "false" and only apply the change in the postflight if the value was send by the server?

russellhancox (approved):

russellhancox:
> I am not really proficient with objective C, but maybe by setting the value as a pointer to a boolean NSNumber in the preflight, it would be possible to maintain the "absent" state separate from "false" and only apply the change in the postflight if the value was sent by the server?

Yes, this will work and what you have now meets our general constraint of receiving new configuration during preflight but applying it during postflight so that syncing acts like a kind of transaction.

Thanks!

liamn:
Hey folks 👋 

I hate to be this person, but do we have any timeframe on a release that will include this fix? I've noticed that things are slightly delayed based on the previous release cadence 🙏

PR #907, opened 2022-10-18
Machine id proto

mlw:
Adds the MachineID field to all basic string serialized messages. And moves the field to the top level SantaMessage message type in the proto schema.

### DIFF
--- a/Source/common/santa.proto
+++ b/Source/common/santa.proto
@@ -251,9 +251,6 @@ message Execution {
   // The original path on disk of the target executable
   // Applies when executables are translocated
   optional string original_path = 15;
-
-  // Machine ID of the host emitting this log
-  optional string machine_id = 16;
 }
 
 // Information about a fork event
@@ -435,11 +432,15 @@ message SantaMessage {
   // Can allow client to relate events across time
   optional string uuid = 1;
 
+  // Machine ID of the host emitting this log
+  // Only valid when EnableMachineIDDecoration configuration option is set
+  optional string machine_id = 2;
+
   // Timestamp when the event occurred
-  optional google.protobuf.Timestamp event_time = 2;
+  optional google.protobuf.Timestamp event_time = 3;
 
   // Timestamp when Santa finished processing the event
-  optional google.protobuf.Timestamp processed_time = 3;
+  optional google.protobuf.Timestamp processed_time = 4;
 
   // Event type being described by this message
   oneof event {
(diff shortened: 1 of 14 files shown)


Source/santad/Logs/EndpointSecurity/Serializers/Protobuf.mm
```diff
@@ -17,7 +17,7 @@
 #include <EndpointSecurity/EndpointSecurity.h>
 #include <Kernel/kern/cs_blobs.h>
 #include <bsm/libbsm.h>
-#include <google/protobuf/arena.h>
+// #include <google/protobuf/arena.h>
```
kallsyms, line 20: nit: uncomment?

mlw: Whoops, moved to header. Fixed. (also fixed an unrelated header that had the same problem - `SNTNotificationManagerTest.m`)

Source/common/santa.proto
```diff
@@ -435,11 +432,15 @@ message SantaMessage {
   // Can allow client to relate events across time
   optional string uuid = 1;
 
+  // Machine ID of the host emitting this log
+  // Only valid when EnableMachineIDDecoration configuration option is set
+  optional string machine_id = 2;
+
   // Timestamp when the event occurred
-  optional google.protobuf.Timestamp event_time = 2;
+  optional google.protobuf.Timestamp event_time = 3;
```
russellhancox, line 440: I assume this proto isn't currently being used by anything?

Source/gui/SNTNotificationManagerTest.m
```diff
@@ -12,8 +12,6 @@
 ///    See the License for the specific language governing permissions and
 ///    limitations under the License.
 
-// #import <MOLCertificate/MOLCertificate.h>
-// #import <MOLCodesignChecker/MOLCodesignChecker.h>
```
russellhancox, line 16: Remove

russellhancox (approved):

pmarkowsky (approved):
LGTM.

Source/common/santa.proto
mlw: Correct.

Source/gui/SNTNotificationManagerTest.m
mlw: Already did :)

PR #908, opened 2022-10-19
Spool writer

mlw:
* Adds a new spool writer to write serialized bytes to the fsspool log batch writer.
* Adds a new santactl command `printlog` to dump log file contents.


Source/santad/Logs/EndpointSecurity/Writers/Spool.mm
```diff
@@ -0,0 +1,102 @@
...
+    : q_(q),
+      timer_source_(timer_source),
+      spool_writer_(base_dir, max_spool_disk_size),
+      log_batch_writer_(&spool_writer_, max_spool_batch_size) {
+  type_url_ = kTypeGoogleApisComPrefix + ::santa::pb::v1::SantaMessage::descriptor()->full_name();
+}
+
+Spool::~Spool() {
+  // Note: `log_batch_writer_` is automatically flushed when destroyed
+  if (flush_task_started_) {
```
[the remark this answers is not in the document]
mlw: Sorta... I have a bug here, but it's not about cancellation.

When a `dispatch_source_t` is created, it has a retain count of 0. When the `dispatch_source_t` is destroyed it will be released. If the object wasn't first `dispatch_resumed` to increment the retain count, then the program will crash.

I did this sorta correctly in the `Metrics` wrapper: https://github.com/google/santa/blob/main/Source/santad/Metrics.mm#L71

However there i should also have done a `dispatch_cancel` prior to `dispatch_resume` to ensure the timer is never fired.

The docs say all this in a roundabout way (IMO...) `dispatch_source_cancel(3)`:
```
If a source is canceled before the first time it is resumed, its event handler will never be called.  (In this case, note that the source must be resumed before it can be released.)
```

So the fix here is to make it look more like the `Metrics` class, but in both cases make sure to `cancel` and then `resume`.

Source/santactl/Commands/SNTCommandPrintLog.mm
```diff
@@ -0,0 +1,129 @@
...
+         @"    ]\n"
+         @"  ]";
+}
+
+- (void)runWithArguments:(NSArray *)arguments {
+  JsonPrintOptions options;
+  options.always_print_enums_as_ints = false;
+  options.always_print_primitive_fields = true;
+  options.preserve_proto_field_names = true;
+  options.add_whitespace = true;
```
pmarkowsky, line 72: nit: We should make these flags.

Source/santad/Logs/EndpointSecurity/Writers/Spool.mm
```diff
@@ -0,0 +1,106 @@
...
+
+  dispatch_resume(timer_source_);
+  flush_task_started_ = true;
+}
+
+bool Spool::Flush() {
+  return log_batch_writer_.Flush().ok();
+}
+
+void Spool::Write(std::vector<uint8_t> &&bytes) {
```
pmarkowsky, line 85: It'd be nice if we could have the a variant of `Spool::write` that was able to call a callback e.g. on line 102. 

This would  allow us to avoid sleeps in our code and could be used to update metrics.

Source/santad/Logs/EndpointSecurity/Writers/SpoolTest.mm
```diff
@@ -0,0 +1,135 @@
...
+    std::make_shared<SpoolPeer>(self.q, self.timer, [self.baseDir UTF8String], 10240, 1024);
+
+  // Set a custom timer interval for this test
+  dispatch_source_set_timer(self.timer, dispatch_time(DISPATCH_TIME_NOW, 0),
+                            NSEC_PER_MSEC * periodicFlushMS, 0);
+
+  spool->Write(std::vector<uint8_t>(writeSize, 'A'));
+
+  // Sleep for 1.25 seconds
+  usleep(waitTime);
```
pmarkowsky, line 105: Sleeps in test code always make me uneasy. I think we can get around this using the callback we talked about in chat.

pmarkowsky (changes requested):
I think if we can get rid of the sleeps in tests this is good to go.

Source/santactl/Commands/SNTCommandPrintLog.mm
mlw: Spoke offline. Like the idea. Going to punt for now until proto out of beta and full options are better scoped.

Source/santad/Logs/EndpointSecurity/Writers/Spool.mm
pmarkowsky: I see you added this to the constructor great.

pmarkowsky (approved):
LGTM.

PR #909, opened 2022-10-21
Proto minimization

mlw:
Create new `ProcessInfoLight` and `FileInfoLight` message variants to reduce on-disk and on-wire size while minimizing the amount of actionable data sacrificed.

### DIFF
--- a/Source/common/santa.proto
+++ b/Source/common/santa.proto
@@ -76,7 +76,7 @@ message Hash {
 }
 
 // File information
-message File {
+message FileInfo {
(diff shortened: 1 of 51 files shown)


russellhancox (approved):

pmarkowsky (approved):
LGTM.

The proto still isn't being used, so it's ok to rename it. In the future we'll need to be more careful. 

Also since we got rid of the UUID, we might want to look into ULIDS next time if we bring back a unique identifier.

PR #915, opened 2022-10-26
USB: usbBlockMessage is not being used.

videlanicolas:
Remove unused `usbBlockMessage` attribute from `SNTConfigurator`.


videlanicolas (commented):
I *think* this property is not being used, a search for it in the code shows no results other than this header file: https://github.com/google/santa/search?q=usbBlockMessage&type=code

pmarkowsky (approved):
LGTM.

PR #916, opened 2022-10-26
Fix: duplicates bug in SNTMetricSet when using multiple fields

pmarkowsky:
Fix a bug in `SNTMetricSet` that was producing duplicates of a metric cell when the metric had multiple fields.


kallsyms (approved):
Seems correct to me

russellhancox (approved):

pmarkowsky:
Closed this as I think there are issues with fields being correctly exported. Will fix and reopen.

PR #918, opened 2022-10-27
Event metrics

mlw:
Add metrics for event counts and event processing times.

Current design passes a block to ES client message handlers to be called when they're done processing and event.

Alternative considered was to have the ES client message handlers return a disposition (instead of calling a block) and leave the work to the base class. This could work but would add some complexity for clients that processed messages async. Given there are only a few clients, it wouldn't be a huge burden to go this route in the future if it provided advantages.

### DIFF
--- a/Source/santad/Metrics.mm
+++ b/Source/santad/Metrics.mm
@@ -87,7 +87,7 @@ const NSString *EventDispositionToString(EventDisposition d) {
     case EventDisposition::kDropped: return kEventDispositionDropped;
     case EventDisposition::kProcessed: return kEventDispositionProcessed;
     default:
-      [NSException raise:@"Invalid processor" format:@"Unknown processor value: %d", d];
+      [NSException raise:@"Invalid disposition" format:@"Unknown disposition value: %d", d];
       return nil;
   }
 }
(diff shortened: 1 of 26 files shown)


Source/santad/SantadDeps.mm
```diff
@@ -109,21 +112,29 @@
     exit(EXIT_FAILURE);
   }
 
-  return std::make_unique<SantadDeps>([configurator metricExportInterval], esapi, std::move(logger),
-                                      control_connection, compiler_controller, notifier_queue,
-                                      syncd_queue, exec_controller, prefix_tree);
+  // TODO: `SNTMetricSet sharedInstance` should be injected
```
russellhancox, line 115: This is done? Or am I missing something?

Source/santad/Metrics.mm
```diff
@@ -13,26 +13,105 @@
...
+      return nil;
+  }
+}
+
+const NSString *EventDispositionToString(EventDisposition d) {
+  switch (d) {
+    case EventDisposition::kDropped: return kEventDispositionDropped;
+    case EventDisposition::kProcessed: return kEventDispositionProcessed;
+    default:
+      [NSException raise:@"Invalid processor" format:@"Unknown processor value: %d", d];
```
russellhancox, line 90: s/processor/disposition

russellhancox (changes requested):

Source/santad/SantadDeps.mm
mlw: Oops, old comment. Removed.

Source/santad/Metrics.mm
mlw: Fixed

russellhancox (approved):

PR #919, opened 2022-10-27
Introduce end-to-end testing

kallsyms:
(This PR looks bigger than it is. Most of the subdirectories in `Testing/integration/VM` are a bazel-ified and slightly adapted version of Apple's virtualization.framework sample code. Happy to take comments on it, but it's not the main point of the PR.)

This PR introduces end-to-end testing for Santa. A fresh VM is spun up for each PR (after approval by a repo admin), which:
* Builds and installs a sync server (currently moroz)
* Builds, installs, and authorizes Santa
* Configures santa with a simple provisioning profile pointing it to moroz
* Runs the basic exec integration tests (`Testing/integration/SNTExecTest.m`)
* Installs multiple provisioning profiles with configuration in them, and ensures Santa changes its behavior appropriately (currently testing USB blocking)
* Changes the sync server configuration, forces a sync, and ensures behavior changes appropriately (currently USB blocking again, plus removal of an executable from the blocklist).

This PR also includes all scripts necessary to build new VM images (the scripts in `Testing/integration/VM`).

**Before this PR is merged, there are a few things that need to happen:**
- [ ] The current macOS 12.6 VM image will need to be slightly rebuilt to reference this repository instead of my fork.
- [ ] The host mac mini the VMs run on will need to be registered to this repo.
- [ ] This repo will need to have the `VM_PASSWORD` key added to its runner secrets.


.github/workflows/ci.yml
```diff
@@ -27,9 +27,47 @@ jobs:
...
+      VM_PASSWORD: ${{ secrets.VM_PASSWORD }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install configuration profile
+        run: bazel run //Testing/integration:install_profile -- Testing/integration/configs/default.mobileconfig
+      - name: Add homebrew to PATH
+        run: echo "/opt/homebrew/bin/" >> $GITHUB_PATH
+      - name: Install and start moroz
+        run: |
+          go install github.com/kallsyms/moroz/cmd/moroz@d0bba3f4
```
tburgin, line 55: Are we able to upstream these moroz changes? If not we need to find a better place to host them. The external_patches dir + a bazel build rule for them may do?

https://github.com/google/santa/tree/main/external_patches

Testing/integration/VM/InstallationTool/BUILD
```diff
@@ -0,0 +1,44 @@
...
+    name = "entitlements",
+    srcs = [
+        "InstallationTool.entitlements",
+    ],
+)
+
+macos_command_line_application(
+    name = "InstallationTool",
+    codesignopts = [
+        "--entitlements=Testing/integration/VM/InstallationTool/InstallationTool.entitlements",
```
tburgin, line 37: Can you use the `entitlements` attribute instead? https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-macos.md#macos_application-entitlements

Testing/integration/VM/VMCLI/BUILD
```diff
@@ -0,0 +1,37 @@
...
+    name = "entitlements",
+    srcs = [
+        "VMCLI.entitlements",
+    ],
+)
+
+macos_command_line_application(
+    name = "VMCLI",
+    codesignopts = [
+        "--entitlements=Testing/integration/VM/VMCLI/VMCLI.entitlements",
```
tburgin, line 30: same

Testing/integration/VM/VMGUI/BUILD
```diff
@@ -0,0 +1,47 @@
...
+    srcs = [
+        "VMGUI.entitlements",
+    ],
+)
+
+macos_application(
+    name = "VMGUI",
+    bundle_id = "com.google.santa.e2e.vmgui",
+    codesignopts = [
+        "--entitlements=Testing/integration/VM/VMGUI/VMGUI.entitlements",
```
tburgin, line 39: same

Testing/integration/allow_sysex.scpt
```diff
@@ -0,0 +1,30 @@
+on run argv
```
tburgin, line 1: Do we need this if `systemextensionsctl developer on` is set?

tburgin (commented):
I have not taken a look at the VM code yet. Though I do think it should be reviewed just as we would any other code we commit to this repo.

Testing/integration/VM/InstallationTool/BUILD
kallsyms: We can, however `entitlements` isn't supported for `macos_command_line_application` so all of these would need to use `macos_application` and have the supporting Info.plist added. Doing that still feels cleaner than this "hack" though, so will change.

Testing/integration/allow_sysex.scpt
mlw: It also affects the upgrades - allowing upgrades to/from any version number

Testing/integration/actions/start-vm.py
```diff
@@ -0,0 +1,102 @@
...
+    if blob.updated > datetime.datetime.fromtimestamp(local_ctime, tz=datetime.timezone.utc):
+        print(f"VM {extracted_path} not present or not up to date, downloading...")
+
+        # Remove the old version of the image if present
+        try:
+          shutil.rmtree(extracted_path)
+        except FileNotFoundError:
+          pass
+
+        blob.download_to_filename(tar_path)
```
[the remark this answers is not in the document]
mlw: Looking closer, seems like this is all sorta handled via exceptions. Do we cleanup bits that were already downloaded in the event some downloads fail, or things like the `check_output` calls below thorw?

Testing/integration/VM/setup.sh
```diff
@@ -0,0 +1,57 @@
...
+	(cd /Applications && xip --expand ${SCRIPT_DIR}/Xcode*.xip)
+	sudo xcode-select -s /Applications/Xcode.app
+	sudo xcodebuild -license accept
+	# ... and open it to fixup stuff
+	open /Applications/Xcode.app
+fi
+
+# Install brew
+if ! command -v brew &> /dev/null; then
+	/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
```
mlw, line 20: Might be better to install from a known release instead of `HEAD`

Testing/integration/VM/setup.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# Install rosetta (for test binaries)
+softwareupdate --install-rosetta --agree-to-license
+
+# Install actions runner
+mkdir ~/actions-runner
+pushd ~/actions-runner
+curl -o actions-runner-osx-arm64-2.296.0.tar.gz -L https://github.com/actions/runner/releases/download/v2.296.0/actions-runner-osx-arm64-2.296.0.tar.gz
+tar xzf ./actions-runner-osx-arm64-2.296.0.tar.gz
+./config.sh --url https://github.com/kallsyms/santa
```
mlw, line 39: Is this correct? Why not use the main repo?

.github/workflows/ci.yml
kallsyms: Doesn't seem like these will be able to be upstreamed on any regular basis, so rewrote with a patch.

Testing/integration/actions/start-vm.py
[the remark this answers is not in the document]
kallsyms: Everything raises an exception if it fails (from downloading to verifying checksums), and all of these should be uncaught so the CI job fails. We don't cleanup anything in between, as it'll just be overwritten on the next run if it's an intermittent failure.

Testing/integration/VM/setup.sh
kallsyms: Doesn't look like this installer script actually has any releases, nor does it provide a way to easily pin a specific version of brew to be installed :(

Testing/integration/VM/setup.sh
kallsyms: Went ahead and changed - this was part of the first bulleted thing to do before merging.

Testing/integration/VM/setup.sh
```diff
@@ -0,0 +1,57 @@
...
+	brew install bazelisk go jq
+fi
+
+# Install rosetta (for test binaries)
+softwareupdate --install-rosetta --agree-to-license
+
+# Install actions runner
+mkdir ~/actions-runner
+pushd ~/actions-runner
+curl -o actions-runner-osx-arm64-2.296.0.tar.gz -L https://github.com/actions/runner/releases/download/v2.296.0/actions-runner-osx-arm64-2.296.0.tar.gz
```
pmarkowsky, line 37: We should check the shasum of this.

Testing/integration/allow_sysex.scpt
pmarkowsky, line 1: nit: comment this file. 

Which host should we be running this on etc. Explain that we need this to get rid of the stale pref.

Could we use automator at all for any of this?

Testing/integration/install_profile.scpt
```diff
@@ -0,0 +1,34 @@
+on run argv
```
pmarkowsky, line 1: nit: comment this.

Testing/integration/test_sync_changes.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+set -x
+
+bazel run //Testing/integration:install_profile -- Testing/integration/configs/default.mobileconfig
+
+# Reset moroz to default config
+killall moroz
+/tmp/moroz -configs="$GITHUB_WORKSPACE/Testing/integration/configs/moroz_default/global.toml" -use-tls=false &
```
pmarkowsky, line 8: In the future, we should probably have test with TLS enabled, to make sure we're testing all paths.

Testing/integration/test_sync_changes.sh
```diff
@@ -0,0 +1,47 @@
...
+  exit 1
+fi
+
+if [[ "$(sudo santactl status --json | jq .daemon.block_usb)" != "false" ]]; then
+  echo "USB blocking enabled with minimal config" >&2
+  exit 1
+fi
+
+# Now change moroz to use the changed config, enabling USB blocking and removing the badbinary block rule
+killall moroz
```
pmarkowsky, line 29: Also in the future, we should patch moroz to re-read its config after a SIGHUP.

Testing/integration/test_sync_changes.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
pmarkowsky, line 1: In the future we should replace this with GitHub  workflow steps.

pmarkowsky (commented):
Looks good. Update comments and we can approve.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,41 @@
+name: CI
```
pmarkowsky, line 1: Change the name of the workflow.

pmarkowsky (approved):
LGTM.

PR #920, opened 2022-10-27
Fix: duplicates bug in SNTMetricSet when using multiple fields

pmarkowsky:
Fix a bug in `SNTMetricSet` that was producing duplicates of a metric cell when the metric had multiple fields.

### Other Changes

- This changes the `santactl metric` command to print metrics with multiple fields as key value pairs e.g. `key=value,key2=value2`
- Updates to JSON test files.

### DIFF
--- a/Source/santactl/Commands/SNTCommandMetrics.m
+++ b/Source/santactl/Commands/SNTCommandMetrics.m
@@ -68,14 +68,33 @@ REGISTER_COMMAND_NAME(@"metrics")
 
     for (NSString *fieldName in metric[@"fields"]) {
       for (NSDictionary *field in metric[@"fields"][fieldName]) {
-        const char *fieldNameStr = [fieldName cStringUsingEncoding:NSUTF8StringEncoding];
-        const char *fieldValueStr = [field[@"value"] cStringUsingEncoding:NSUTF8StringEncoding];
+        //const char *fieldNameStr = [fieldName cStringUsingEncoding:NSUTF8StringEncoding];
+        //const char *fieldValueStr = [field[@"value"] cStringUsingEncoding:NSUTF8StringEncoding];
         const char *createdStr = [field[@"created"] UTF8String];
         const char *lastUpdatedStr = [field[@"last_updated"] UTF8String];
         const char *data = [[NSString stringWithFormat:@"%@", field[@"data"]] UTF8String];
 
-        if (strlen(fieldNameStr) > 0) {
-          printf("  %-25s | %s=%s\n", "Field", fieldNameStr, fieldValueStr);
+        NSArray<NSString *> *fields = [fieldName componentsSeparatedByString:@","];
+        NSArray<NSString *> *fieldValues = [field[@"value"] componentsSeparatedByString:@","];
+
+        if (fields.count != fieldValues.count) {
+          //TODO Log Error
+          continue;
+        }
+
+        NSString *fieldDisplayString = @"";
+
+        if (fields.count >= 1 && ![@"" isEqualToString: fields[0]]) {
+        for (int i = 0; i < fields.count; i++) {
+          fieldDisplayString = [fieldDisplayString stringByAppendingString:[NSString stringWithFormat:@"%@=%@",fields[i], fieldValues[i]]];
+          if (i < fields.count - 1) {
+            fieldDisplayString = [fieldDisplayString stringByAppendingString:@","];
+          }
+        }
+        }
+
+        if (![fieldDisplayString isEqualToString: @""]) {
+          printf("  %-25s | %s\n", "Field", [fieldDisplayString UTF8String]);
         }
 
         printf("  %-25s | %s\n", "Created", createdStr);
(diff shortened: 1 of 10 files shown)


Source/santametricservice/Formats/SNTMetricMonarchJSONFormat.m
```diff
@@ -54,7 +55,7 @@ - (instancetype)init {
   return self;
 }
 
-- (void)encodeValueAndStreamKindFor:(NSString *)metricName
+- (void)encodeValueTypeAndStreamKindFor:(NSString *)metricName
```
russellhancox, line 58: Line up subsequent colons

Source/santactl/Commands/SNTCommandMetrics.m
```diff
@@ -68,14 +68,33 @@ - (void)prettyPrintMetricValues:(NSDictionary *)metrics {
 
     for (NSString *fieldName in metric[@"fields"]) {
       for (NSDictionary *field in metric[@"fields"][fieldName]) {
-        const char *fieldNameStr = [fieldName cStringUsingEncoding:NSUTF8StringEncoding];
-        const char *fieldValueStr = [field[@"value"] cStringUsingEncoding:NSUTF8StringEncoding];
+        //const char *fieldNameStr = [fieldName cStringUsingEncoding:NSUTF8StringEncoding];
+        //const char *fieldValueStr = [field[@"value"] cStringUsingEncoding:NSUTF8StringEncoding];
```
russellhancox, line 72: Remove

Source/santactl/Commands/SNTCommandMetrics.m
```diff
@@ -68,14 +68,33 @@ - (void)prettyPrintMetricValues:(NSDictionary *)metrics {
...
         const char *lastUpdatedStr = [field[@"last_updated"] UTF8String];
         const char *data = [[NSString stringWithFormat:@"%@", field[@"data"]] UTF8String];
 
-        if (strlen(fieldNameStr) > 0) {
-          printf("  %-25s | %s=%s\n", "Field", fieldNameStr, fieldValueStr);
+        NSArray<NSString *> *fields = [fieldName componentsSeparatedByString:@","];
+        NSArray<NSString *> *fieldValues = [field[@"value"] componentsSeparatedByString:@","];
+
+        if (fields.count != fieldValues.count) {
+          //TODO Log Error
```
russellhancox, line 81: Why todo? Can LOGE not be used?

Source/santactl/Commands/SNTCommandMetrics.m
```diff
@@ -68,14 +68,33 @@ - (void)prettyPrintMetricValues:(NSDictionary *)metrics {
...
+          continue;
+        }
+
+        NSString *fieldDisplayString = @"";
+
+        if (fields.count >= 1 && ![@"" isEqualToString: fields[0]]) {
```
russellhancox, line 87: I think the second condition would be clearer as `fields[0].length`, assuming I'm reading the intention properly (checking that fields[0] is not an empty string)

Source/santactl/Commands/SNTCommandMetrics.m
```diff
@@ -68,14 +68,33 @@ - (void)prettyPrintMetricValues:(NSDictionary *)metrics {
...
+        for (int i = 0; i < fields.count; i++) {
+          fieldDisplayString = [fieldDisplayString stringByAppendingString:[NSString stringWithFormat:@"%@=%@",fields[i], fieldValues[i]]];
```
russellhancox, line 89: Add a space after first comma

Source/santametricservice/Formats/SNTMetricMonarchJSONFormat.m
```diff
@@ -109,7 +110,19 @@ - (void)encodeValueAndStreamKindFor:(NSString *)metricName
       NSMutableDictionary *monarchDataEntry = [[NSMutableDictionary alloc] init];
 
       if (![fieldName isEqualToString:@""]) {
-        monarchDataEntry[@"field"] = @[ @{kName : fieldName, kStringValue : entry[@"value"]} ];
+        // We encode multiple fields as a single comma separated string.
+        NSArray<NSString *> *fieldNames = [fieldName componentsSeparatedByString:@","];
+        NSArray<NSString *> *fieldValues = [entry[@"value"] componentsSeparatedByString:@","];
+
+        if (fieldNames.count != fieldValues.count) {
+          LOGE(@"malformed metric data encounterd: %@", fieldName);
```
russellhancox, line 118: typo

russellhancox (changes requested):

Source/santactl/Commands/SNTCommandMetrics.m
mlw: Why only check the first index of `fields` for the empty string - should each index be checked in the loop below? (if not, maybe deserves a comment)

Source/santactl/Commands/SNTCommandMetrics.m
pmarkowsky: This is a user facing so probably best to print to stderr instead no? Otherwise we need to add logging as a dependency to the command.

mlw: fwiw, `LOGE` will print to stderr for things like `santactl` ran from the terminal

Source/santactl/Commands/SNTCommandMetrics.m
pmarkowsky: In this case its because `@""` in the first field position is treated as a metric without any fields.

mlw (approved):

russellhancox (approved):

PR #923, opened 2022-10-28
Change order that ES clients are enabled

mlw:
Ensure the `Authorizer` client is enabled first so that all 3rd party execs have policy appropriately applied.


pmarkowsky (approved):
LGTM.

PR #929, opened 2022-11-03
Batch SantaMessage messages before wrapping in the Any message

mlw:
The Any message type includes a `type_url` field that contains type information useful for unpacking. However this is duplicated for every message for `repeated` `Any` messages, causing unnecessary overhead.

This change introduces an intermediate type, `SantaMessageBatch` that contains a `repeated SantaMessage` field. Then this one field is wrapped in the `Any` message, amortizing the cost of the `type_url`.


mlw:
Closing this for now. After some internal discussion, we've decided to proceed without batching first and add in later if the need arises.

PR #930, opened 2022-11-03
Update spool to flush on size thresholds instead of batch counts

mlw:
Fixes an issue where the "file size" configuration option was instead treated as "batch count".


russellhancox (approved):

pmarkowsky (approved):
LGTM.

PR #931, opened 2022-11-05
Prefix tree updates

mlw:
* Nodes can now be of type `prefix` or `literal`
* Nodes now store values
    * New `Unit` type added for trees that don't need this feature
* Added `LookupLongestMatchingPrefix` API to return the value for the longest matching prefix (or exact literal match)
* HasPrefix largely works as before, with the caveat that literal nodes only match exactly
* Simplified locking, now use only a single lock
    * Now use `absl::Mutex` type
* Tree now stores all added prefixes (previously, only the shortest prefix was stored. Adding a shorter prefix pruned the subtree that wasn't previously necessary)


Source/common/Unit.h
```diff
@@ -0,0 +1,24 @@
...
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+#ifndef SANTA__COMMON__UNIT_H
+#define SANTA__COMMON__UNIT_H
+
+namespace santa::common {
+
+struct Unit {};
```
kallsyms, line 20: May just be me but I think a name like `Empty` would work better here... Or maybe spell out `UnitType`?

Source/common/PrefixTree.h
```diff
@@ -0,0 +1,297 @@
...
+    return InsertLocked(s, value, NodeType::kPrefix);
+  }
+
+  bool InsertLiteral(const char *s, ValueT value) {
+    absl::MutexLock lock(&lock_);
+    return InsertLocked(s, value, NodeType::kLiteral);
+  }
+
+  bool HasPrefix(const char *input) {
+    absl::ReaderMutexLock lock(&lock_);
```
tburgin, line 57: Ah, so much cleaner!

Source/common/PrefixTree.h
```diff
@@ -0,0 +1,297 @@
...
+    return match ? std::make_optional<ValueT>(match->value_) : std::nullopt;
+  }
+
+  ABSL_EXCLUSIVE_LOCKS_REQUIRED(lock_)
+  void PruneLocked(TreeNode *target) {
+    if (!target) {
+      return;
+    }
+
+    // For deep trees, a recursive approach will generate too many stack frames.
```
tburgin, line 203: IIRC this was a limitation of the kernel's stack. Update the comment now that we are in user space only.

tburgin (approved):

pmarkowsky (dismissed):
LGTM.

Source/common/PrefixTree.h
mlw: It could still get pretty deep in user space, depending on the size of the tree. A default max of 1024 may not blow through the stack on some systems (most?), but likely still safer to keep the current non-recursive approach. Updated the comment still, though.

Source/common/Unit.h
mlw: Spoke offline. Going to keep `Unit`

tburgin (approved):

PR #932, opened 2022-11-07
Don't add messages when accumulated bytes exceeds threshold

mlw:


russellhancox (approved):

pmarkowsky (dismissed):
LGTM. 

As per discussion in chat the leniency factor should help consumers deal with bursts of logs.

russellhancox (approved):

PR #938, opened 2022-11-28
Include SD Card Mounting in the USB Block Functionality

liamn:
This PR includes SD Card Mounting in the USB Block functionality. Currently, SD cards are able to be mounted when utilising the USB Storage Block functionality.

I understand this may be intended, so also happy to refactor this to allow the administrator to control the configuration (being weary of bloating the number of configuration parameters) - but, from an end user perspective, my expectation would be that SD Cards are included in any USB Storage block (and that has certainly been the case for ourselves)


pmarkowsky:
@liamn First off thanks for this. Second off we can't take code without the CLA being signed. 

Other than than the CLA this looks fine except we need to update the comment to match the checks.

liamn:
@pmarkowsky Thanks for the heads up and review! I've sorted the CLA and have updated the comment to emphasise that SD Cards are also included in the block.

russellhancox (approved):

PR #943, opened 2022-11-29
Rename santa_vnode_id_t to SantaVnode

mlw:
Also add factory method to create a `SantaVnode` from an `es_file_t`.

### DIFF
--- a/Source/common/SNTCommon.h
+++ b/Source/common/SNTCommon.h
@@ -47,8 +47,8 @@ typedef enum {
 
 // Struct to manage vnode IDs
 typedef struct SantaVnode {
-  uint64_t fsid;
-  uint64_t fileid;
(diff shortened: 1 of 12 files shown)


Source/common/SNTCommon.h
```diff
@@ -45,15 +46,22 @@ typedef enum {
    x == ACTION_RESPOND_ALLOW_COMPILER)
 
 // Struct to manage vnode IDs
-typedef struct santa_vnode_id_t {
+typedef struct SantaVnode {
   uint64_t fsid;
   uint64_t fileid;
```
russellhancox, line 51: WDYT about using this as a chance to chance the type of these to `st_dev` and `st_ino` so we don't need any casting?

russellhancox (approved):

PR #948, opened 2022-12-05
santad: Change workaround for glob header with blocks

russellhancox:


Source/santad/DataLayer/WatchItems.mm
```diff
@@ -172,21 +157,21 @@ bool ConfirmValidWatchItemConfig(const NSDictionary *watch_item_dict) {
                                  PrefixTree<std::shared_ptr<WatchItemPolicy>> &tree,
                                  std::set<std::string> &paths) {
   for (const std::shared_ptr<WatchItemPolicy> &item : watch_items) {
-    glob_t g;
-    int err = glob(item->path.c_str(), 0, nullptr, &g);
+    glob_t *g = (glob_t *)alloca(sizeof(glob_t));
```
mlw, line 160: This should be moved outside of the loop

mlw (changes requested):

Source/santad/DataLayer/WatchItems.mm
russellhancox: Done. I've also added a call to globfree to ensure any space allocated by glob() is freed at the end.

mlw: Good catch on missing `globfree` - but I think it should be moved to inside the loop since we're not using the `GLOB_APPEND` flag

russellhancox: Good call, done

mlw (approved):

PR #951, opened 2022-12-07
FS Access Config Version, Policy decision enums

mlw:
This PR includes some various cleanup and fine tuning from the initial work adding the FS access ES client (https://github.com/google/santa/pull/949). Namely:

1. Add a `Version` requirement to the watch items config
    * This is in preparation for reporting access events to provide better clarity to just a policy name, a version can also be reported.
1. When policy is applied, replace that board `es_auth_result_t` result with a more fine-grained enum type
    * This is also in preparation for reporting to provide better analysis to why a log was emitted
1. When checking for an invalid signature when applying policy, only deny if the `EnableBadSignatureProtection` configuration item was set to `true`


russellhancox (approved):

pmarkowsky (approved):
LGTM.

PR #959, opened 2022-12-08
Dynamically enable/disable FS Access client based on config

mlw:
This change allows the FS Access client to be dynamically enabled/disabled based on whether or not any configured watch items exist. It also adopts watching the changes to the appropriate config key to see if the config path for watch items is modified.


kallsyms (approved):

pmarkowsky (approved):
LGTM.

PR #961, opened 2022-12-09
Use the appropriate variable when asynchronously processing auth messages

mlw:
There isn't an underlying bug here, retain counts are still appropriately set. It just looks weird so cleaning it up.


pmarkowsky (approved):
LGTM.

PR #962, opened 2022-12-12
Enrich file access events, prepare for logging

mlw:
This PR performs enrichment for file access events in preparation for logging. The following changes/additions are included:

1. Update the `LogAccess` interface to receive the necessary information to serialize a file access event
1. Simplified `GetPathTargets` interface
1. Skip truncated paths
1. Added `EnrichOptions` enum that can flag an enrichment as "local only", meaning it will not perform any tasks that could potentially trigger actions both external processes.


Source/santad/EventProviders/EndpointSecurity/Enricher.mm
```diff
@@ -81,26 +81,30 @@
...
 }
 
-std::optional<std::shared_ptr<std::string>> Enricher::UsernameForUID(uid_t uid) {
+std::optional<std::shared_ptr<std::string>> Enricher::UsernameForUID(uid_t uid,
+                                                                     EnrichOptions options) {
   std::optional<std::shared_ptr<std::string>> username = username_cache_.get(uid);
 
   if (username.has_value()) {
     return username;
+  } else if (options == EnrichOptions::kLocalOnly) {
```
kallsyms, line 105: maybe this would be better named `kCacheOnly`? to me local only sounds like it would read out of /etc/passwd or similar (but not cause processes to spawn or network accesses), but this only returns if something had previously set it in our local cache.

mlw: Maybe `kNoExternalLookups`? The key is meant to describe that any enrichment performed must not reach out to external processes.

kallsyms: Ah, ok. In that case I think it's fine to leave as-is - just wasn't sure if it was always going to be "cache only" or may in the future read locally.

kallsyms (approved):

Source/santad/Logs/EndpointSecurity/Logger.h
```diff
@@ -59,7 +59,11 @@ class Logger {
   void LogDiskAppeared(NSDictionary *props);
   void LogDiskDisappeared(NSDictionary *props);
 
-  virtual void LogAccess(const santa::santad::event_providers::endpoint_security::Message &msg);
+  virtual void LogAccess(
```
pmarkowsky, line 62: nit: This should probably be `LogFileAccess`

pmarkowsky (commented):
Fix the path separator issue and we're good.

kallsyms (approved):

PR #971, opened 2022-12-16
Address policy consistency issues

mlw:
This PR addresses two issues:

1. Policies from different configurations could be applied to event types with more than target path (e.g. rename with src/dst) if a config update occurred at the same time as the event being handled.
1. If a config update occurred after a policy was applied but before it was logged, it would be possible to log the new policy version for the event that had the old policy applied.


pmarkowsky (approved):
LGTM.

PR #975, opened 2022-12-20
Adopt new ES APIs to monitor target paths

mlw:
This PR adopts the new ES API to limit received events to only the configured watch paths.

### DIFF
--- a/Source/santad/Santad.mm
+++ b/Source/santad/Santad.mm
@@ -321,8 +321,11 @@ void SantadMain(std::shared_ptr<EndpointSecurityAPI> esapi, std::shared_ptr<Logg
   // execution policy appropriately.
   [authorizer_client enable];
   [tamper_client enable];
-  // Start monitoring any watched items
-  watch_items->BeginPeriodicTask();
+  if (@available(macOS 13.0, *)) {
+    // Start monitoring any watched items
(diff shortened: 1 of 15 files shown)


Source/santad/Santad.mm
```diff
@@ -321,8 +321,11 @@ void SantadMain(std::shared_ptr<EndpointSecurityAPI> esapi, std::shared_ptr<Logg
   // execution policy appropriately.
   [authorizer_client enable];
   [tamper_client enable];
-  // Start monitoring any watched items
-  watch_items->BeginPeriodicTask();
+  if (@available(macOS 13.0, *)) {
```
kallsyms, line 324: nit: I feel like this would be better to check within `BeginPeriodicTask`

kallsyms (approved):

pmarkowsky (dismissed):
LGTM.

Source/santad/Santad.mm
mlw: I'd like to keep it here for now unless there is a strong argument against it. There's nothing about the `WatchItems` object that is macOS 13 specific and if we move the check into that object, it will mean the tests for that method will not work on older OSs.

pmarkowsky (approved):
LGTM

PR #976, opened 2022-12-21
Revitalize Fuzzing

kallsyms:
This PR sets up [rules_fuzzing](https://github.com/bazelbuild/rules_fuzzing) for Santa with a new fuzz test case for the Mach-O parsing code. 1,000,000 iterations are configured to run nightly.


.github/workflows/ci.yml
```diff
@@ -54,3 +54,21 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           path-to-lcov: ./bazel-out/_coverage/_coverage_report.dat
           flag-name: Unit
+
+
+  fuzz:
```
pmarkowsky: Move this to its own workflow YAML.

Fuzzing/install_libclang_fuzzer.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+# Xcode doesn't include the fuzzer runtime, but the one LLVM ships is compatible with Apple clang.
+set -uexo pipefail
```
pmarkowsky, line 3: Nice this elevates the need to maintain a toolchain.

Fuzzing/install_libclang_fuzzer.sh
```diff
@@ -0,0 +1,14 @@
...
+
+CLANG_VERSION=$(clang --version | head -n 1 | cut -d' ' -f 4)
+DST_PATH="/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/${CLANG_VERSION}/lib/darwin/libclang_rt.fuzzer_osx.a"
```
russellhancox, line 6: Use `xcode-select -p` to find the path to the selected Developer dir instead of hardcoding this path

Fuzzing/fuzzing.bzl
```diff
@@ -0,0 +1,20 @@
...
+        deps = deps,
+        **kwargs
+    )
+
+    cc_fuzz_test(
+        name = name,
+        deps = [
+            "%s_lib" % name,
+        ],
+        linkopts = kwargs.get("linkopts", []),
```
russellhancox, line 18: Why not have linkopts as a parameter to the rule (with a default of [])?

Fuzzing/install_libclang_fuzzer.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
russellhancox, line 1: May as well remove `Fuzzing/build.sh` in this PR too.

.github/workflows/fuzz.yml
```diff
@@ -0,0 +1,24 @@
...
+  fuzz:
+    runs-on: self-hosted
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup libfuzzer
+        run: Fuzzing/install_libclang_fuzzer.sh
+      - name: Cleanup any existing fuzz runs
+        run: rm -rf /tmp/fuzzing/artifacts  # N.B. keeps the corpus
+      - name: Fuzz Mach-O Parsing
+        run: bazel run --config=fuzz //Fuzzing:MachOParse_run -- -- -max_len=32768 -runs=1000000
```
russellhancox, line 18: Could you use `Fuzzing:all` or add a test_suite() rule that we can reference so we don't have to remember to update this workflow?

WORKSPACE
```diff
@@ -142,3 +142,20 @@ objc_library(
     remote = "https://github.com/erikdoe/ocmock",
     shallow_since = "1609349457 +0100",
 )
+
+# Fuzzing
+
+git_repository(
+    name = "rules_fuzzing",
+    commit = "b193df79b10dbfb4c623bda23e825e835f12bada",  # PR 213 which fixes macOS
```
pmarkowsky, line 150: nit: change the comment to Commit post PR 213 which fixes macOS

pmarkowsky (approved):
LGTM.

.github/workflows/ci.yml
pmarkowsky: This is done.

.github/workflows/fuzz.yml
pmarkowsky: One advantage of leaving this here as a GitHub action is that we can open up issues automatically for crashes using the GitHub API and reference the fuzzing step that produced it.

kallsyms: Using `bazel query` now to iterate over all fuzz targets. Eliminates having to update the actions config, but will also let us open issues with the specific fuzz target which failed.

russellhancox (approved):

PR #985, opened 2022-12-26
Fix SNTFileInfo Fuzzing

kallsyms:
Fixes a file descriptor leak which was killing fuzzing. Also automatically turns off/destroys the fuzzing VM after it's done instead of waiting for the 15 minute timeout.

### DIFF
--- a/Source/common/SNTFileInfo.m
+++ b/Source/common/SNTFileInfo.m
@@ -642,7 +642,9 @@ extern NSString *const NSURLQuarantinePropertiesKey WEAK_IMPORT_ATTRIBUTE;
 ///
 - (NSData *)safeSubdataWithRange:(NSRange)range {
   @try {
-    if ((range.location + range.length) > self.fileSize) return nil;
+    if (range.location > self.fileSize || range.length > self.fileSize ||
+        (range.location + range.length) > self.fileSize)
+      return nil;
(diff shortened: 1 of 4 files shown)


mlw (dismissed):

Source/common/SNTFileInfo.m
```diff
@@ -642,7 +642,9 @@ - (struct mach_header *)firstMachHeader {
 ///
 - (NSData *)safeSubdataWithRange:(NSRange)range {
   @try {
-    if ((range.location + range.length) > self.fileSize) return nil;
+    if (range.location > self.fileSize || range.length > self.fileSize ||
+        (range.location + range.length) > self.fileSize)
+      return nil;
```
mlw, line 647: Is the goal here to make this safer for large values? Seems there could still be an issue on the upper boundaries - this issue previously existed as well... Maybe should change to something like:

```suggestion
    NSUInteger size;
    if (__builtin_add_overflow(range.location, range.length, &size) || size > self.fileSize) {
      return nil;
    }
```

mlw (changes requested):

Source/common/SNTFileInfo.m
kallsyms: Yeah this arose from size being near UINT64_MAX. Changed.

mlw (approved):

PR #986, opened 2022-12-28
Fix nightly run cron specification

kallsyms:
Just realized these were wrong while checking on fuzzing last night, causing a lot more runs than intended.


pmarkowsky (approved):
LGTM.

mlw (approved):

PR #988, opened 2022-12-29
DRAFT: Current vs. Proposed config for FS monitor client

mlw:
This draft proposes a new config structure to hopefully better encapsulate the goals of a configuration and make it more expandable for the future.


PROPOSED-fsmon-config.plist
```diff
@@ -0,0 +1,64 @@
...
+        <false/>
+        <!-- Expandable to support future options. Potential upcoming ones: -->
+        <!-- Potential upcoming options: -->
+        <!--   * InvertProcessLogic: By default, listed processes are allowed -->
+        <!--     to perform the op. this key could invert that logic such that -->
+        <!--     the listed processes are denied and everything else is allowed. -->
+        <!--   * ProcessLogicAND: By default, a process matching any of the criteria -->
+        <!--     is allowed. This key =would require a process meet all the criteria. -->
+      </dict>
+      <key>Processes</key>
```
mlw, line 42: Another potential name here would be `Exceptions` instead of `Processes`. This would provide more flexibility for attributes we want to add in the future. A potential downside is interoperability with other tools that only operate on process exceptions and would make a common superset config more difficult (can discuss offline).

pmarkowsky: When we're describing exceptions are we `and`-ing or `or`-ing the key value pairs?

mlw: The current default is `or`. Above I commented on a potential future `ProcessLogicAND` key in the `Options` dictionary.

pmarkowsky (commented):
I like the updated proposal best. It seems expressive enough to describe the use cases we have.

PROPOSED-fsmon-config.plist
```diff
@@ -0,0 +1,64 @@
...
+			<key>Paths</key>
+      <array>
+        <!-- Paths can be strings with defaults set (IsPrefix=false) -->
+        <string>/Users/*/path1</string>
+        <!-- Paths can also be dicts with keys to override defautls -->
+        <!-- Currently one attribute, IsPrefix, maybe more in the future -->
+        <dict>
+          <key>Path</key>
+          <string>/Users/*/path2</string>
+          <key>IsPrefix</key>
```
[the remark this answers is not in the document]
mlw: Discussed offline, but putting here for posterity:

There is a subtle feature of the configuration around prefix paths, and paths ending with `*`. Paths ending with the glob pattern `*` will match all items at a glob path at a point in time, and all those matches will be watched. Separately, `IsPrefix` being set would have all of the paths matching the glob be prefix paths, and anything under those paths would also match.

For example, the config:
```
<dict>
    <key>Path</key>
    <string>/Users/FakeUser/D*</string>
    <key>IsPrefix</key>
    <true/>
</dict>
```
When the config is applied (and at every update interval), all paths matching that glob would be watched (e.g., `/Users/FakeUser/Downloads`, and `/Users/FakeUser/Desktop`), and then any access down in a nested directory would then be monitored (e.g. an access of `/Users/FakeUser/Downloads/Unzipped/New/script.sh` would match).

Using the syntactic sugar to get this behavior, the configured path would need to be `/Users/FakeUser/D***`. I'm hesitant to support this because of how awkward the triple `*` looks, as well as potentially confusing users that globstar is supported within a path when it currently isn't (and no current plans to add support).

mlw:
Closing this draft as it was only meant to provide example proposals and a common place to view and gather feedback.

Will be moving forward with the `UPDATED-PROPOSAL` variant.

PR #990, opened 2023-01-03
Fuzz embedded plist reading

kallsyms:
This also includes a guard around MachO command size to ensure the fuzzer doesn't infinite loop. `cmdsize` is checked by the OS during loading, so it would never be an issue in real event processing.

### DIFF
--- a/Fuzzing/common/MachOParse.mm
+++ b/Fuzzing/common/MachOParse.mm
@@ -29,6 +29,8 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     // Mach-O Parsing
     [fi architectures];
     [fi isMissingPageZero];
+
+    [fi infoPlist];
   }
 
   if (num_fds_pre != get_num_fds()) {
--- a/Fuzzing/common/MachOParse.mm
+++ b/Fuzzing/common/MachOParse.mm
@@ -29,7 +29,6 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     // Mach-O Parsing
     [fi architectures];
     [fi isMissingPageZero];
-
     [fi infoPlist];
   }
 
--- a/Source/common/SNTFileInfo.m
+++ b/Source/common/SNTFileInfo.m
@@ -572,10 +572,8 @@ extern NSString *const NSURLQuarantinePropertiesKey WEAK_IMPORT_ATTRIBUTE;
     NSData *cmdData = [self safeSubdataWithRange:NSMakeRange(offset, sz_segment)];
     if (!cmdData) return nil;
 
-    struct load_command *lc = (struct load_command *)[cmdData bytes];
-    if (lc->cmdsize < sizeof(struct load_command)) {
+    if (((struct load_command *)[cmdData bytes])->cmdsize < sizeof(struct load_command))
       return nil;
-    }
 
     if (is64) {
       struct segment_command_64 *lc = (struct segment_command_64 *)[cmdData bytes];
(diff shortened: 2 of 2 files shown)


Fuzzing/common/MachOParse.mm
```diff
@@ -29,6 +29,8 @@ int get_num_fds() {
     // Mach-O Parsing
     [fi architectures];
     [fi isMissingPageZero];
+
```
russellhancox, line 32: Why the empty line?

Source/common/SNTFileInfo.m
```diff
@@ -572,6 +572,11 @@ - (NSDictionary *)embeddedPlist {
     NSData *cmdData = [self safeSubdataWithRange:NSMakeRange(offset, sz_segment)];
     if (!cmdData) return nil;
 
+    struct load_command *lc = (struct load_command *)[cmdData bytes];
+    if (lc->cmdsize < sizeof(struct load_command)) {
+      return nil;
+    }
```
russellhancox, line 578: What's this for? `cmdData` was declared to be the size of either `segment_command_64` or `segment_command` and safeSubdataWithRange will return nil if there isn't enough data to read (which was checked on line 573)

kallsyms: If the command isn't a segment load command, then we only increment `offset` by `lc->cmdsize` which the fuzzer would sometimes set to 0, causing us to re-parse the same command `ncmds` times (up to 2**32 times) eventually OOMing.

russellhancox: Oh, I see now this was in the opening comment, sorry for not reading.

Fuzzing/common/MachOParse.mm
kallsyms: I was originally distinguishing this from the pure Mach-O parsing but since the majority of info plist loading _is_ more binary parsing I'll remove to group it with the others.

Source/common/SNTFileInfo.m
```diff
@@ -572,6 +572,11 @@ - (NSDictionary *)embeddedPlist {
     NSData *cmdData = [self safeSubdataWithRange:NSMakeRange(offset, sz_segment)];
     if (!cmdData) return nil;
 
+    struct load_command *lc = (struct load_command *)[cmdData bytes];
+    if (lc->cmdsize < sizeof(struct load_command)) {
```
russellhancox, line 576: As `lc` isn't needed again you could combine these (which also avoids having `lc` re-declared in an inner scope of a potentially different type):

```
if (((struct load_command *lc)[cmdData bytes])->cmdsize < sz_segment) return nil;
```

This also checks the size is at least `sz_segment` size.

russellhancox (changes requested):

Source/common/SNTFileInfo.m
kallsyms: Consolidated, however some valid commands are smaller than `segment_command` so left the comparison against `sizeof(struct load_command)` (which is just the common fields between all command types).

Source/common/SNTFileInfo.m
```diff
@@ -572,6 +572,9 @@ - (NSDictionary *)embeddedPlist {
     NSData *cmdData = [self safeSubdataWithRange:NSMakeRange(offset, sz_segment)];
     if (!cmdData) return nil;
 
+    if (((struct load_command *)[cmdData bytes])->cmdsize < sizeof(struct load_command))
+      return nil;
```
russellhancox, line 576: These should either be on a single line or have braces.

russellhancox (changes requested):

russellhancox (approved):

PR #992, opened 2023-01-05
More event type support

mlw:
Add support for the TRUNCATE and CREATE event types and add them to the protobuf schema and serializers.


Source/santad/EventProviders/SNTEndpointSecurityFileAccessAuthorizer.mm
```diff
@@ -118,13 +118,46 @@ es_auth_result_t CombinePolicyResults(es_auth_result_t result1, es_auth_result_t
 
 void PopulatePathTargets(const Message &msg, std::vector<PathTarget> &targets) {
   switch (msg->event_type) {
-    case ES_EVENT_TYPE_AUTH_OPEN:
-      PushBackIfNotTruncated(targets, msg->event.open.file, true);
+    case ES_EVENT_TYPE_AUTH_CLONE:
+      PushBackIfNotTruncated(targets, msg->event.clone.source, true);
+      PushBackIfNotTruncated(targets, msg->event.clone.target_dir, msg->event.clone.target_name);
+      break;
+
```
mlw, line 125: Note: Only the CREATE and TRUNCATE cases are new, the rest were just reorganized.

PR #993, opened 2023-01-09
lower fuzz case timeout to 5s

kallsyms:
Got an email about https://github.com/google/santa/actions/runs/3871168437/jobs/6598695596 failing over the weekend, thinking maybe it found something but the VM timed out killing the run before libfuzzer timed out.


pmarkowsky (approved):
LGTM.

PR #994, opened 2023-01-09
Adopt new FS Access Auth config format and policy application logic

mlw:
This change adopts the new config file format proposed here: https://github.com/google/santa/pull/988

This change also adopts the new logic associated with matching instigating processes to matching process configurations - namely that ALL attributes of a configured process must match an instigating process for the policy to apply. (E.g., attributes are AND'd together).

This change also standardizes error handling while parsing the config file to attempt to make error messages more standard and actionable.

### DIFF
--- a/Source/santad/DataLayer/WatchItemPolicy.h
+++ b/Source/santad/DataLayer/WatchItemPolicy.h
@@ -30,12 +30,13 @@ enum class WatchItemPathType {
 };
 
 struct WatchItemPolicy {
+  // TODO: Flip order: make path and path_type next to each other
   WatchItemPolicy(std::string_view n, std::string_view p, bool wo = false,
                   WatchItemPathType pt = WatchItemPathType::kLiteral,
-                  bool ao = true, std::set<std::string> &&abp = {},
-                  std::set<std::string> &&ati = {},
-                  std::set<std::array<uint8_t, CS_CDHASH_LEN>> &&ach = {},
-                  std::set<std::string> &&acs = {})
+                  bool ao = true, std::set<std::string> abp = {},
+                  std::set<std::string> ati = {},
+                  std::set<std::array<uint8_t, CS_CDHASH_LEN>> ach = {},
+                  std::set<std::string> acs = {})
       : name(n),
         path(p),
         write_only(wo),
--- a/Source/santad/DataLayer/WatchItemPolicy.h
+++ b/Source/santad/DataLayer/WatchItemPolicy.h
@@ -35,10 +35,9 @@ static constexpr bool kWatchItemPolicyDefaultAllowReadAccess = false;
 static constexpr bool kWatchItemPolicyDefaultAuditOnly = true;
 
 struct WatchItemPolicy {
-  // TODO: Flip order: make path and path_type next to each other
   WatchItemPolicy(std::string_view n, std::string_view p,
                   WatchItemPathType pt = kWatchItemPolicyDefaultPathType,
-                  bool wo = kWatchItemPolicyDefaultAllowReadAccess,
+                  bool ara = kWatchItemPolicyDefaultAllowReadAccess,
                   bool ao = kWatchItemPolicyDefaultAuditOnly,
                   std::set<std::string> abp = {},
                   std::set<std::string> ati = {},
@@ -47,7 +46,7 @@ struct WatchItemPolicy {
       : name(n),
         path(p),
         path_type(pt),
-        write_only(wo),
+        allow_read_access(ara),
         audit_only(ao),
         allowed_binary_paths(std::move(abp)),
         allowed_team_ids(std::move(ati)),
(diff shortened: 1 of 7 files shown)


kallsyms (dismissed):

Source/santad/DataLayer/WatchItemPolicy.h
```diff
@@ -29,32 +28,63 @@ enum class WatchItemPathType {
...
+  struct Process {
+    Process(std::string bp, std::string ti, std::vector<uint8_t> cdh,
+            std::string ch)
+        : binary_path(bp),
+          team_id(ti),
+          cdhash(std::move(cdh)),
+          certificate_sha256(ch) {}
+
+    bool operator==(const Process &other) const {
+      return binary_path == other.binary_path && team_id == other.team_id &&
```
russellhancox, line 46: It'd be nice to keep one condition per-line, it's more readable.

Of course, if clang-format overrides this, then ignore.

Source/santad/DataLayer/WatchItemPolicy.h
```diff
@@ -29,32 +28,63 @@ enum class WatchItemPathType {
...
+        allow_read_access(ara),
         audit_only(ao),
-        allowed_binary_paths(std::move(abp)),
-        allowed_team_ids(std::move(ati)),
-        allowed_cdhashes(std::move(ach)),
-        allowed_certificates_sha256(std::move(acs)) {}
+        processes(std::move(procs)) {}
+
+  bool operator==(const WatchItemPolicy &other) const {
+    return name == other.name && path == other.path &&
```
russellhancox, line 72: Same here

Source/santad/DataLayer/WatchItems.mm
```diff
@@ -28,168 +29,443 @@
...
 
-  for (int i = 0; i < str.length; i++) {
-    if (!isxdigit([str characterAtIndex:i])) {
+  std::vector<uint8_t> bytes;
+  bytes.reserve(str.length / 2);
+
+  char cur_byte[3];
+  cur_byte[2] = '\0';
+
+  for (int i = 0; i < [str length] / 2; i++) {
```
russellhancox, line 109: s/`[str length]`/`str.length`/

Source/santad/EventProviders/SNTEndpointSecurityFileAccessAuthorizer.mm
```diff
@@ -310,6 +310,48 @@ - (FileAccessPolicyDecision)specialCaseForPolicy:(std::shared_ptr<WatchItemPolic
   return FileAccessPolicyDecision::kNoPolicy;
 }
 
+/// An An `es_process_t` must match all criteria within the given
+/// WatchItemPolicy::Process to be considered a match.
+- (bool)policyProcess:(const WatchItemPolicy::Process &)policyProc
+     matchesESProcess:(const es_process_t *)esProc {
+  if (esProc->codesigning_flags & CS_SIGNED) {
```
russellhancox, line 317: Shouldn't you also be checking for CS_VALID?

Source/santad/EventProviders/SNTEndpointSecurityFileAccessAuthorizer.mm
```diff
@@ -310,6 +310,48 @@ - (FileAccessPolicyDecision)specialCaseForPolicy:(std::shared_ptr<WatchItemPolic
...
+    // Check if the instigating process has an allowed TeamID
+    if (policyProc.team_id.length() > 0 && esProc->team_id.data &&
+        policyProc.team_id != esProc->team_id.data) {
+      return false;
+    }
```
pmarkowsky, line 322: nit: team ID is checked last elsewhere in the code. 

In this case it doesn't matter as we said that a process has to match *all* specified fields.

Source/santad/DataLayer/WatchItemPolicy.h
mlw: clang-format won't let that slide :(

Source/santad/EventProviders/SNTEndpointSecurityFileAccessAuthorizer.mm
mlw: The `CS_VALID` check is handled prior to this method being called. This method is used for individually checking each of the configured process exceptions while the check for a valid code signature is more broad and applies whether or not process exceptions exist.

This does look confusing though and I'll add a comment.

Source/santad/EventProviders/SNTEndpointSecurityFileAccessAuthorizer.mm
mlw: Spoke offline. Will leave as-is since it's in order of expected evaluation cost.

pmarkowsky (approved):
LGTM.

PR #998, opened 2023-01-11
Support configuring signing IDs for process exceptions

mlw:
Also expanded on some negative test cases for all attributes


pmarkowsky (approved):
LGTM.

PR #1003, opened 2023-01-18
Configurator: Apply config updates in non-daemon processes

### LINKED ISSUE #1002: After changing the profile,some fields cannot take effect immediately
Hey,
When I change fields such as BannedBlockMessage, EventDetailText, EventDetailURL, UnknownBlockMessage and other fields of the profile, these configuration fields will only take effect after I restart the machine. 

**Is there a command to reload the santa-related service components to make the changes take effect immediately?** 

**If there is no such command, can the code be optimized to support this requirement?**

russellhancox:
Fixes #1002.


pmarkowsky (approved):
LGTM.

Issue #1006, opened 2023-01-20
UI: Re-write GUI using SwiftUI


russellhancox:
The existing Santa UI is very basic and uses 'old-fashioned' XIBs for layout etc. The UI hasn't been modified in quite a long time and any changes now would be difficult to review. We're also as a team not using Xcode anymore so requiring interface builder for modifications is suboptimal too.

SwiftUI is the future of UI-work on Apple platforms and has very nice syntax for programmatically creating views. While we generally have avoided Swift (for good reasons), this limited use is easy to integrate, particularly thanks to the good interop with existing ObjC code.

PR #1007, opened 2023-01-20
GUI: Re-write AboutWindow view in SwiftUI

russellhancox:
This is the first in a series of PRs that re-write the GUI using SwiftUI. The AboutWindow is the simplest of the views/windows that it shows so I've started there. This is functionally equivalent and I've tested all the configurable states (no `MoreInfoURL` configured, valid `MoreInfoURL` configured, no `AboutText` configured, short `AboutText` configured, long `AboutText` configured.).

Part of #1006.


Source/gui/SNTAboutWindowView.swift
```diff
@@ -0,0 +1,65 @@
...
+      } else {
+          Text("""
+          Santa is an application control system for macOS.
+
+          There are no user-configurable settings.
+          """).multilineTextAlignment(.center)
+      }
+
+      HStack {
+        if let u = c.moreInfoURL {
```
tburgin, line 30: We can skip the nested `if` with:
```
        if c.moreInfoURL?.absoluteString.isEmpty == false {
          Button(action: moreInfoButton) {
            Text("More Info...").frame(width: 90.0)
          }
        }
```

Source/gui/SNTAboutWindowController.h
```diff
@@ -14,11 +14,5 @@
...
 #import <Cocoa/Cocoa.h>
 
-@interface SNTAboutWindowController : NSWindowController
-
-@property IBOutlet NSTextField *aboutTextField;
-@property IBOutlet NSButton *moreInfoButton;
-
-- (IBAction)openMoreInfoURL:(id)sender;
-
+@interface SNTAboutWindowController : NSWindowController <NSWindowDelegate>
```
tburgin, line 17: Has this file been formatted? I thought protocol subscriptions didn't have a space? `NSWindowController<NSWindowDelegate>`

russellhancox: It didn't have a space and the linter complained, fix.sh put the space in.

tburgin (approved):

PR #1008, opened 2023-01-20
Revert "Configurator: Apply config updates in non-daemon processes"

russellhancox:
There's a chance this is causing issues with `santactl sync` not picking up the SyncBaseURL.

Reverts google/santa#1003


pmarkowsky (approved):
LGTM.

PR #1009, opened 2023-01-23
shadow rules_python for fuzzing

kallsyms:
The update to rules_python needed for Swift breaks fuzzing. Pin the version that rules_fuzzing requires in a new target, and shadow rules_python with it in fuzzing.


pmarkowsky (approved):
LGTM.

mlw (approved):

PR #1010, opened 2023-01-23
GUI: Migrate DeviceMessageWindow to SwiftUI

russellhancox:
Functionally equivalent to the existing window, the only real difference being that there is a titlebar as this allows the window to immediately be key, allowing the Enter key to dismiss without having to interact with the window itself. Santa.app also becomes a foreground app when this notification is on-screen, so it appears in the dock and app switcher.

Part of #1006


tburgin:
A General question about the move to SwiftUI.

Xcode has a wonderful GUI editor that generates SwiftUI code. Can we take advantage of this somehow? Maybe a reduced Xcode project that allows us to edit the .swift files with the help of Xcode? I realize we don't want to maintain two build systems, but it may be worth it to maintain an Xcode project with SwiftUI + the necessary Santa decencies.

Do you see this as useful or necessary? It does seem like you have been getting on okay without Xcode. What has the development cycle been like for you?

mlw (approved):
Difficult to review UI code... but nothing jumps out at me as concerning.

PR #1011, opened 2023-01-25
Ensure watch item names conform to naming requirements

mlw:
Names must be legal C identifiers.

### DIFF
--- a/Source/santad/DataLayer/WatchItems.mm
+++ b/Source/santad/DataLayer/WatchItems.mm
@@ -421,6 +421,29 @@ bool ParseConfigSingleWatchItem(NSString *name, NSDictionary *watch_item,
   return true;
 }
 
+bool IsWatchItemNameValid(NSString *watch_item_name, NSError **err) {
+  if (!watch_item_name) {
+    // This shouldn't be possible as written, but handle just in case
+    PopulateError(err, [NSString stringWithFormat:@"nil watch item name"]);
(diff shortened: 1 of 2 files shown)


Source/santad/DataLayer/WatchItems.mm
```diff
@@ -421,6 +421,29 @@ bool ParseConfigSingleWatchItem(NSString *name, NSDictionary *watch_item,
...
+  if (!watch_item_name) {
+    // This shouldn't be possible as written, but handle just in case
+    PopulateError(err, [NSString stringWithFormat:@"nil watch item name"]);
+    return false;
+  }
+
+  NSRegularExpression *regex =
+    [NSRegularExpression regularExpressionWithPattern:@"^[A-Za-z_][A-Za-z0-9_]*$"
+                                              options:0
+                                                error:nil];
```
russellhancox, line 434: WDYT about making this static with a dispatch_once() for initialization? It's my understanding that this is quite slow.

kallsyms (approved):

Source/santad/DataLayer/WatchItems.mm
mlw: Done. Realistically this code is run once every 10 min by default. But if you had a bunch of rules would compile a bunch of time unnecessarily.

russellhancox (approved):

PR #1019, opened 2023-01-27
Fix: Rewrite the SNTMetricHTTPWriter to avoid potential stack corruption

pmarkowsky:
We've seen some intermittent crashes when logging metrics via the SNTMetricHTTPWriter.

### DIFF
--- a/Source/santametricservice/Writers/SNTMetricHTTPWriter.m
+++ b/Source/santametricservice/Writers/SNTMetricHTTPWriter.m
@@ -43,22 +43,21 @@
 
   _authSession.serverHostname = url.host;
   NSURLSession *_session = _authSession.session;
+  NSURLSessionDataTask *task;
 
   dispatch_group_t requests = dispatch_group_create();
-
-  [metrics enumerateObjectsUsingBlock:^(id value, NSUInteger index, BOOL *stop) {
+  for (NSData *metric in metrics) {
     dispatch_group_enter(requests);
 
-    request.HTTPBody = (NSData *)value;
-    NSURLSessionDataTask *task = [_session
+    request.HTTPBody = (NSData *)metric;
+    task = [_session
       dataTaskWithRequest:request
         completionHandler:^(NSData *_Nullable data, NSURLResponse *_Nullable response,
                             NSError *_Nullable err) {
           if (err != nil) {
             _blockError = err;
-            *stop = YES;
           } else if (response == nil) {
-            *stop = YES;
+            _blockError = nil;
           } else if ([response isKindOfClass:[NSHTTPURLResponse class]]) {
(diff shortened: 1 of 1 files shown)


Source/santametricservice/Writers/SNTMetricHTTPWriter.m
```diff
@@ -43,22 +43,21 @@ - (BOOL)write:(NSArray<NSData *> *)metrics toURL:(NSURL *)url error:(NSError **)
...
+    task = [_session
       dataTaskWithRequest:request
         completionHandler:^(NSData *_Nullable data, NSURLResponse *_Nullable response,
                             NSError *_Nullable err) {
           if (err != nil) {
             _blockError = err;
-            *stop = YES;
           } else if (response == nil) {
-            *stop = YES;
+            _blockError = nil;
```
mlw, line 60: This will overwrite any previous error from other requests

mlw: Also should ensure that the assignments are atomic (for both here and the assignment above when in the case `err != nil`).

mlw (approved):

mlw (changes requested):
Clicked the wrong button before - see comments above

Source/santametricservice/Writers/SNTMetricHTTPWriter.m
```diff
@@ -72,30 +74,29 @@ - (BOOL)write:(NSArray<NSData *> *)metrics toURL:(NSURL *)url error:(NSError **)
                           [NSString stringWithFormat:@"received http status code %ld from %@",
                                                      httpResponse.statusCode, url]
                       }];
-              *stop = YES;
+            LOGD(@"SNTMetricHTTPWriter: %@", err);
```
mlw, line 77: This left in intentionally?

Source/santametricservice/Writers/SNTMetricHTTPWriter.m
```diff
@@ -72,30 +71,28 @@ - (BOOL)write:(NSArray<NSData *> *)metrics toURL:(NSURL *)url error:(NSError **)
...
           }
           dispatch_group_leave(requests);
         }];
 
     [task resume];
+  }
 
-    SNTConfigurator *config = [SNTConfigurator configurator];
-    int64_t timeout = (int64_t)config.metricExportTimeout;
+  SNTConfigurator *config = [SNTConfigurator configurator];
```
mlw: Minor: Could we move this to the top of the function, something like `static SNTConfigurator *config = [SNTConfigurator configurator];`. Even though this code won't be hit that often, we've seen in measurements that this lookup takes a surprising amount of time, would be good to eliminate it when sending metrics.

mlw (approved):

PR #1030, opened 2023-02-01
Small test fixes to make sanitizers happy

kallsyms:
* Lower the test count from 10->5 since we're running _just_ over 1 hour and getting killed sometimes
* Set the `event_type` of an `es_message` which was uninitialized causing us to sporadically call `RespondAuthFlags` instead of `RespondAuthResult`. No MSAN for macOS which would have caught this directly :(
* Reinstantiate `mockAuthClient` in each "sub-test" since TSAN sees a data race in mock recording

### DIFF
--- a/Source/santad/EventProviders/SNTEndpointSecurityAuthorizerTest.mm
+++ b/Source/santad/EventProviders/SNTEndpointSecurityAuthorizerTest.mm
@@ -117,10 +117,10 @@ class MockAuthResultCache : public AuthResultCache {
   {
     SNTEndpointSecurityAuthorizer *authClient =
       [[SNTEndpointSecurityAuthorizer alloc] initWithESAPI:mockESApi
-                                                  metrics:nullptr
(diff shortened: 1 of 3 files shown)


Source/santad/EventProviders/SNTEndpointSecurityAuthorizerTest.mm
```diff
@@ -145,11 +145,22 @@ - (void)testHandleMessage {
...
+
+    [mockAuthClient stopMocking];
   }
 
   // Test SNTExecutionController determines the event should be processed and
   // processMessage:handler: is called.
   {
+    SNTEndpointSecurityAuthorizer *authClient =
+      [[SNTEndpointSecurityAuthorizer alloc] initWithESAPI:mockESApi
+                                                  metrics:nullptr
```
russellhancox, line 157: Same

mlw (dismissed):

PR #1034, opened 2023-02-03
Replace SNTDecisionCache dictionary with SantaCache

mlw:
CPU traces were showing a lot of contention in the dispatch blocks synchronizing access to the dictionary. This helps relieve the contention. Also removes a second lookup for a cached decision that was previously necessary on the EXEC logging path.

### DIFF
--- a/Source/santad/SNTDecisionCache.mm
+++ b/Source/santad/SNTDecisionCache.mm
@@ -24,7 +24,6 @@
 #import "Source/santad/SNTDatabaseController.h"
 
 @interface SNTDecisionCache ()
-@property dispatch_queue_t detailStoreQueue;
 // Cache for sha256 -> date of last timestamp reset.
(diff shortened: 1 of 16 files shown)


Source/santad/SNTDecisionCache.mm
```diff
@@ -44,7 +48,7 @@ - (instancetype)init {
     // TODO(mlw): We should protect this structure with a read/write lock
     // instead of a serial dispatch queue since it's expected that most most
     // accesses will be lookups, not caching new items.
-    _detailStore = [NSMutableDictionary dictionaryWithCapacity:1000];
+    // _detailStore = [NSMutableDictionary dictionaryWithCapacity:1000];
```
russellhancox, line 51: Can remove? Along with the queue and the comment above?

russellhancox (approved):

PR #1036, opened 2023-02-06
Report log type in santactl status

mlw:


Source/santactl/Commands/SNTCommandStatus.m
```diff
@@ -69,6 +69,8 @@ - (void)runWithArguments:(NSArray *)arguments {
   }];
 
   BOOL fileLogging = ([[SNTConfigurator configurator] fileChangesRegex] != nil);
+  NSString *eventLogType =
+    [[[SNTConfigurator configurator] eventLogTypeRaw] lowercaseString] ?: @"unknown";
```
russellhancox, line 73: The fallback case here is impossible

russellhancox (approved):

PR #1041, opened 2023-02-16
Add support for platform binary to process exceptions

mlw:


Source/santad/DataLayer/WatchItemPolicy.h
```diff
@@ -36,18 +37,22 @@ static constexpr bool kWatchItemPolicyDefaultAuditOnly = true;
...
 
     bool operator==(const Process &other) const {
       return binary_path == other.binary_path &&
              signing_id == other.signing_id && team_id == other.team_id &&
              cdhash == other.cdhash &&
-             certificate_sha256 == other.certificate_sha256;
+             certificate_sha256 == other.certificate_sha256 &&
+             platform_binary.has_value() == other.platform_binary.has_value() &&
+             platform_binary.value_or(false) ==
+                 other.platform_binary.value_or(false);
```
russellhancox, line 55: Am I reading correctly that this allows asserting in a policy that a binary is **not** a platform binary?

russellhancox: (I don't really mean that line specifically, but it was the easiest place to add the comment)

mlw: Yes, it does support that. At first I was only going to support `PlatformBinary=True`, but it seemed weird that `False` meant "don't check", so i added the 3rd state via the `optional`.

russellhancox: SGTM. I had started writing a comment about not needing the has_value condition if we can just treat unset as false but supporting explicit false makes sense, I can see uses for it.

russellhancox (approved):

PR #1050, opened 2023-03-08
Fix santactl rule team id validation

pmarkowsky:
This PR adds validation for the team IDs in the `santactl rule` command. It also adds the ability to get a teamid from a binary specified via a path.

### DIFF
--- a/Source/santactl/Commands/SNTCommandRule.m
+++ b/Source/santactl/Commands/SNTCommandRule.m
@@ -126,7 +126,7 @@ REGISTER_COMMAND_NAME(@"rule")
 #else
   ) {
 #endif
-    printf("(SyncBaseURL/StaticRules is set, rules are managed centrally.)");
+    printf("(SyncBaseURL/StaticRules is set, rules are managed centrally.)\n");
     exit(1);
   }
 
--- a/Source/santactl/Commands/SNTCommandRule.m
+++ b/Source/santactl/Commands/SNTCommandRule.m
@@ -13,6 +13,7 @@
 ///    limitations under the License.
 
 #import <Foundation/Foundation.h>
+#include "Source/common/SNTCommonEnums.h"
 #import <MOLCertificate/MOLCertificate.h>
 #import <MOLCodesignChecker/MOLCodesignChecker.h>
 #import <MOLXPCConnection/MOLXPCConnection.h>
@@ -64,6 +65,28 @@ REGISTER_COMMAND_NAME(@"rule")
   NSCharacterSet *validChars = [NSCharacterSet alphanumericCharacterSet];
 
   return [@"" isEqualToString:[teamID stringByTrimmingCharactersInSet:validChars]];
+}
+
+- (void) checkIfRuleIsWellformed:(SNTRule *)rule{
+  switch (rule.type) {
+    case SNTRuleTypeBinary:
+    case SNTRuleTypeCertificate:
+      if (![self isWellformedSHA256:[rule.identifier uppercaseString]]) {
+        [self printErrorUsageAndExit:@"BINARY or CERTIFICATE rules require a valid SHA-256"];
+      }
+      break;
+    case SNTRuleTypeTeamID:
+      if (![self isWellformedTeamID:[rule.identifier uppercaseString]]) {
+        [self printErrorUsageAndExit:
+                @"Invalid Team ID supplied (must be 10 alphanumeric characters (e.g. EQHXZ8M8AV)"];
+      }
+      break;
+    default:
+      // This is a programming error.
+      [self printErrorUsageAndExit:@"Invalid rule type specified"];
+      break;
+  }
+
 }
 
 + (NSString *)longHelpText {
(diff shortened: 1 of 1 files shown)


Source/santactl/Commands/SNTCommandRule.m
```diff
@@ -45,6 +46,49 @@ + (NSString *)shortHelpText {
...
+  if (teamID.length != 10) {
+    return NO;
+  }
+
+  NSCharacterSet *validChars = [NSCharacterSet alphanumericCharacterSet];
+
+  return [@"" isEqualToString:[teamID stringByTrimmingCharactersInSet:validChars]];
+}
+
+- (void) checkIfRuleIsWellformed:(SNTRule *)rule{
```
russellhancox: No space before method name

Source/santactl/Commands/SNTCommandRule.m
```diff
@@ -13,6 +13,7 @@
 ///    limitations under the License.
 
 #import <Foundation/Foundation.h>
+#include "Source/common/SNTCommonEnums.h"
```
russellhancox: Group with the other project imports below

Source/santactl/Commands/SNTCommandRule.m
```diff
@@ -45,6 +46,49 @@ + (NSString *)shortHelpText {
...
 
+- (BOOL)isWellformedSHA256:(NSString *)hash {
+  if (hash.length != 64) {
+    return NO;
+  }
+
+  NSCharacterSet *hexChars =
+    [NSCharacterSet characterSetWithCharactersInString:@"0123456789ABCDEF"];
+
+  return [@"" isEqualToString:[hash stringByTrimmingCharactersInSet:hexChars]];
```
russellhancox, line 57: I think it'll be quicker in the average case to filter out the inverted set and _then_ check the length == 64.

Source/santactl/Commands/SNTCommandRule.m
```diff
@@ -45,6 +46,49 @@ + (NSString *)shortHelpText {
...
+}
+
+- (BOOL)isWellformedTeamID:(NSString *)teamID {
+  if (teamID.length != 10) {
+    return NO;
+  }
+
+  NSCharacterSet *validChars = [NSCharacterSet alphanumericCharacterSet];
+
+  return [@"" isEqualToString:[teamID stringByTrimmingCharactersInSet:validChars]];
```
russellhancox, line 67: Same here

Source/santactl/Commands/SNTCommandRule.m
```diff
@@ -45,6 +46,49 @@ + (NSString *)shortHelpText {
...
+}
+
+- (void) checkIfRuleIsWellformed:(SNTRule *)rule{
+  switch (rule.type) {
+    case SNTRuleTypeBinary:
+    case SNTRuleTypeCertificate:
+      if (![self isWellformedSHA256:[rule.identifier uppercaseString]]) {
```
russellhancox, line 74: We should uppercase the identifier when parsing the arguments instead so it's just one location it needs to be handled.

Source/santactl/Commands/SNTCommandRule.m
```diff
@@ -171,6 +213,13 @@ - (void)runWithArguments:(NSArray *)arguments {
       MOLCodesignChecker *cs = [fi codesignCheckerWithError:NULL];
       newRule.identifier = cs.leafCertificate.SHA256;
     } else if (newRule.type == SNTRuleTypeTeamID) {
+      // Get the team ID from the binary.
+      MOLCodesignChecker *cs = [fi codesignCheckerWithError:NULL];
+      if (![self isWellformedTeamID:cs.signingInformation[@"teamid"]]) {
+        [self printErrorUsageAndExit:
+                @"Invalid Team ID in binary (must be 10 alphanumeric characters (e.g. EQHXZ8M8AV)"];
+      };
```
russellhancox, line 218: This is an odd check, generally if there isn't a team ID the binary is unsigned or adhoc signed, we should report that as this error could be confusing.

russellhancox (changes requested):

Source/santactl/Commands/SNTCommandRule.m
mlw: If I'm understanding correctly, I think this could leave invalid data. E.g. the string `AA...Z...AA` might return a length of 64 after being trimmed, but the charset is still invalid

pmarkowsky: @russellhancox So you mean something like the following?

```objective-c
- (BOOL)isWellformedSHA256:(NSString *)hash {
NSCharacterSet *nonHexChars =
    [[NSCharacterSet characterSetWithCharactersInString:@"0123456789ABCDEF"] invertedSet];

  return [hash stringByTrimmingCharactersInSet:nonHexChars].length == 64;
}
```

russellhancox: Yeah, perfect

mlw: `stringByTrimmingCharactersInSet` only trims from the front and back of a string. The proposed method doesn't properly verify strings like the following, that start and end with valid chars, but have invalid chars in the middle:
```
[self isWellformedSHA256:@"A!KLM(DSNSINNCIOSDNMLKSDM<SDM><><S>D<F>SD<F<EWIOJFEWJCOIWECNM./A"];
```
This returns `true`, but obviously not desired.

russellhancox: Ah, good point. I don't know if NSString has a better way but this would work:

```objc
	NSCharacterSet *nonHexChars =
	    [[NSCharacterSet characterSetWithCharactersInString:@"0123456789abcdef"] invertedSet];
	hash = [[hash componentsSeparatedByCharactersInSet:nonHexChars] componentsJoinedByString:@""];
	return hash.length == 64;
```

mlw: Seems like this would probably work. But also I think this would require at least 2 copies and additional allocations. I don't think this code is on a hot path so (??) so it maybe doesn't matter, but the original impl is likely more performant. 

If this is hot path and we do care about speed, it might be worth abstracting the code in `WatchItems.mm` that checks for valid hex strings of a given length in one pass without extra allocations/copies: https://github.com/google/santa/blob/main/Source/santad/DataLayer/WatchItems.mm#LL89

russellhancox: This isn't on a hot path but we should put that somewhere in common and re-use it, so we don't have multiple implementations of the same thing. I'm sure there are other places it could (or should) be used.

pmarkowsky:
I'm closing this to move the validation into SNTRule will open a separate PR.

PR #1060, opened 2023-03-27
sync: Allow server to override the header for transmitting XSRF tokens

russellhancox:
This change allows a sync server to change the header that Santa will use to send XSRF tokens on subsequent requests by putting the header name in the  header.


pmarkowsky (dismissed):
LGTM. We should probably add an integration test for this.

russellhancox:
> LGTM. We should probably add an integration test for this.

Done.

pmarkowsky (approved):
LGTM.

PR #1063, opened 2023-03-30
Change types of repeated args and envs fields

mlw:
This change is necessary in order to properly encode non-UTF8 strings which aren't too uncommon in exec args/envs. Proto spec requires strings to be UTF8.

This is a compatible change - `string --> bytes` type change is safe.. Proto generated via the old schema can be successfully unmarshalled via the new schema.

### DIFF
--- a/Source/common/santa.proto
+++ b/Source/common/santa.proto
@@ -1,3 +1,4 @@
+
 // Important: This schema is currently in BETA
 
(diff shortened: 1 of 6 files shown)


pmarkowsky (approved):
LGTM.

russellhancox (approved):

PR #1064, opened 2023-03-31
Rearrange string log fields to ease text parsing

mlw:


mlw:
Not going to pursue this for now due to compat concerns.

PR #1068, opened 2023-04-06
sync: Add more complete XSSI prefix to be stripped.

russellhancox:
Sync will try stripping both the new longer prefix and the existing short prefix if the response data begins with either. This should have no impact on existing sync servers but will allow sync servers in the future to use the longer prefix if they wish.

### DIFF
--- a/Source/santasyncservice/SNTSyncStage.m
+++ b/Source/santasyncservice/SNTSyncStage.m
@@ -183,9 +183,15 @@
 }
 
 - (NSData *)stripXssi:(NSData *)data {
-  static const char xssi[3] = {']', ')', '}'};
-  if (data.length < 3 || strncmp(data.bytes, xssi, 3)) return data;
-  return [data subdataWithRange:NSMakeRange(3, data.length - 3)];
+  static const char xssiOne[5] = {')', ']', '}', '\'', '\n'};
+  static const char xssiTwo[3] = {']', ')', '}'};
+  if (data.length >= 5 && strncmp(data.bytes, xssiOne, 5) == 0) {
+    return [data subdataWithRange:NSMakeRange(5, data.length - 5)];
+  }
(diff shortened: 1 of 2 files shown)


Source/santasyncservice/SNTSyncStage.m
```diff
@@ -183,9 +183,15 @@ - (NSData *)performRequest:(NSURLRequest *)request
...
-  if (data.length < 3 || strncmp(data.bytes, xssi, 3)) return data;
-  return [data subdataWithRange:NSMakeRange(3, data.length - 3)];
+  static const char xssiOne[5] = {')', ']', '}', '\'', '\n'};
+  static const char xssiTwo[3] = {']', ')', '}'};
+  if (data.length >= 5 && strncmp(data.bytes, xssiOne, 5) == 0) {
+    return [data subdataWithRange:NSMakeRange(5, data.length - 5)];
+  }
+  if (data.length >= 3 && strncmp(data.bytes, xssiTwo, 3) == 0) {
+    return [data subdataWithRange:NSMakeRange(3, data.length - 3)];
+  }
```
mlw, line 193: Nit: We should use `sizeof` instead of the hardcoded values
```suggestion
  if (data.length >= sizeof(xssiOne) && strncmp(data.bytes, xssiOne, sizeof(xssiOne)) == 0) {
    return [data subdataWithRange:NSMakeRange(sizeof(xssiOne), data.length - sizeof(xssiOne))];
  }
  if (data.length >= sizeof(xssiTwo) && strncmp(data.bytes, xssiTwo, sizeof(xssiTwo)) == 0) {
    return [data subdataWithRange:NSMakeRange(sizeof(xssiTwo), data.length - sizeof(xssiTwo))];
  }
```

mlw (changes requested):

mlw (approved):

pmarkowsky (approved):
LGTM.

PR #1070, opened 2023-04-13
Fix string length issues

mlw:
This PR addresses issues mostly related to proto encoding where an improper length was being used when copying UTF8String strings.


Source/santad/Logs/EndpointSecurity/Serializers/Protobuf.mm
pmarkowsky, line 1: The lambdas make this pretty clean.

pmarkowsky (approved):
LGTM.

PR #1076, opened 2023-04-21
Make the sync client content encoding a tunable

pmarkowsky:
This adds a new configuration option `SyncClientContentEncoding` which allows you to tweak the content encoding used by the sync client. Currently this supports the following options:

| Value | Expected Behavior |
| --- | --- |
| "" (default) | zlib with `Content-Encoding` set to `deflate` or "zlib" depending on `EnableBackwardsCompatibleContentEncoding` *old behavior* |
| "gzip" | gzip compressed requests with `Content-Encoding` set to `gzip` |
| "zlib" |   zlib with `Content-Encoding` set to `deflate`   |
| "none" | Content is not compressed at all |

### Testing

Tested with a hacked up version of moroz.

### Headers when `SyncClientContentEncoding` is set to `None`

```
POST /v1/santa/preflight/<machine_id>
Host: santa:8080
Content-Type: application/json
Connection: keep-alive
Accept: */*
User-Agent: santactl-sync/9999.1.1
Content-Length: 536
Accept-Language: en-US,en;q=0.9
Accept-Encoding: gzip, deflate
```

### DIFF
--- a/Source/common/SNTCommonEnums.h
+++ b/Source/common/SNTCommonEnums.h
@@ -13,6 +13,7 @@
 ///    limitations under the License.
 
 #import <Foundation/Foundation.h>
+#include <objc/objc-api.h>
 
 ///
 ///  These enums are used in various places throughout the Santa client code.
@@ -127,6 +128,12 @@ typedef NS_ENUM(NSInteger, SNTSyncStatusType) {
   SNTSyncStatusTypeDaemonTimeout,
   SNTSyncStatusTypeSyncStarted,
   SNTSyncStatusTypeUnknown,
+};
+
+typedef NS_ENUM(NSInteger, SNTSyncCompressionEncoding) {
+  SNTSyncCompressionEncodingNone,
+  SNTSyncCompressionEncodingZlib,
+  SNTSyncCompressionEncodingGzip,
 };
 
 typedef NS_ENUM(NSInteger, SNTMetricFormatType) {
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -410,6 +410,8 @@
 ///
 @property(nonatomic) BOOL syncCleanRequired;
 
+#pragma mark - USB Settings
+
 ///
 /// USB Mount Blocking. Defaults to false.
 ///
@@ -519,6 +521,12 @@
 ///  Defaults to false.
 ///
 @property(readonly, nonatomic) BOOL enableBackwardsCompatibleContentEncoding;
+
+///
+/// If set, "santactl sync" will use the supplied "Content-Encoding", possible
+/// settings include gzip, deflate, none.
+///
+@property(readonly, nonatomic) SNTSyncCompressionEncoding syncClientContentEncoding;
 
 ///
 ///  Contains the FCM project name.
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -13,6 +13,7 @@
 ///    limitations under the License.
 
 #import "Source/common/SNTConfigurator.h"
+#import "Source/common/SNTCommonEnums.h"
 
 #include <sys/stat.h>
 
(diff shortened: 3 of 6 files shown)


Source/common/SNTCommonEnums.h
```diff
@@ -13,6 +13,7 @@
 ///    limitations under the License.
 
 #import <Foundation/Foundation.h>
+#include <objc/objc-api.h>
```
russellhancox, line 16: What's this for?

Source/common/SNTConfigurator.m
```diff
@@ -715,6 +717,20 @@ - (NSString *)syncClientAuthCertificateIssuer {
   return self.configState[kClientAuthCertificateIssuerKey];
 }
 
+- (SNTSyncCompressionEncoding)syncClientContentEncoding {
+  NSString *contentEncoding = [self.configState[kClientContentEncoding] lowercaseString];
+  if ([contentEncoding isEqualToString:@"zlib"]) {
```
russellhancox, line 722: Make this deflate, to match what the sent heading is

Source/common/SNTConfigurator.m
```diff
@@ -715,6 +717,20 @@ - (NSString *)syncClientAuthCertificateIssuer {
...
+    return SNTSyncCompressionEncodingZlib;
+  } else if ([contentEncoding isEqualToString:@"gzip"]) {
+    return SNTSyncCompressionEncodingGzip;
+  } else if ([contentEncoding isEqualToString:@"none"]) {
+    return SNTSyncCompressionEncodingNone;
+  } else {
+    // Ensure we have the same
```
russellhancox, line 729: the same...?

Source/santasyncservice/SNTSyncState.h
```diff
@@ -13,6 +13,7 @@
 ///    limitations under the License.
 
 #import <Foundation/Foundation.h>
+#include <objc/NSObjCRuntime.h>
```
russellhancox, line 16: What for?

Source/common/SNTConfigurator.m
pmarkowsky: That should have read `Ensure we have the same default zlib behavior Santa's always had by default.`

pmarkowsky: I'll fix the wording.

Source/common/SNTCommonEnums.h
pmarkowsky: I think this is a clangd artifact that I didn't clean up.

Source/santasyncservice/SNTSyncState.h
pmarkowsky: This is another clangd artifact.

Source/common/SNTConfigurator.h
```diff
@@ -520,6 +522,12 @@
 ///
 @property(readonly, nonatomic) BOOL enableBackwardsCompatibleContentEncoding;
```
russellhancox, line 523: I think we should remove this property and the key that it refers to, and add a note in the release notes that it's gone.

pmarkowsky: Works for me.

Source/common/SNTConfigurator.m
mlw: I find this a bit confusing. It seems odd to have the configuration string of `gzip` or `deflate` be either a format or algorithm.

mlw: We talked offline a bit. I guess it's odd, but fairly consistent with how inconsistently these things are referred to throughout the web...

Source/common/SNTCommonEnums.h
```diff
@@ -129,6 +129,12 @@ typedef NS_ENUM(NSInteger, SNTSyncStatusType) {
   SNTSyncStatusTypeUnknown,
 };
 
+typedef NS_ENUM(NSInteger, SNTSyncContentEncoding) {
+  SNTSyncContentEncodingNone,
+  SNTSyncContentEncodingZlib,
```
mlw, line 134: If we're going to key off the string `deflate` for the configuration value, I think it would be nice to also name the enum the same.

pmarkowsky: Renamed to `SNTSyncContentEncodingDeflate`.

mlw (approved):

PR #1077, opened 2023-04-21
Docs: Added instructions for how to use config-overrides.plist

pmarkowsky:
This PR adds instructions for how to use config overrides on dev versions of Santa.that was added in #957.

### DIFF
--- a/docs/development/building.md
+++ b/docs/development/building.md
@@ -133,8 +133,8 @@ bazel test :unit_tests --define=SANTA_BUILD_TYPE=adhoc --test_output=errors
 Debug versions of Santa have the ability to set/override config settings using an override file, that will be applied over the top of the configuration from a profile.
 
 1. Create a plist in `/var/db/santa/config-overrides.plist`
-  - Make sure it's readable. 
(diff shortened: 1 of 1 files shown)


docs/development/building.md
```diff
@@ -128,6 +128,28 @@ Run all the logic / unit tests
 bazel test :unit_tests --define=SANTA_BUILD_TYPE=adhoc --test_output=errors
 ```
 
+##### Testing Config Options Using `/var/db/santa/config-overrides.plist`
+
+Debug versions of Santa have the ability to set/override config settings using
```
russellhancox, line 133:
```suggestion
Debug versions of Santa have the ability to set/override config settings using an override file, that will be applied over the top of the configuration from a profile.
```

mlw (dismissed):

russellhancox (approved):

pmarkowsky:
I'm holding off on merging this until I figure out why the linter is failing. As I don't want all of our future doc commits to also fail.

mlw (approved):

PR #1078, opened 2023-04-23
Inject additional dependencies into the serializers

mlw:
The serializers were using static variables in a couple of places that was fine for production, but made some types of tests difficult since the static variables would outlive the mocks.

This change should have no functional impact on Santa - nothing should change at runtime as part of this restructuring to inject the dependencies.


mlw:
Please hold up on review - going to store the client mode in the SNTCachedDecision so that lazy client mode lookup isn't needed (also fixes a long standing race that could occur where the mode applied at policy time might not match the mode at logging time, tho this would be quite rare).

mlw:
PR updated to store the client mode at decision time and remove the client mode lookup function injection.

Good to resume reviewing!

Source/common/SNTCachedDecision.h
```diff
@@ -42,4 +42,6 @@
 @property NSString *customMsg;
 @property BOOL silentBlock;
 
+@property SNTClientMode decisionClientMode;
```
russellhancox, line 45: Group this with the decision further up

russellhancox (changes requested):
LGTM

russellhancox (approved):

PR #1082, opened 2023-04-27
Fix documentation for clean sync field in the preflight request.

faizanrashid:
The 'request_clean_sync' field is set here: https://github.com/google/santa/blob/main/Source/santasyncservice/SNTSyncPreflight.m#L76
The constant is defined here: https://github.com/google/santa/blob/main/Source/common/SNTSyncConstants.m#L27

### DIFF
--- a/docs/development/sync-protocol.md
+++ b/docs/development/sync-protocol.md
@@ -98,7 +98,7 @@ The request consists of the following JSON keys:
 | transitive_rule_count | NO | int | Number of transitive rules the client has at the time of sync |
 | teamid_rule_count | NO | int | Number of TeamID rules the client has at the time of sync | 24 |
 | client_mode | YES | string | the mode the client is operating in, either "LOCKDOWN" or "MONITOR" | LOCKDOWN |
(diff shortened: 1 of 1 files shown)


docs/development/sync-protocol.md
```diff
@@ -98,7 +98,7 @@ The request consists of the following JSON keys:
 | transitive_rule_count | NO | int | Number of transitive rules the client has at the time of sync |
 | teamid_rule_count | NO | int | Number of TeamID rules the client has at the time of sync | 24 |
 | client_mode | YES | string | the mode the client is operating in, either "LOCKDOWN" or "MONITOR" | LOCKDOWN |
-| clean_sync | NO | bool | the client has requested that a clean sync of its rules from the server. | true |
+| request_clean_sync | NO | bool | the client has requested that a clean sync of its rules from the server. | true |
```
russellhancox, line 101: /
```suggestion
| request_clean_sync | NO | bool | the client has requested a clean sync of its rules from the server. | true |
```

russellhancox (approved):

PR #1106, opened 2023-06-01
Basic dialog functionality when access to a watch item is denied

mlw:
Note: This is a work in progress. PR is being cut in the current state to help keep the review smaller. All policies are hard coded to be silent and nothing will be displayed for now. But, if the `silent` member of `WatchItemPolicy` were changed to `false` then a dialog would be displayed as expected.


Source/gui/SNTFileAccessMessageWindowController.m
```diff
@@ -0,0 +1,77 @@
...
+  [super windowWillClose:notification];
+}
+
+- (NSAttributedString *)attributedCustomMessage {
+  return [SNTBlockMessage formatMessage:self.customMessage];
+}
+
+- (NSString *)messageHash {
+  return [NSString stringWithFormat:@"%@|%@|%d", self.event.ruleName, self.event.ruleVersion,
+                                    [self.event.pid intValue]];
```
pmarkowsky, line 74: Does this overfit to a specific process shouldn't that be name or another ID that's more stable?

Source/gui/SNTFileAccessMessageWindowView.swift
```diff
@@ -0,0 +1,156 @@
...
+    }.frame(maxWidth:800.0).fixedSize()
+  }
+
+  func publisherInfo() {
+    print("showing publisher popup...")
+  }
+
+  func openButton() {
+    print("opening event info...")
+  }
```
pmarkowsky, line 122: This probably needs the URL extension to the sync protocol to redirect uses to the right place.

Source/gui/SNTFileAccessMessageWindowView.swift
```diff
@@ -0,0 +1,156 @@
...
+
+      Spacer()
+
+    }.frame(maxWidth:800.0).fixedSize()
+  }
+
+  func publisherInfo() {
+    print("showing publisher popup...")
+  }
+
```
pmarkowsky, line 119: Please add a TODO to wire up cert info from the process.

Source/gui/SNTNotificationManager.h
```diff
@@ -17,6 +17,7 @@
 #import "Source/common/SNTXPCNotifierInterface.h"
 #import "Source/gui/SNTBinaryMessageWindowController.h"
 #import "Source/gui/SNTDeviceMessageWindowController.h"
+#import "Source/gui/SNTFileAccessMessageWindowController.h"
```
pmarkowsky, line 20: This include probably don't need this in the header just the `.m`

Source/santad/EventProviders/SNTEndpointSecurityFileAccessAuthorizer.mm
```diff
@@ -481,6 +485,27 @@ - (FileAccessPolicyDecision)handleMessage:(const Message &)msg
                               targetPathCopy, policyDecision);
                           }];
     }
+
+    if (!optionalPolicy.value()->silent && self.fileAccessBlockCallback) {
```
pmarkowsky, line 489: nit: This will always be false.

pmarkowsky (approved):
Overall looks good. I think this is ok to approve as we know there more PRs to follow and this leaves the features inert and keeps the size of the PR more manageable.

Source/gui/SNTFileAccessMessageWindowView.swift
mlw: Added TODO

Source/santad/EventProviders/SNTEndpointSecurityFileAccessAuthorizer.mm
mlw: As discussed, on purpose for now as noted in `WatchItemPolicy.h`

Source/gui/SNTFileAccessMessageWindowController.m
pmarkowsky: I opened #1109 to track this.

pmarkowsky (approved):
Reapproving.

PR #1107, opened 2023-06-08
Migrate to new SNTRuleType enum values

mlw:
This change migrates the the SNTRuleType enum to have new values. They are now in order of precedence and allow room to grow for future rule types.

This requires a database version bump.

### DIFF
--- a/Source/common/SNTCommonEnums.h
+++ b/Source/common/SNTCommonEnums.h
@@ -38,11 +38,11 @@ typedef NS_ENUM(NSInteger, SNTAction) {
 
 // Supported Rule Types
 //
-// IMPORTANT: These enum values should be in order of decreasing precedence as
-// evaluated by Santa. That is, the lowest enum value has the highest precedence.
-// When adding new rule types, be sure to leave appropriate space between
-// surrounding values so that additional rules can more easily be added in the
-// future.
+// Note: These enum values should be in order of decreasing precedence as
+// evaluated by Santa. When adding new enum values, leave some space so that
+// additional rules can be added without violating this. The ordering isn't
+// strictly necessary but improves readability and may preemptively prevent
+// issues should SQLite behavior change.
 typedef NS_ENUM(NSInteger, SNTRuleType) {
   SNTRuleTypeUnknown = 0,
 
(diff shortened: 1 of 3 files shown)


Source/santad/DataLayer/SNTRuleTable.m
```diff
@@ -210,12 +210,23 @@ - (uint32_t)initializeDatabase:(FMDatabase *)db fromVersion:(uint32_t)version {
...
     [db executeUpdate:@"ALTER TABLE 'rules' RENAME COLUMN 'shasum' TO 'identifier'"];
     newVersion = 4;
   }
 
+  if (version < 5) {
+    // Migrate SNTRuleType enum values
+    [db executeUpdate:@"UPDATE rules SET type = 1000 WHERE type = 1"];
+    [db executeUpdate:@"UPDATE rules SET type = 3000 WHERE type = 2"];
+    [db executeUpdate:@"UPDATE rules SET type = 4000 WHERE type = 3"];
+    [db executeUpdate:@"UPDATE rules SET type = 2000 WHERE type = 4"];
```
russellhancox, line 225: Add a comment here that the re-ordering is intentional

Source/common/SNTCommonEnums.h
```diff
@@ -36,13 +36,20 @@ typedef NS_ENUM(NSInteger, SNTAction) {
 #define RESPONSE_VALID(x) \
   (x == SNTActionRespondAllow || x == SNTActionRespondDeny || x == SNTActionRespondAllowCompiler)
 
+// Supported Rule Types
+//
+// IMPORTANT: These enum values should be in order of decreasing precedence as
```
russellhancox, line 41: Why?

pmarkowsky: Agree with @russellhancox here not sure why this matters since we're always taking the first rule we get out of the sqlite database so the identifiers could be anything as long as the order in the query is right. This seems like an unnecessary callout.

mlw: The goal of the PR is to protect against what is (I thought) an undocumented side effect of sqlite evaluation, described here: https://github.com/google/santa/blob/main/Source/santad/DataLayer/SNTRuleTable.m#L336

I would argue that even if this is a documented and guaranteed feature of sqlite, we are still better served by having the rule types be in order of precedence as it makes things more readable.

Definitely happy to comment.

PR #1123, opened 2023-07-06
Add kSyncEnableCleanSyncEventUpload to the _forcedConfigKeyTypes dict

### LINKED ISSUE #1101: After turning on the clean_sync switch, the event logs are no longer being uploaded
Hey:
In the profile, I set _EnableAllEventUpload_ and _SyncEnableCleanSyncEventUpload_ to **true**. 
Additionally, I sent the _clean_sync_ field to the santa client through the sync server and set it as **true**. 
However, I noticed that aside from block event logs that occasionally upload, the event logs of Allow_XXX type are not being uploaded. 
My understanding was that all logs should be uploaded like they were when _clean_sync_ was set to **false**. Unfortunately, since I am not familiar with OC language, I was unable to comprehend the logic by reading the source code. Did I misunderstand something?

my santa client version: 
```
santad          | 2023.1 (build 503475832)
santactl        | 2023.1 (build 503475832)
SantaGUI        | 2023.1 (build 503475832)
```

pmarkowsky:
Fixes #1101.

### DIFF
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -315,7 +315,7 @@ static NSString *const kSyncCleanRequired = @"SyncCleanRequired";
   return [self configStateSet];
 }
 
-+ (NSSet *)keyPathsForValuesAffectingSyncEnableCleanSyncEventUpload {
++ (NSSet *)keyPathsForValuesAffectingEnableCleanSyncEventUpload {
   return [self configStateSet];
 }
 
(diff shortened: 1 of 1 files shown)


Source/common/SNTConfigurator.m
```diff
@@ -314,6 +315,10 @@ + (NSSet *)keyPathsForValuesAffectingSyncBaseURL {
   return [self configStateSet];
 }
 
++ (NSSet *)keyPathsForValuesAffectingSyncEnableCleanSyncEventUpload {
```
russellhancox, line 318:
```suggestion
+ (NSSet *)keyPathsForValuesAffectingEnableCleanSyncEventUpload {
```

russellhancox (approved):

mlw (approved):

PR #1128, opened 2023-07-25
Add more file access config options

mlw:
IMPORTANT: This PR is only a piece of the overall work to begin supporting user-facing messages when file access operations are denied in order to provide a better experience to users who violate policies. No actual user facing dialogs are presented yet, all deny operations are silent.

This PR adds support for 3 new keys for file access configurations:

1. `EnableSilentMode`: When true, will not show dialogs to the user when the rule is violated (mimics `EnableSilentMode` of the main Santa daemon configuration)
2. `EnableSilentTTYMode`: When true, will not print messages to the same TTY as that attached to the executing process  when the rule is violated (mimics `EnableSilentTTYMode` of the main Santa daemon configuration)
3. `BlockMessage`: When set, will control a custom message that will be displayed to the user (both GUI and TTY) when the rule is violated (mimics `BannedBlockMessage` of the main Santa daemon configuration)


Source/santad/EventProviders/SNTEndpointSecurityFileAccessAuthorizer.mm
```diff
@@ -485,7 +485,7 @@ - (FileAccessPolicyDecision)handleMessage:(const Message &)msg
                               targetPathCopy, policyDecision);
                           }];
     }
-
+#if 0
```
russellhancox, line 488: Is this intentional?

mlw: Yes - since this is now configurable, i didn't want to actually pop open dialogs yet. So keeping it hardcoded as disabled.

russellhancox (approved):

PR #1132, opened 2023-07-26
Enforce expected case for various rule type identifiers

### LINKED ISSUE #1131: Certificate SHA256 hashes are case-sensitive
I copy/pasted a certificate SHA256 hash from a Santa block popup message, which display in all caps like "EB 34 84 AD...", removed the spaces (so it became "EB3484AD...") and dropped it into my sync server to push down to devices.  The rule was added to the database but the application was still blocked.  I then used the SHA256 from "santactl fileinfo" which is all lowercase (so "eb3484ad...") and this worked successfully.

As these are hexadecimal I don't believe case should matter right?  As the Santa client displays both upper and lower case it should probably take a case-insensitive approach to rule matching as well.

mlw:
This PR enforces the following:
1. Binary and Certificate rule identifiers (sha 256 hex strings) will be converted to lowercase
2. TeamID rule identifiers (10 character alphanumeric strings) will be converted to uppercase
3. SigningID rule identifiers will have the TeamID portion of the rule be converted to uppercase, but the Signing ID portion will be left alone (e.g. `tid:Sid` --> `TID:Sid`)
    * However, if the hard coded TeamID "platform" is used, it will not have its case converted

fixes #1131

### DIFF
--- a/Source/common/SNTRule.m
+++ b/Source/common/SNTRule.m
@@ -59,9 +59,10 @@ static const NSUInteger kExpectedTeamIDLength = 10;
 
     } else if (type == SNTRuleTypeSigningID) {
       // SigningID rules are a combination of `TeamID:SigningID`. The TeamID should
-      // be forced to be uppercase, but very loose rules existing for SigningIDs, the
-      // case will be kept as-is. As an exception, platform binaries are expected to
-      // have the hardcoded string "platform" as the team ID.
+      // be forced to be uppercase, but because very loose rules exist for SigningIDs,
+      // their case will be kept as-is. However, platform binaries are expected to
+      // have the hardcoded string "platform" as the team ID and the case will be left
+      // as is.
       NSArray *sidComponents = [identifier componentsSeparatedByString:@":"];
       if (!sidComponents || sidComponents.count != 2) {
         return nil;
--- a/Source/common/SNTRule.m
+++ b/Source/common/SNTRule.m
@@ -14,6 +14,9 @@
 
 #import "Source/common/SNTRule.h"
 
+#include <CommonCrypto/CommonCrypto.h>
+#include <os/base.h>
+
 #import "Source/common/SNTSyncConstants.h"
 
 // https://developer.apple.com/help/account/manage-your-team/locate-your-team-id/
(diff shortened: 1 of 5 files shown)


Source/santad/SNTDecisionCacheTest.mm
```diff
@@ -85,8 +85,8 @@ - (void)testResetTimestampForCachedDecision {
 
   OCMExpect([self.mockRuleDatabase
     resetTimestampForRule:[OCMArg checkWithBlock:^BOOL(SNTRule *rule) {
-      return rule.identifier == cd.sha256 && rule.state == SNTRuleStateAllowTransitive &&
-             rule.type == SNTRuleTypeBinary;
+      return [rule.identifier isEqualToString:cd.sha256] &&
```
mlw, line 88: this change is not related to the overall PR, however this test started failing because it was using the wrong equality comparison for ObjC objects.

Not sure how this ever passed previously ???  ¯\\\_(ツ)_/¯

pmarkowsky: That's concerning. We should probably set aside time to make sure our tests are accurate for the next fixit.

pmarkowsky (dismissed):
LGTM.

Source/common/SNTRule.m
```diff
@@ -28,20 +32,69 @@ - (instancetype)initWithIdentifier:(NSString *)identifier
...
+      [[NSCharacterSet characterSetWithCharactersInString:@"0123456789abcdef"] invertedSet];
+    NSCharacterSet *nonUppercaseAlphaNumeric = [[NSCharacterSet
+      characterSetWithCharactersInString:@"ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"] invertedSet];
+
+    if (type == SNTRuleTypeBinary || type == SNTRuleTypeCertificate) {
+      // For binary and certificate rules, force the hash identifier to be lowercase hex.
+      identifier = [identifier lowercaseString];
+
+      identifier = [identifier stringByTrimmingCharactersInSet:nonHex];
+      if (identifier.length != 64) {
```
russellhancox, line 49: Make 64 a constant too, to match team ID.

russellhancox (approved):

PR #1135, opened 2023-07-26
Change "exponential" backoff in SNTSyncStage.m to be exponential

alexgraehl:
Fixes the sync --clean "exponential" backoff to be actually exponential, instead of its current linear version (2, 4, 6, 8, 10).

Suggests an improvement to the log message to indicate that the job will ALWAYS abort after N retries. Previously, it was not clear why it would try 5 times and then give up.

This is motivated by an actual incident on July 26, 2023, where the backoff was insufficient, causing users to be unable to download updated rules.

Previously, a user would see something like this when running sudo santactl sync --clean:

```
sudo santactl sync --clean
Uploaded 1 events
Received 697 rules
Performing request, attempt 1
Received 784 rules
Performing request, attempt 1
Received 715 rules
Performing request, attempt 1
Performing request, attempt 2
Performing request, attempt 3
Performing request, attempt 4
Performing request, attempt 5  <-- is not clear to the user why this would be the last attempt
Rule download failed, aborting run
```

### DIFF
--- a/Source/santasyncservice/SNTSyncStage.m
+++ b/Source/santasyncservice/SNTSyncStage.m
@@ -115,7 +115,7 @@
       nanosleep(&ts, NULL);
     }
 
-    SLOGD(@"Performing request, attempt %d", attempt);
+    SLOGD(@"Performing request, attempt %d (of %d maximum)...", attempt, maxAttempts);
     data = [self performRequest:request timeout:timeout response:&response error:&error];
     if (response.statusCode == 200) break;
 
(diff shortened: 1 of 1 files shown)


mlw:
LGTM but I don't see the referenced change to the log message.

alexgraehl:
Whoops, you're right, I didn't copy that log change part over. Let me fix that.

alexgraehl (commented):
Included the logging change that I forgot earlier. Thanks for pointing that out!

russellhancox (approved):

Issue #1087, opened 2023-05-02
Add per-rule customized URLs


russellhancox:
## Overview

The UI that displays block notifications currently has a single "Open.." button that can open a single site with templated values from the blocked execution. For some block rules, it would be nice to be able to customize the URL to be completely different or disable the button altogether.

## Proposal

Add a `custom_url` field to the Rule type that can be populated from a sync server and will be stored in the rules database (similarly to the `custom_msg` field). When a notification is displayed:

* If the field contains a valid URL, the "Open" button will open this URL instead of the normal templated URL from the config. If the GUI config normally customizes the button's text, that customization will be ignored and the label will say `Open...`.
* If the field contains the string `null`, the button will not be shown on the dialog.
* Any other value, or a missing value for that rule will show the normal button.

mlw:
fixed by #1140

PR #1140, opened 2023-08-03
sync/UI: Add ability to send custom URLs for blocking rules.

russellhancox:
This allows a sync server to send a `custom_url` field along with a rule blocking execution and this will be used as the URL for the "open" button in place of the normally generated URL.

### DIFF
--- a/Source/common/SNTBlockMessage.m
+++ b/Source/common/SNTBlockMessage.m
@@ -109,19 +109,31 @@
   return [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
 }
 
+// Returns either the generated URL for the passed in event, or an NSURL from the passed in custom
+// URL string. If the custom URL string is the string "null", nil will be returned. If no custom
+// URL is passed and there is no configured EventDetailURL template, nil will be returned.
+// The following "format strings" will be replaced in the URL, if they are present:
+//
+//   %file_identifier%           - The SHA-256 of the binary being executed.
+//   %bundle_or_file_identifier% - The hash of the bundle containing this file or the file itself, 
+//                                 if no bundle hash is present.
+//   %username%                  - The executing user's name.
+//   %machine_id%                - The configured machine ID for this host.
+//   %hostname%                  - The machine's FQDN.
+//   %uuid%                      - The machine's UUID.
+//   %serial%                    - The machine's serial number.
+//
 + (NSURL *)eventDetailURLForEvent:(SNTStoredEvent *)event customURL:(NSString *)url {
-  if (url.length) {
-    if ([url isEqualToString:@"null"]) return nil;
-    return [NSURL URLWithString:url];
-  }
-
   SNTConfigurator *config = [SNTConfigurator configurator];
 
   NSString *hostname = [SNTSystemInfo longHostname];
   NSString *uuid = [SNTSystemInfo hardwareUUID];
   NSString *serial = [SNTSystemInfo serialNumber];
-  NSString *formatStr = config.eventDetailURL;
+
+  NSString *formatStr = url;
+  if (!url.length) formatStr = config.eventDetailURL;
   if (!formatStr.length) return nil;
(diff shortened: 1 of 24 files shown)


pmarkowsky:
This implements and fixes #1087.

Expected behavior for the `custom_url` field is as follows:

| custom_url value | Behavior |
|---|---|
| string "null" | No open button is shown in the Santa UI |
| (A valid URL as a string e.g. "http://" | Open button is shown and clicking it takes the user to the custom URL |
| otherwise the default URL is shown | Santa behave normally |

Source/common/SNTBlockMessage.m
```diff
@@ -109,7 +109,12 @@ + (NSString *)stringFromHTML:(NSString *)html {
   return [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
 }
 
-+ (NSURL *)eventDetailURLForEvent:(SNTStoredEvent *)event {
++ (NSURL *)eventDetailURLForEvent:(SNTStoredEvent *)event customURL:(NSString *)url {
```
pmarkowsky, line 112: nit: probably worth commenting on the behavior for url with "null" vs. other strings that result in an invalid URL.

Source/common/SNTBlockMessage.m
```diff
@@ -109,7 +109,12 @@ + (NSString *)stringFromHTML:(NSString *)html {
...
+  if (url.length) {
+    if ([url isEqualToString:@"null"]) return nil;
+    return [NSURL URLWithString:url];
```
pmarkowsky, line 115: If an invalid URL is passed I think we should log here.

pmarkowsky (commented):
Looks good. I just think we should add an info / warning log if the  custom url is somehow malformed since it'll make the button disappear in the dialog.

Source/santad/SNTNotificationQueue.h
```diff
@@ -21,6 +21,6 @@
 
 @property(nonatomic) MOLXPCConnection *notifierConnection;
 
-- (void)addEvent:(SNTStoredEvent *)event customMessage:(NSString *)message;
+- (void)addEvent:(SNTStoredEvent *)event customMessage:(NSString *)message url:(NSString *)url;
```
mlw, line 24: Nit: We have interfaces using `andCustomURL`, `customURL` and `url`. It'd be nice to have these be more consistent (but don't think we should block on it if you don't have time right now).

mlw (approved):

Source/common/SNTBlockMessage.m
russellhancox: Done. In the process I also extended it to allow format strings in the custom URL and documented that goo.

Source/common/SNTBlockMessage.m
russellhancox: Done

Source/santad/SNTNotificationQueue.h
russellhancox: Done, I think.

Source/common/SNTBlockMessage.m
```diff
@@ -109,7 +109,12 @@ + (NSString *)stringFromHTML:(NSString *)html {
   return [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
 }
 
-+ (NSURL *)eventDetailURLForEvent:(SNTStoredEvent *)event {
++ (NSURL *)eventDetailURLForEvent:(SNTStoredEvent *)event customURL:(NSString *)url {
+  if (url.length) {
+    if ([url isEqualToString:@"null"]) return nil;
```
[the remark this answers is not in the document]
russellhancox: You could put all the details in a custom_msg and then have no reason to take the user elsewhere, setting the `custom_url` to `null` will allow hiding the normal "Open" button.

pmarkowsky (approved):
LGTM.

mlw (approved):

PR #1144, opened 2023-08-03
sync: Add SyncExtraHeaders config option.

### LINKED ISSUE #1143: Config: Add ability to include custom headers in sync requests
#### Discussed in https://github.com/google/santa/discussions/1127

This could be used to pass static auth tokens or custom headers that a server may require.

russellhancox:
Fixes #1143

### DIFF
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -391,6 +391,27 @@
 @property(readonly, nonatomic) NSDictionary *syncProxyConfig;
 
 ///
+///  Extra headers to include in all requests made during syncing.
+///  Keys and values must all be strings, another other type will be silently ignored.
+///  Some headers cannot be set through this key, including:
+///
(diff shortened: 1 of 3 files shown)


Source/common/SNTConfigurator.h
```diff
@@ -390,6 +390,27 @@
 ///
 @property(readonly, nonatomic) NSDictionary *syncProxyConfig;
 
+///
+///  Extra headers to include in all requests made during syncing.
+///  Keys and values must all be strings, another other type will be silently ignored.
```
pmarkowsky, line 395:
```suggestion
///  Keys and values must all be strings, any other type will be silently ignored.
```

pmarkowsky (commented):
LGTM. Just have the one typo. This should make it easy to include bearer tokens if so desired from MDM.

pmarkowsky (approved):
LGTM.

PR #1145, opened 2023-08-04
Add hot cache for targets of read only policies

mlw:
This adds a new cache to the FAA client that caches target paths that are allowed to be read by policy. When future OPEN events are received, if the target isn't being opened for writing, the cache is first checked  to see if the target has previously been allowed. If so, a response is sent immediately without processing the message asynchronously or having to lookup policy.

### DIFF
--- a/Source/santad/EventProviders/SNTEndpointSecurityFileAccessAuthorizer.mm
+++ b/Source/santad/EventProviders/SNTEndpointSecurityFileAccessAuthorizer.mm
@@ -69,8 +69,6 @@ struct PairHash {
   }
 };
 
-using FileSet = std::unordered_set<std::pair<dev_t, ino_t>, PairHash>;
-
 NSString *kBadCertHash = @"BAD_CERT_HASH";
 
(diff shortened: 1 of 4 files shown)


Source/santad/EventProviders/SNTEndpointSecurityFileAccessAuthorizer.mm
```diff
@@ -59,6 +64,13 @@
 using santa::santad::event_providers::endpoint_security::Message;
 using santa::santad::logs::endpoint_security::Logger;
 
+struct PairHash {
+  template <typename T1, typename T2>
+  std::size_t operator()(const std::pair<T1, T2> &pair) const {
+    return std::hash<T1>()(pair.first) << 1 ^ std::hash<T2>()(pair.second);
```
pmarkowsky, line 70: Is it worth using [`absl::Hash`](https://abseil.io/docs/cpp/guides/hash) here instead?

pmarkowsky (commented):
Overall looks good to me. Just worth checking if we should also use absl::Hash if not then I'm happy to approve this.

pmarkowsky (dismissed):
LGTM.

pmarkowsky (approved):
LGTM.

PR #1150, opened 2023-08-14
Added TransitiveWhitelisting explanation to rules.md

p-harrison:
Added a section to explain TransitiveWhitelisting and Transitive/Compiler rules

### DIFF
--- a/docs/concepts/rules.md
+++ b/docs/concepts/rules.md
@@ -7,8 +7,7 @@ parent: Concepts
 ## Rule Types
 
 Rules provide the primary evaluation mechanism for allowing and blocking
-binaries with Santa on macOS. There are four types of rules: binary, signing ID,
-certificate, and Team ID.
+binaries with Santa on macOS.
 
 ### Binary Rules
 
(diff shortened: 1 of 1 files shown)


docs/concepts/rules.md
```diff
@@ -86,6 +85,16 @@ as a single developer account can and frequently will request/rotate between
 multiple different signing certificates and entitlements. This is an even more
 powerful rule with broader reach than individual certificate rules.
 
+### Compiler/Transitive Rules
+
+The transitive whitelist capability of Santa can automatically whitelist any files that are created by a set of specified binaries.  A typical use-case is allowing any binaries compiled with XCode on developer machines to execute, as it would be slow and impractical to use other rule types to permit these.
+
+To begin using transitive whitelisting, `EnableTransitiveWhitelisting` should be set to true and Compiler rules (rules with the policy `ALLOWLIST_COMPILER`) should be added to indicate the binaries which will be writing the new files to be whitelisted.  Santa will create and manage Transitive rules in its database automatically, they cannot be created directly.
```
mlw, line 92: Minor tweaks to use the new preferred naming and the non-deprecated configuration key:

```suggestion
To begin using transitive allowlisting, `EnableTransitiveRules` should be set to true and Compiler rules (rules with the policy `ALLOWLIST_COMPILER`) should be added to indicate the binaries which will be writing the new files to be allowlisted.  Santa will create and manage Transitive rules in its database automatically, they cannot be created directly.
```

docs/concepts/rules.md
```diff
@@ -86,6 +85,16 @@ as a single developer account can and frequently will request/rotate between
 multiple different signing certificates and entitlements. This is an even more
 powerful rule with broader reach than individual certificate rules.
 
+### Compiler/Transitive Rules
+
+The transitive whitelist capability of Santa can automatically whitelist any files that are created by a set of specified binaries.  A typical use-case is allowing any binaries compiled with XCode on developer machines to execute, as it would be slow and impractical to use other rule types to permit these.
```
mlw, line 90: Minor tweaks to use the new preferred naming:

```suggestion
The transitive allowlist capability of Santa can automatically allowlist any files that are created by a set of specified binaries.  A typical use-case is allowing any binaries compiled with XCode on developer machines to execute, as it would be slow and impractical to use other rule types to permit these.
```

mlw (changes requested):

mlw (approved):

PR #1158, opened 2023-08-21
Add SigningID/TeamID to Event definition in sync-protocol.md

p-harrison:
Document SigningID and TeamID in the definition of Events in the EventUpload stage

### DIFF
--- a/docs/development/sync-protocol.md
+++ b/docs/development/sync-protocol.md
@@ -234,11 +234,11 @@ sequenceDiagram
     ],
     "quarantine_timestamp": 0,
     "signing_chain": [{
-        "cn": "Apple Development: Google Development (XXXXXXXXXX)",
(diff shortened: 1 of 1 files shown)


docs/development/sync-protocol.md
```diff
@@ -273,7 +275,8 @@ sequenceDiagram
       "markowsky@ttys001",
       "markowsky@ttys003"
     ],
-    "team_id": "XXXXXXXXXX"
+    "team_id": "XXXXXXXXXX",
+    "signing_id": "XXXXXXXXXX:YYY"
```
pmarkowsky, line 279: We should use the Google Team ID and `com.google.Chrome` to make this a more concrete example and match the other docs.

p-harrison: Good point, team_id was already like that so I wasn't sure about messing with it.  I'll change now.

PR #1159, opened 2023-08-21
Correction to sync-protocol.md

p-harrison:
I think file_bundle_binary_count and file_bunde_hash were accidentally added to the Rule definition and examples, rather than to the Event section, so I propose moving them.


pmarkowsky (approved):
LGTM. Thanks for fixing this.

PR #1162, opened 2023-08-23
Fix new buildifier issues

mlw:
Addresses buildifier issues that suddenly started.

Example: https://github.com/google/santa/actions/runs/5953025386/job/16146336259?pr=1161


pmarkowsky (approved):
LGTM.

PR #1164, opened 2023-08-25
Restore file_bundle_hash & file_bundle_binary_count to Sync Protocol Docs

pmarkowsky:
Restore `file_bundle_hash` and `file_bundle_binary_count` to docs that were mistakenly removed in #1159.

They're used in `SNTSyncRuleDownload's https://github.com/google/santa/blob/4fdc1e5e411d14a07aa087e26e57d0841d177644/Source/santasyncservice/SNTSyncRuleDownload.m#L139


mlw (approved):

p-harrison:
@pmarkowsky sorry for leading you astray with my correction.  Do you mind me asking what these two keys are used for in a rule?  My understanding of bundle processing from the documentation is that its purpose is to scan a bundle and create Events for all binaries in that bundle, so when permitting a new bundle the administrator can 'get ahead' of the other binaries that will be blocked once the first to trigger a block is allowed.  But there's obviously a bit more going on based on that code you linked to?

pmarkowsky:
@p-harrison This was my fault for not double checking the code first. 

This is used to trigger notifications for bundles so we can send notifications to users like foo.app is unblocked.

Issue #331, opened 2019-01-04
Condsider removing LogUpload phase of santactl/sync


russellhancox:
It doesn't upload the right things and is nowhere near as useful as I thought it would be.

PR #1168, opened 2023-08-29
Remove logupload stage from syncing-overview.md

p-harrison:
The logupload stage was referred to in this document but was removed in #331.

FYI this document also refers to santactl performing syncs, which I believe is now handled by santasyncservice, but I am not familiar enough with it to document sorry.


mlw:
LGTM, thank you!

mlw (approved):

PR #1170, opened 2023-08-31
Add basic support for importing and exporting rules to/from JSON

### LINKED ISSUE #56: json 'pack' format to allow import/export of a group of rules
Similar to how osquery collects common queries into 'packs' to be enabled in their config, it would be great to be able to import/export rules from the database as an option to santactl. Bonus points would be if output from `santactl fileinfo` was in that format to start with, so you could build up a collection from the results of fingerprinting a binary or its cert.

pmarkowsky:
This PR adds support for importing and exporting rules to/from a JSON file when Santa is not configured to use a sync server.

Fixes #56.

The format of the file is:

```json
{"rules": [{ "policy" : "BLOCKLIST",
                   "rule_type" : "BINARY",
                  "identifier" : "84de9c61777ca36b13228e2446d53e966096e78db7a72c632b5c185b2ffe68a6"
                  "custom_url" : "",
                  "custom_msg" : "/bin/ls block for demo"}
               ]}
```

## Example usage of export

```shell
$ sudo santactl rules --export rules.json
```

## Example usage of import

```shell
$ sudo santactl rules --import rules.json
```


Source/santactl/Commands/SNTCommandRule.m
```diff
@@ -154,11 +175,40 @@ - (void)runWithArguments:(NSArray *)arguments {
...
+               [arg caseInsensitiveCompare:@"-h"] == NSOrderedSame) {
+      printf("%s\n", self.class.longHelpText.UTF8String);
+      exit(0);
     } else {
       [self printErrorUsageAndExit:[@"Unknown argument: " stringByAppendingString:arg]];
     }
   }
 
+  if (![jsonFilePath isEqualToString:@""]) {
+    if (importRules) {
```
mlw, line 205: This is implemented to only allow either an import or export operation, not both simultaneously. That's fine, but we should probably print some sort of error when parsing flags above.

Source/santactl/Commands/SNTCommandRule.m
```diff
@@ -154,11 +175,40 @@ - (void)runWithArguments:(NSArray *)arguments {
...
+      [self importJSONFile:jsonFilePath];
+    } else if (exportRules) {
+      [self exportJSONFile:jsonFilePath];
+    }
+  }
```
mlw, line 210: Should we just return at the end of this if-block? Would likely lead to weird operations that might occur if different flags are mixed

Source/santad/SNTDaemonControlController.mm
```diff
@@ -156,6 +156,35 @@ - (void)staticRuleCount:(void (^)(int64_t count))reply {
...
+  for (SNTRule *rule in rules) {
+    // Omit transitive and remove rules as they're not relevan.
+    if (rule.state == SNTRuleStateAllowTransitive || rule.state == SNTRuleStateRemove) {
+      continue;
+    }
+
+    [rulesAsDicts addObject:[rule dictionaryRepresentation]];
+  }
+
+  reply(rulesAsDicts, nil);
```
mlw, line 185: It might be interesting to consider switching things up here a bit - instead of creating the dictionary in the daemon, reply with the array of rules and do the construction within `santactl`. `SNTRule` already conforms to `NSSecureCoding` so it can be sent over XPC.

This would have the benefit of offloading work out of the daemon and also make this retrieval more versatile for future operations.

pmarkowsky: I agree. I'll make the change.

Source/santactl/Commands/SNTCommandRule.m
```diff
@@ -81,13 +84,28 @@ + (NSString *)longHelpText {
...
+          @"                       \"custom_msg\": \"/bin/ls block for demo\"}\n"
+          @"                      ]}\n");
 }
 
 - (void)runWithArguments:(NSArray *)arguments {
   SNTConfigurator *config = [SNTConfigurator configurator];
-  if ((config.syncBaseURL || config.staticRules.count) &&
-      ![arguments containsObject:@"--check"]
+  if ((config.syncBaseURL || config.staticRules.count) && ![arguments containsObject:@"--check"] &&
+      ![arguments containsObject:@"--import"] &&
```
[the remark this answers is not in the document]
pmarkowsky: My thought here was that we should just disable any local rule management import / export if a sync server is being used.

I'm not opposed to making export work. As it'd let you bootstrap a standalone instance from one with a sync server.

Source/santactl/Commands/SNTCommandRule.m
```diff
@@ -154,11 +175,44 @@ - (void)runWithArguments:(NSArray *)arguments {
     } else if ([arg caseInsensitiveCompare:@"--force"] == NSOrderedSame) {
       // Don't do anything special.
 #endif
+    } else if ([arg caseInsensitiveCompare:@"--json"] == NSOrderedSame) {
+      if (++i > arguments.count - 1) {
+        [self printErrorUsageAndExit:@"--json requires an argument"];
+      }
+      jsonFilePath = arguments[i];
+    } else if ([arg caseInsensitiveCompare:@"--import"] == NSOrderedSame) {
+      importRules = YES;
```
mlw, line 184: Add a similar check for `exportRules` here

Source/santactl/Commands/SNTCommandRule.m
```diff
@@ -404,40 +408,48 @@ - (void)importJSONFile:(NSString *)jsonFilePath {
...
-                                                         options:NSJSONWritingPrettyPrinted
-                                                           error:&error];
-      // Print error
-      if (error) {
-        printf("Failed to jsonify rules: %s", [error.localizedDescription UTF8String]);
-        LOGD(@"Failure reason: %@", error.localizedFailureReason);
-        exit(1);
+    }
+
+    if (rules.count == 0 || !rules) {
```
mlw, line 418: Nit: Simplify to just `rules.count == 0`, can remove `!rules`

Source/santactl/Commands/SNTCommandRule.m
pmarkowsky: Errors messages added.

Source/santactl/Commands/SNTCommandRule.m
pmarkowsky: Done.

Source/santactl/Commands/SNTCommandRule.m
pmarkowsky: Return added.

Source/santad/SNTDaemonControlController.mm
pmarkowsky: Change made. We're now converting the rules to json in `santactl`.

Source/santactl/Commands/SNTCommandRule.m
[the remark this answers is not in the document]
pmarkowsky: Should be fixed now.

mlw (approved):

PR #1172, opened 2023-09-01
Update rules.md with more detail on Transitive/Compiler rules

p-harrison:
Updated the description of Transitive/Compiler rules to clarify that only rules of type BINARY are allowed.


pmarkowsky (approved):
LGTM. Thanks for improving our docs.

pmarkowsky:
Work related to #585.

PR #1185, opened 2023-09-14
Use 'set -xo pipefail' instead for lint.sh

tnek:
`set -xo pipefail` implicitly does this behavior; the script will continue but without masking errors from individual commands.


pmarkowsky (approved):
LGTM. Thanks!

PR #1187, opened 2023-09-18
Update sync-protocol.md

p-harrison:
Fields like pid, ppid, execution_time, current_sessions etc. are not supplied in Event uploads when the decision is BUNDLE_BINARY (ie. Events generated by the bundle scanning service, rather than actual executions) so I have marked these as not required in the API definition. Few other small formatting tidy-ups while I was there.


pmarkowsky (approved):
LGTM.

PR #1191, opened 2023-09-27
Only eval TID and SID rules when the binary signature is valid

mlw:
Team ID and Signing ID rules should only be evaluated when the binary's signature is valid. This change ensures TID/SID values are not used when looking up rules if code signature checks fail.

Additionally this updates the test `rules.db` to the latest schema and completely revamped the SantadRules to better ensure the stated purpose of each test is actually met.


pmarkowsky (approved):
LGTM.

PR #1192, opened 2023-09-27
Remove defunct reference to a screenshot/video in the README.md

u269c:
Cleaning things up.


pmarkowsky:
Thanks for spotting this. We found the video and have #1193 which puts it back in the readme.

PR #1196, opened 2023-09-29
santactl/sync: Drop root requirement

### LINKED ISSUE #1195: Options for speeding up the sync of new rules
Hey guys,

We have Santa configured to sync once an hour with our sync server.  This is really too frequent for a ruleset that only changes every few days, but not fast enough for a user who needs access to something urgently :).  So I wonder if I can ask two questions -

1. I know there was some discussion about making `santactl sync` available to non-admin users.  I wonder if it is something that could be considered again?  It would mostly solve our issue without adding too much complexity (for me, I'm not sure what it would mean for you :) ).  If there are concerns about it being abused, perhaps a config key could enable/disable the ability for users to force a sync.

2. I can see a couple of config keys for Push Notifications and more in the code, but no documentation.  Does this mean Push Notifications are only available in Google's deployment of Santa or is it something others can use?  It would be more complex to setup, but would be superior to users performing syncs themselves, and I like the idea of pop-ups to inform users when something has been added to the allowlist.

Cheers,
Philip

russellhancox:
Previously the sync command required root in order to establish a connection to santad with enough privilege to use the XPC methods for adding rules. Now that santasyncservice exists this requirement is no longer necessary and there is no risk in allowing unprivileged users to initiate a sync.

We still ensure that privileges are dropped, just in case someone does execute as root.

Fixes #1195

### DIFF
--- a/Source/santactl/Commands/SNTCommandSync.m
+++ b/Source/santactl/Commands/SNTCommandSync.m
@@ -32,7 +32,7 @@ REGISTER_COMMAND_NAME(@"sync")
 #pragma mark SNTCommand protocol methods
 
 + (BOOL)requiresRoot {
-  return YES;
+  return NO;
 }
 
 + (BOOL)requiresDaemonConn {


pmarkowsky (approved):
LGTM.

Source/santactl/Commands/SNTCommandSync.m
```diff
@@ -32,7 +32,7 @@ @implementation SNTCommandSync
 #pragma mark SNTCommand protocol methods
 
 + (BOOL)requiresRoot {
-  return YES;
+  return NO;
```
mlw, line 35: Is there any reason not to remove the call to `DropRootPrivileges` below? We don't do this for other commands that don't require root

russellhancox: Some commands have a reason to run as root even if it's not required (e.g. fileinfo) whereas there is no benefit to running this as root and due to the XPC work it does I think it's safer if it drops privileges before it makes any connections.

mlw: Missed this blurb in the description:
```
We still ensure that privileges are dropped, just in case someone does execute as root.
```
But i guess i'm not sure the harm?

mlw: Fair enough. We may want to consider doing this on additional commands as well then (e.g. `status`, maybe `version`)

PR #1199, opened 2023-10-12
Fix issue preventing rule import / export from working

pmarkowsky:
This fixes an issue with the order of argument checking. That's blocking the use of the `--import` / `--export`.

It also adds a check to make sure that import and export only run by themselves.


kallsyms (commented):
Can you update the help string above removing the `--json` help (L67) and noting `--import` and `--export` take a path arg?

kallsyms (approved):

PR #1204, opened 2023-10-17
Update configuration.md that push notifications not widely available

p-harrison:
Update the configuration.md document to note that FCM based push notifications are not currently available outside the internal Google deployment of Santa


pmarkowsky (approved):
LGTM. 

Thank you for doing this.

PR #1205, opened 2023-10-17
Update syncing-overview.md with note on push notifications

p-harrison:
Update the syncing-overview.md document to note that FCM based push notifications are not currently available outside the internal Google deployment of Santa


pmarkowsky (approved):
LGTM.

Thanks for doing this.

PR #1209, opened 2023-10-27
Enable e2e testing on macOS 14

kallsyms:


Testing/integration/VM/bash_control.sh
```diff
@@ -3,5 +3,5 @@
 # through applescript.
 # It's run as part of the template VM creation process.
 
-osascript -e 'tell application "System Preferences" to activate'
-osascript -e 'tell application "System Events" to tell process "System Preferences" to click menu item "Profiles" of menu 1 of menu bar item "View" of menu bar 1'
+osascript -e 'tell application "System Settings" to activate'
```
pmarkowsky, line 6: I think `System Preferences` should be an env var so we're not editing this everywhere.

Testing/integration/test_sync_changes.sh
```diff
@@ -25,6 +25,11 @@ if [[ "$(sudo santactl status --json | jq .daemon.block_usb)" != "false" ]]; the
   exit 1
 fi
 
+# Wait for the UI to have come up
+sleep 5
```
pmarkowsky, line 29: Can we wait for the distributed notification instead of sleeping?

I believe you can use `notifyutil(1)`   to wait for the notification.

pmarkowsky (commented):
Overall this is great a few minor things.

Testing/integration/test_sync_changes.sh
pmarkowsky: Sorry I got it confused with [dndump](https://github.com/nktzbkv/dndump/tree/main)

.github/workflows/e2e.yml
```diff
@@ -1,25 +1,28 @@
...
+on:
+  schedule:
+    - cron: '0 4 * * *' # Every day at 4:00 UTC (not to interfere with fuzzing)
+  workflow_dispatch:
 
 jobs:
   start_vm:
     runs-on: e2e-host
     steps:
       - uses: actions/checkout@v3
```
mlw, line 12: We should use `ratchet` on this action similar to `ci.yml`.

Testing/integration/test_sync_changes.sh
kallsyms: We could, however there's still some delay between the notification being received and the UI coming up, and since the UI is the bit we actually care about here, we're back to square one.

mlw (approved):

.github/workflows/e2e.yml
```diff
@@ -1,25 +1,28 @@
...
       - name: Start VM
-        run: python3 Testing/integration/actions/start_vm.py macOS_12.bundle.tar.gz
+        run: python3 Testing/integration/actions/start_vm.py macOS_14.bundle.tar.gz
```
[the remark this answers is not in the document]
pmarkowsky: Sorry I meant for python3.

Testing/integration/VM/bash_control.sh
pmarkowsky: This was a nit.

Testing/integration/test_sync_changes.sh
pmarkowsky: Yeah I guess the thought would be that you could wait for the block then only sleep for 1 second as you know the dialog will pop up.

We can look into this later.

.github/workflows/e2e.yml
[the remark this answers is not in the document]
pmarkowsky: This is a nit so we can get it later.

PR #1220, opened 2023-11-03
Fix remount issue for APFS formatted drives

mlw:
Fixes issue where APFS drives were not being properly mounted due to unsupported options being sent to `mount_apfs`.

Credit to @kallsyms for finding the issue

### DIFF
--- a/Source/santad/EventProviders/SNTEndpointSecurityDeviceManager.mm
+++ b/Source/santad/EventProviders/SNTEndpointSecurityDeviceManager.mm
@@ -214,7 +214,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (uint32_t)updatedMountFlags:(struct statfs*)sfs {
   uint32_t mask = sfs->f_flags | mountArgsToMask(self.remountArgs);
 
-  // NB: APFS mounts are get MNT_JOURNALED implicitly set. However, mount_apfs
+  // NB: APFS mounts get MNT_JOURNALED implicitly set. However, mount_apfs
   // does not support the `-j` option so this flag needs to be cleared.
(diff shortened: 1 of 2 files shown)


Source/santad/EventProviders/SNTEndpointSecurityDeviceManager.mm
```diff
@@ -211,6 +211,18 @@ - (instancetype)initWithESAPI:(std::shared_ptr<EndpointSecurityAPI>)esApi
   return self;
 }
 
+- (uint32_t)updatedMountFlags:(struct statfs*)sfs {
+  uint32_t mask = sfs->f_flags | mountArgsToMask(self.remountArgs);
+
+  // NB: APFS mounts get MNT_JOURNALED implicitly set. However, mount_apfs
+  // does not support the `-j` option so this flag needs to be cleared.
+  if (strncmp(sfs->f_fstypename, "apfs", sizeof(sfs->f_fstypename)) == 0) {
```
pmarkowsky, line 219: nit: Couldn't this just be `(strncmp(sfs->f_fstypename, "apfs", 4) == 0)`?

pmarkowsky (approved):
LGTM.

Source/santad/EventProviders/SNTEndpointSecurityDeviceManager.mm
mlw: eh, six of one... Technically would want 5 to include the null terminator. Given `strncmp` will stop at the first null terminator of either string it doesn't really matter. Doing something like `strncmp(sfs->f_fstypename, "apfs", sizeof("apfs"))` just looks weird to me.

PR #1222, opened 2023-11-03
Only remount on startup if remount args are set

mlw:
Changes behavior of #1216 (support remount on startup) to only remount if the remount args are set. The makes for consistent behavior of the runtime mount protections which will leave a device unmounted if the remount args aren't set.


pmarkowsky (approved):
LGTM.

PR #1224, opened 2023-11-07
Instruction to Remove Adhoc Build

eopeter:
This adds instruction to the development/building docs on how to remove the adhoc build when done with development.

### DIFF
--- a/docs/development/building.md
+++ b/docs/development/building.md
@@ -163,3 +163,10 @@ infrastructure.
 ```sh
 bazel build --apple_generate_dsym -c opt //:release
 ```
+
+### Remove Ad-Hoc Build
+if you would like to remove the ad-hoc build of Santa, you can run the following command:
+
+```sh
+systemextensionsctl uninstall - com.google.santa.daemon
+```


docs/development/building.md
```diff
@@ -163,3 +163,10 @@ infrastructure.
 ```sh
 bazel build --apple_generate_dsym -c opt //:release
 ```
+
+### Remove Ad-Hoc Build
+if you would like to remove the ad-hoc build of Santa, you can run the following command:
+
+```sh
+systemextensionsctl uninstall - com.google.santa.daemon
```
russellhancox, line 171: This only works with SIP disabled. Removing /Applications/Santa.app from Finder will uninstall the extension and most components, or the `Conf/uninstall.sh` script from the source tree will remove all components

eopeter: Is SIP Disabled not a requirement for adhoc?

russellhancox: Not necessarily: development provisioning profiles limited to specific machines can be used without SIP disabled.

PR #1225, opened 2023-11-07
Add support for logging entitlements in EXEC events

mlw:
Adds a new repeated KV message to the `Execution` message in `santa.proto`.

This will be followed up by a second PR that will add various configuration/filter options.


Source/common/santa.proto
```diff
@@ -213,6 +213,11 @@ message CertificateInfo {
   optional string common_name = 2;
 }
 
+message Entitlement {
+	string key = 1;
```
kallsyms, line 217: super nit: inconsistent tabs/spaces

Source/santad/Logs/EndpointSecurity/Serializers/Protobuf.mm
```diff
@@ -449,6 +451,89 @@ static inline void EncodeCertificateInfo(::pbv1::CertificateInfo *pb_cert_info,
...
+    }
+
+    NSError *err;
+    NSData *jsonData;
+    @try {
+      id val = obj;
+
+      // Fixup some types with data that can be better represented in JSON
+      if ([obj isKindOfClass:[NSDate class]]) {
+        // Example format output: "November 6, 2023 at 10:25:20 AM EST"
```
kallsyms, line 481: Would an ISO 8601-ish format be easier for others to consume?

Source/santad/Logs/EndpointSecurity/Serializers/Protobuf.mm
```diff
@@ -449,6 +451,89 @@ static inline void EncodeCertificateInfo(::pbv1::CertificateInfo *pb_cert_info,
...
+      }
+
+      if (!jsonData) {
+        @try {
+          // As a final fallback, simply serialize an error message so that the
+          // entitlement key is still logged.
+          jsonData = [NSJSONSerialization dataWithJSONObject:@"JSON Serialization Failed"
+                                                     options:NSJSONWritingFragmentsAllowed
+                                                       error:&err];
+        } @catch (NSException *e) {
```
kallsyms, line 516: Does this even need to be wrapped in a try/catch since the serialization is on a constant (valid) object?

Source/santad/testdata/protobuf/v1/exec.json
```diff
@@ -120,5 +120,43 @@
...
+   "key": "key_with_arr_val",
+   "value": "[\"v1\",\"v2\",\"v3\"]"
+  },
+  {
+   "key": "key_with_data_val",
+   "value": "\"SGVsbG8gV29ybGQ=\""
+  },
+  {
+   "key": "key_with_arr_val_multitype",
+   "value": "\"(\\n    v1,\\n    v2,\\n    v3,\\n    123,\\n    \\\"2023-11-07 17:00:02 +0000\\\"\\n)\""
```
kallsyms, line 135: Sorta related to date formatting q above: this is a different format than the `key_with_date_val` below. I know nested values aren't common so it's not the primary concern, but seeing different formats is confusing...

Source/santad/Logs/EndpointSecurity/Serializers/Protobuf.mm
mlw: Probably not? Under real circumstances I've yet to be able to even find a set of conditions that this code is even reached (the string returned from the `description` method above should be good enough). Since it's kinda an unknown, I just left it to be safe since it doesn't really hurt anything.

Source/santad/Logs/EndpointSecurity/Serializers/Protobuf.mm
pmarkowsky: Most definitely.

mlw: Spoke offline. There is some complications here due to fallback JSON serializations would be a different date encoding due to how the `description` method works.

Will instead pre-process the dictionary by recursively walking the structure and encoding things like NSDate and NSData into desired formats first so as not to have to worry about how the `description` method is implemented.

Source/santad/testdata/protobuf/v1/exec.json
mlw: See other comment - will make things consistent by first standardizing types in nested objects.

pmarkowsky (approved):
LGTM.

Source/common/santa.proto
```diff
@@ -213,6 +213,11 @@ message CertificateInfo {
   optional string common_name = 2;
 }
 
+message Entitlement {
```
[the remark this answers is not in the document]
mlw: Will put into the next PR!

Issue #1212, opened 2023-10-30
Bundle Scanning Does Not Start When EnableSilentMode = YES


eopeter:
It seems Bundle Scanning happens only when `showQueuedWindow` calls `hashBundleBinariesForEvent` in SNTNotificationManager.m but  its never called when EnableSilentMode is true. So in SilentMode, the bundle scanning service never starts. Is this a bug?

```objc
- (void)queueMessage:(SNTMessageWindowController *)pendingMsg {
  // Post a distributed notification, regardless of queue state.
  [self postDistributedNotification:pendingMsg];

  // If GUI is in silent mode or if there's already a notification queued for
  // this message, don't do anything else.
  if ([SNTConfigurator configurator].enableSilentMode) return;
  if ([self notificationAlreadyQueued:pendingMsg]) return;

  .
  .
  .

  pendingMsg.delegate = self;
  [self.pendingNotifications addObject:pendingMsg];

  if (!self.currentWindowController) {
    [self showQueuedWindow];
  }
}
```

russellhancox:
That's working as intended. Bundle scanning is a potentially intensive process and allowing it to run unchecked could cause severe performance issues, so we tie the scanning process to the UI notification such that if the notification never appears or the user dismisses it with no action the bundle scanning never starts or is interrupted.

eopeter:
Can I independently start the Bundle Scanning service when I receive the distributed notification? If so, how would I start it with the right parameters for the current blocked bundle?

russellhancox:
No, that's not possible right now. 

It sounds like a good feature req., if EnableSilentMode is on we can have santabundleservice listen for a distributed notification and have it publish notifications too. So the flow would look something like this:

```mermaid
sequenceDiagram
    santad->>+Distributed Listener: Execution blocked
    Distributed Listener->>+santabundleservice: Hash bundle with path
    santabundleservice->>+Distributed Listener: Bundle hash progress: 10%
    santabundleservice->>+Distributed Listener: Bundle hash progress: 20%
    santabundleservice->>+Distributed Listener: Bundle hash complete
```

Some very rough steps for implementation:

1) Move the existing code in `SantaGUI` that forwards the results from `santabundleservice` to `santad`. Once `santabundleservice` finishes scanning a bundle it should send the results to `santad` directly. Not only will this stop wasting time doing 2 transfers, it also means we can stop sending all the sub-event info to `SantaGUI`
[...]

PR #1227, opened 2023-11-07
Add Bundle Scanning with EnableSilentMode

eopeter:
This PR seeks to enable Bundle Scanning with Distributed Notification when EnableSilentMode is enabled. In this mode, if a listener is listening to the distributed notification channels `com.google.santa.notification.blockedeexecution.bundlehash.progress` or `com.google.santa.notification.blockedeexecution.bundlehash`, they will be notified of the progress of Bundle Scanning and the hashes produced during the scan on the channels respectively.

This will enable usage where Santa Notification that does not use the Notification Window to be able to still listen for Bundle Scannig Events.


Source/gui/SNTNotificationManager.m
```diff
@@ -100,7 +100,20 @@ - (void)queueMessage:(SNTMessageWindowController *)pendingMsg {
...
   // If GUI is in silent mode or if there's already a notification queued for
   // this message, don't do anything else.
-  if ([SNTConfigurator configurator].enableSilentMode) return;
+  if ([SNTConfigurator configurator].enableSilentMode) {
+    // hash bundle with path
+    dispatch_async(self.hashBundleBinariesQueue, ^{
+          SNTBinaryMessageWindowController *wc = (SNTBinaryMessageWindowController *)pendingMsg;
+          [self hashBundleBinariesForEvent:wc.event withProgressHandler:^(NSUInteger progressCount) {
+              // progress is reported to the distributed notification
+              [self postDistributedNotification: pendingMsg withProgress: progressCount];
```
russellhancox, line 109: Remove the spaces after the colons in method calls

Source/gui/SNTNotificationManager.m
```diff
@@ -100,7 +100,20 @@ - (void)queueMessage:(SNTMessageWindowController *)pendingMsg {
 
   // If GUI is in silent mode or if there's already a notification queued for
   // this message, don't do anything else.
-  if ([SNTConfigurator configurator].enableSilentMode) return;
+  if ([SNTConfigurator configurator].enableSilentMode) {
+    // hash bundle with path
+    dispatch_async(self.hashBundleBinariesQueue, ^{
+          SNTBinaryMessageWindowController *wc = (SNTBinaryMessageWindowController *)pendingMsg;
```
russellhancox, line 106: The indentation inside this block should be +2 spaces from the previous line (i.e. -4 spaces from where it is now).

Running `./Testing/fix.sh` from your checked out repo should fix this and other style issues automatically.

Source/gui/SNTNotificationManager.m
```diff
@@ -129,6 +142,14 @@ - (void)queueMessage:(SNTMessageWindowController *)pendingMsg {
   }
 }
 
+- (void)postDistributedNotificationWithUserInfo:(NSDictionary *)userInfo notificationName:(NSString *)notificationName {
```
russellhancox, line 145: We enforce a per-line character limit of 100 characters. Break this line before `notificationName:` and then line up the colons of each parameter.

Source/gui/SNTNotificationManager.m
```diff
@@ -170,10 +190,38 @@ - (void)postDistributedNotification:(SNTMessageWindowController *)pendingMsg {
     kSigningChain : signingChain,
   };
 
-  [dc postNotificationName:@"com.google.santa.notification.blockedeexecution"
-                    object:@"com.google.santa"
-                  userInfo:userInfo
-        deliverImmediately:YES];
+  [self postDistributedNotificationWithUserInfo:userInfo
+                                  notificationName:@"com.google.santa.notification.blockedeexecution"];
```
russellhancox, line 194: Line up the colons

Source/gui/SNTNotificationManager.m
```diff
@@ -170,10 +190,38 @@ - (void)postDistributedNotification:(SNTMessageWindowController *)pendingMsg {
...
+}
+
+- (void)postDistributedNotificationWithEvent:(SNTStoredEvent *)event withBundleHash: (NSString *) bundleHash {
```
russellhancox, line 197: No spaces around the second parameters type

Source/gui/SNTNotificationManager.m
```diff
@@ -170,10 +190,38 @@ - (void)postDistributedNotification:(SNTMessageWindowController *)pendingMsg {
...
+  NSDictionary *userInfo = @{
+    kFileSHA256: event.fileSHA256 ?: @"",
+    kFilePath : event.filePath ?: @"",
+    kFileBundleName : event.fileBundleName ?: @"",
+    kFileBundleID : event.fileBundleID ?: @"",
+    kFileBundleHash: bundleHash ?: @"",
+  };
+
+  [self postDistributedNotificationWithUserInfo:userInfo
+                                  notificationName:@"com.google.santa.notification.blockedeexecution.bundlehash"];
```
russellhancox, line 207: Line up colons

Source/gui/SNTNotificationManager.m
```diff
@@ -170,10 +190,38 @@ - (void)postDistributedNotification:(SNTMessageWindowController *)pendingMsg {
...
+}
+
+- (void)postDistributedNotification:(SNTMessageWindowController *)pendingMsg withProgress: (NSUInteger) progressCount {
```
russellhancox, line 210: No spaces around type

Source/gui/SNTNotificationManager.m
```diff
@@ -170,10 +190,38 @@ - (void)postDistributedNotification:(SNTMessageWindowController *)pendingMsg {
...
+  if (![pendingMsg isKindOfClass:[SNTBinaryMessageWindowController class]]) {
+    return;
+  }
+  SNTBinaryMessageWindowController *wc = (SNTBinaryMessageWindowController *)pendingMsg;
+  NSDictionary *userInfo = @{
+    kFileBundleProgress : [NSString stringWithFormat:@"%lu", (unsigned long)progressCount] ?: @"0",
```
russellhancox, line 216: The `?: @"0"` part here isn't needed, I'm pretty sure it's impossible

Source/gui/SNTNotificationManager.m
```diff
@@ -170,10 +190,38 @@ - (void)postDistributedNotification:(SNTMessageWindowController *)pendingMsg {
...
+    kFileSHA256: wc.event.fileSHA256 ?: @"",
+    kFilePath : wc.event.filePath ?: @"",
+    kFileBundleName : wc.event.fileBundleName ?: @"",
+    kFileBundleID : wc.event.fileBundleID ?: @"",
+  };
+
+  [self postDistributedNotificationWithUserInfo:userInfo
+                                  notificationName:@"com.google.santa.notification.blockedeexecution.bundlehash.progress"];
```
russellhancox, line 224: Line up colons

Source/gui/SNTNotificationManager.m
```diff
@@ -190,17 +238,26 @@ - (void)showQueuedWindow {
       if ([self.currentWindowController isKindOfClass:[SNTBinaryMessageWindowController class]]) {
         SNTBinaryMessageWindowController *controller =
           (SNTBinaryMessageWindowController *)self.currentWindowController;
+        controller.foundFileCountLabel.stringValue = @"Searching for files...";  
         dispatch_async(self.hashBundleBinariesQueue, ^{
-          [self hashBundleBinariesForEvent:controller.event withController:controller];
+          [self hashBundleBinariesForEvent:controller.event withProgressHandler:^(NSUInteger progressCount) {
+              dispatch_async(dispatch_get_main_queue(), ^{
```
russellhancox, line 244: Indent -2

Source/gui/SNTNotificationManager.m
```diff
@@ -190,17 +238,26 @@ - (void)showQueuedWindow {
...
+                //  Progress is reported to the root NSProgress
```
russellhancox, line 245: Since space after `//`

Source/gui/SNTNotificationManager.m
```diff
@@ -190,17 +238,26 @@ - (void)showQueuedWindow {
...
+                [controller.progress becomeCurrentWithPendingUnitCount:progressCount];
+              });
+          } completionHandler:^(SNTStoredEvent *event, NSString *bundleHash) {
+              [controller updateBlockNotification:event withBundleHash:bundleHash];
```
russellhancox, line 249: Indent -2

Source/gui/SNTNotificationManager.m
```diff
@@ -190,17 +238,26 @@ - (void)showQueuedWindow {
...
         });
       }
     }
   });
 }
 
 - (void)hashBundleBinariesForEvent:(SNTStoredEvent *)event
-                    withController:(SNTBinaryMessageWindowController *)withController {
-  withController.foundFileCountLabel.stringValue = @"Searching for files...";
+                withProgressHandler:(void (^)(NSUInteger progressCount))progressHandler
```
russellhancox, line 259: Line up colons

Source/gui/SNTNotificationManager.m
```diff
@@ -212,26 +269,24 @@ - (void)hashBundleBinariesForEvent:(SNTStoredEvent *)event
...
+  // Start hashing. Progress is reported 
   [[bc remoteObjectProxy]
     hashBundleBinariesForEvent:event
                          reply:^(NSString *bh, NSArray<SNTStoredEvent *> *events, NSNumber *ms) {
                            // Revert to displaying the blockable event if we fail to calculate the
                            // bundle hash
                            if (!bh)
-                             return [withController updateBlockNotification:event
-                                                             withBundleHash:nil];
+                             return completionHandler(event, nil);
```
russellhancox, line 289: The condition braces can be omitted only for a single line. Either of these are acceptable:

```
if (!bh) return completionHandler(event, nil);
```

```
if (!bh) {
  return completionHandler(event, nil);
}
```

(The previous code here was wrong)

Source/gui/SNTNotificationManagerTest.m
```diff
@@ -23,7 +23,8 @@
 
 @interface SNTNotificationManager (Testing)
 - (void)hashBundleBinariesForEvent:(SNTStoredEvent *)event
-                    withController:(SNTBinaryMessageWindowController *)controller;
+                withProgressHandler:(void (^)(NSUInteger progressCount))progressHandler
```
russellhancox, line 26: Line up colons

russellhancox (changes requested):
Most of these comments are code-style (we follow https://google.github.io/styleguide/objcguide.html pretty strictly). You can probably fix all of them by running `./Testing/fix.sh` in your repo.

tburgin:
Bundle scanning has historically been tied to a GUI pop up. It is I/O and CPU intensive in its current implementation and we want to give the user a way to cancel the scan.

If we were to move forward with silent mode bundle scanning, limits will need to be implemented on the scanning so we don’t hammer the system without and recourse for the user.

eopeter:
> Bundle scanning has been intentionally tied to a GUI pop up. It is I/O and CPU intensive in its current implementation and we want to give the user a way to cancel the scan.
> 
> If we were to move forward with silent mode bundle scanning, limits will need to be implemented on the scanning so we don’t hammer the system without recourse for the user.

Will it be acceptable to add option `EnableSilentModeBundleScanning` so that it has to be set to `true` by the user so they are opting in before we attempt bundle scanning when `EnableSilentMode` is `true`?

Source/gui/SNTNotificationManager.m
eopeter: fixed with `./Testing/fix.sh`

Source/gui/SNTNotificationManager.m
eopeter: fixed with `./Testing/fix.sh`

Source/gui/SNTNotificationManager.m
eopeter: fixed with `./Testing/fix.sh`

Source/gui/SNTNotificationManager.m
eopeter: fixed with `./Testing/fix.sh`

Source/gui/SNTNotificationManager.m
eopeter: fixed with `./Testing/fix.sh`

Source/gui/SNTNotificationManager.m
eopeter: fixed with `./Testing/fix.sh`

Source/gui/SNTNotificationManager.m
eopeter: fixed with `./Testing/fix.sh`

Source/gui/SNTNotificationManager.m
eopeter: fixed with `./Testing/fix.sh`

Source/gui/SNTNotificationManager.m
eopeter: fixed with `./Testing/fix.sh`

Source/gui/SNTNotificationManagerTest.m
eopeter: fixed with `./Testing/fix.sh`

Source/gui/SNTNotificationManager.m
eopeter: fixed with ./Testing/fix.sh

Source/gui/SNTNotificationManager.m
eopeter: fixed with ./Testing/fix.sh

Source/gui/SNTNotificationManager.m
eopeter: You are right; removed the extraneous code

Source/gui/SNTNotificationManager.m
eopeter: fixed with ./Testing/fix.sh

tburgin:
I don’t think EnableSilentModeBundleScanning is the right approach. Bundle scanning in its current state is optimized for speed at all costs due to the assumption that a user is waiting for the results. The resource impact on the system is real and should be carefully considered. Allowing Santa to behave in such a way, even behind a flag, would be misguided I believe.

I propose a second bundle scanning implementation that optimizes for low usage of system resources for use during silent mode. I am happy to consult more on how to accomplish this.

eopeter:
The implementation in this PR keeps the existing functionality as is but needs a way for the distributed listener to be able to cancel the scanning. That is the bone of contention here. Am I right? 

@russellhancox, I can attempt to impl your solution as proposed but will this current PR do for now if we can find a way to cancel the scan? 

@tburgin what is your proposal to optimize in SilentMode?

russellhancox:
> needs a way for the distributed listener to be able to cancel the scanning. That is the bone of contention here. Am I right?

We don't want a Santa process to be using significant system resources without the user being able to see what it's doing, how much progress it's making, _and_ have a way to stop it.

> I can attempt to impl your solution as proposed but will this current PR do for now if we can find a way to cancel the scan?

The current PR still allows a scan to begin without a UI present and will impact any Santa user using silent mode right now so I would not be comfortable approving this as-is.

tburgin:
> I think [my proposal](https://github.com/google/santa/issues/1212#issuecomment-1787625603) is a suitable approach without the need for new config keys or a new bundle scanning implementation, it continues to tie scanning to a UI popup but allows a custom implementation to be used.
> 
> We cannot enforce that the distributed listener process is presenting a pop-up while scanning occurs (even if it is very strongly encouraged) but at least the scanning is tied to another process on the system.

If we build this we know there are going to be scans without a UI. I think we should do everything we can to mitigate the impact of those scans. A separate scanning implementation for those instances seems reasonable. We can enforce which implementation is used, we control both ends.

eopeter:
Closing this for now and will handle bundle scanning upon receiving the distributed notification on the listener side.

tburgin:
Russell got me up to speed on the usage of silent mode to support third party custom UIs. His [proposal](https://github.com/google/santa/issues/1212#issuecomment-1787625603) makes a lot more sense to me now and should be the path forward. We don't need a second scanning implementation if the linked plan is implemented. Hope that clears some things up!

PR #1233, opened 2023-11-10
Entitlements logging config options

mlw:
Adds two configuration options supporting entitlement logging:
* `EntitlementsTeamIDFilter` - Filter out entitlements from binaries from configured organizations
* `EntitlementsPrefixFilter` - Filter out specific entitlement prefixes

### DIFF
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -643,6 +643,18 @@
 @property(readonly, nonatomic) NSUInteger metricExportTimeout;
 
 ///
+/// List of TeamIDs for which entitlements should not be logged. Use the string
+/// "platform" to refer to platform binaries.
+///
+@property(readonly, nonatomic) NSArray<NSString *> *entitlementsPrefixFilter;
+
+///
+/// List of prefix strings for which individual entitlement keys with a matching
+/// prefix should not be logged.
+///
+@property(readonly, nonatomic) NSArray<NSString *> *entitlementsTeamIDFilter;
+
+///
 ///  Retrieve an initialized singleton configurator object using the default file path.
 ///
 + (instancetype)configurator;
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -643,14 +643,14 @@
 @property(readonly, nonatomic) NSUInteger metricExportTimeout;
 
 ///
-/// List of TeamIDs for which entitlements should not be logged. Use the string
-/// "platform" to refer to platform binaries.
+/// List of prefix strings for which individual entitlement keys with a matching
+/// prefix should not be logged.
 ///
 @property(readonly, nonatomic) NSArray<NSString *> *entitlementsPrefixFilter;
 
 ///
-/// List of prefix strings for which individual entitlement keys with a matching
-/// prefix should not be logged.
+/// List of TeamIDs for which entitlements should not be logged. Use the string
+/// "platform" to refer to platform binaries.
 ///
 @property(readonly, nonatomic) NSArray<NSString *> *entitlementsTeamIDFilter;
 
--- a/Source/santad/SNTExecutionController.mm
+++ b/Source/santad/SNTExecutionController.mm
@@ -14,6 +14,7 @@
 ///    limitations under the License.
 
 #import "Source/santad/SNTExecutionController.h"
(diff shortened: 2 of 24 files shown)


Source/common/SNTConfigurator.h
```diff
@@ -642,6 +642,18 @@
 ///
 @property(readonly, nonatomic) NSUInteger metricExportTimeout;
 
+///
```
kallsyms, line 645: comments are flipped :)

Source/santad/Santad.mm
```diff
@@ -358,6 +358,52 @@ void SantadMain(std::shared_ptr<EndpointSecurityAPI> esapi, std::shared_ptr<Logg
...
+
+              // Get the value from the configurator since that method ensures proper structure
+              [exec_controller
+                updateEntitlementsTeamIDFilter:[configurator entitlementsTeamIDFilter]];
+
+              // Clear the AuthResultCache, then clear the ES cache to ensure
+              // future execs get SNTCachedDecision entitlement values filtered
+              // with the new settings.
+              auth_result_cache->FlushCache(FlushCacheMode::kAllCaches,
+                                            FlushCacheReason::kEntitlementsPrefixFilterChanged);
```
kallsyms, line 381:
```suggestion
                                            FlushCacheReason::kEntitlementsTeamIDFilterChanged);
```

Source/common/SNTConfigurator.h
mlw: Oops

Source/santad/Santad.mm
mlw: Who knew there could be downsides to copy/paste

Source/santad/SNTExecutionController.mm
```diff
@@ -38,18 +43,39 @@
...
+#include "absl/synchronization/mutex.h"
 
+using santa::common::PrefixTree;
+using santa::common::Unit;
 using santa::santad::TTYWriter;
 using santa::santad::event_providers::endpoint_security::Message;
 
 static const size_t kMaxAllowedPathLength = MAXPATHLEN - 1;  // -1 to account for null terminator
 
+void UpdateTeamIDFilterLocked(std::set<std::string> &filterSet, NSArray<NSString *> *filter) {
```
pmarkowsky, line 62: nit: We should probably write up somewhere that our "Locked" naming convention means we assume the lock is taken already before calling..

Source/santad/SNTExecutionController.mm
```diff
@@ -63,6 +89,9 @@ @interface SNTExecutionController ()
 
 @implementation SNTExecutionController {
   std::shared_ptr<TTYWriter> _ttyWriter;
+  absl::Mutex _filterMutex;
```
pmarkowsky, line 92: We should probably name this `_entitlementFilterMutex`. While we're not filtering anything else today if we we'll need to deconflict the name later.

Source/santad/SNTPolicyProcessor.m
```diff
@@ -94,8 +98,17 @@ - (nonnull SNTCachedDecision *)decisionForFileInfo:(nonnull SNTFileInfo *)fileIn
...
 
-      cd.entitlements =
-        [csInfo.signingInformation[(__bridge NSString *)kSecCodeInfoEntitlementsDict] sntDeepCopy];
+      NSDictionary *entitlements =
+        csInfo.signingInformation[(__bridge NSString *)kSecCodeInfoEntitlementsDict];
+
+      if (entitlementsFilterCallback) {
+        NSDictionary *filtered = entitlementsFilterCallback(entitlements);
+        cd.entitlements = filtered;
+        cd.entitlementsFiltered = filtered.count == entitlements.count;
```
pmarkowsky, line 107:
```suggestion
        // Set entitlements and filter status based on reduced number of entitlements.
        cd.entitlements = entitlementsFilterCallback(entitlements);
        cd.entitlementsFiltered = cd.entitlements.count == entitlements.count;
```

Seems like we don't need the filtered dictionary here.

pmarkowsky (changes requested):
Only thing is renaming the mutex. Otherwise this seems fine. 

Please treat the other suggestions as nits.

pmarkowsky (approved):
LGTM. 

Thanks for making that change.

PR #1237, opened 2023-11-16
ProcessTree: add macOS specific loader and ES adapter (2/4)

kallsyms:
The process tree library is meant to be usable by Santa, other macOS agents, and even other OSs. This PR adds the macOS specific `LoadPID` to backfill running processes on startup, and an adapter to transform from EndpointSecurity events to mutations on the tree.

### DIFF
--- a/Source/santad/ProcessTree/EndpointSecurityAdapter.mm
+++ b/Source/santad/ProcessTree/EndpointSecurityAdapter.mm
@@ -21,11 +21,7 @@
 
 namespace santa::santad::process_tree {
 
-void InformFromESEvent(int client, ProcessTree &tree, const es_message_t *msg) {
-  if (!tree.Step(client, msg->mach_time)) {
-    return;
-  };
-
+void InformFromESEvent(ProcessTree &tree, const es_message_t *msg) {
   struct pid event_pid = PidFromAuditToken(msg->process->audit_token);
   auto proc = tree.Get(event_pid);
 
--- a/Source/santad/ProcessTree/EndpointSecurityAdapter.mm
+++ b/Source/santad/ProcessTree/EndpointSecurityAdapter.mm
@@ -1,0 +1,68 @@
+/// Copyright 2023 Google LLC
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///     https://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+#include <EndpointSecurity/EndpointSecurity.h>
+#include <Foundation/Foundation.h>
+#include <bsm/libbsm.h>
+
+#include "Source/santad/ProcessTree/process_tree.h"
+#include "Source/santad/ProcessTree/process_tree_macos.h"
+#include "absl/status/statusor.h"
+
+namespace santa::santad::process_tree {
+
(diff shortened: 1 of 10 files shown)


Source/santad/ProcessTree/EndpointSecurityAdapter.mm
```diff
@@ -0,0 +1,64 @@
...
+
+  if (!proc) {
+    return;
+  }
+
+  switch (msg->event_type) {
+    case ES_EVENT_TYPE_AUTH_EXEC:
+    case ES_EVENT_TYPE_NOTIFY_EXEC: {
+      std::vector<std::string> args;
+      args.reserve(es_exec_arg_count(&msg->event.exec));
```
mlw, line 36: We probably need (or at least need to consider...) for these ES API calls to go through an instance of `EndpointSecurityAPI.h` so that they can be mocked out. There aren't tests here, but once this is hooked into Santa proper and we have tests for the new client (e.g. SNTEndpointSecurityTreeAwareClient.mm), we'll need to make sure we can properly control for these calls.

Source/santad/ProcessTree/process_tree_macos.mm
```diff
@@ -25,9 +30,111 @@
...
+  return local_argv;
+}
+}  // namespace
+
+struct Pid PidFromAuditToken(const audit_token_t &tok) {
+  return (struct Pid){.pid = audit_token_to_pid(tok),
+                      .pidversion = (uint64_t)audit_token_to_pidversion(tok)};
+}
+
 absl::StatusOr<Process> LoadPID(pid_t pid) {
```
mlw, line 104: Nit: I went looking for usages of this to see how it was going to be used, but I don't see see it ever called, even looking at Pt 4. Is this coming later still, or am I looking at the wrong thing?: https://github.com/google/santa/compare/main...kallsyms:santa:pt-4

mlw (changes requested):

Source/santad/ProcessTree/process_tree_macos.mm
kallsyms: This OS-specific impl is called from the tree's `Backfill()`: https://github.com/google/santa/blob/c35e9978d3c69e88ca8f8d91e1ade2a131784660/Source/santad/ProcessTree/process_tree_macos.mm#L53

Source/santad/ProcessTree/SNTEndpointSecurityAdapter.mm
```diff
@@ -0,0 +1,70 @@
...
+#include "Source/santad/ProcessTree/process_tree.h"
+#include "Source/santad/ProcessTree/process_tree_macos.h"
+#include "absl/status/statusor.h"
+
+using santa::santad::event_providers::endpoint_security::EndpointSecurityAPI;
+
+namespace santa::santad::process_tree {
+
+void InformFromESEvent(ProcessTree &tree, std::shared_ptr<EndpointSecurityAPI> esapi,
+                       const es_message_t *msg) {
```
mlw, line 30: I think we need to consider taking taking a `const Message &` here, which already wraps the appropriate `EndpointSecurityAPI` instance (especially looking ahead at future PRs where this function will be called from...). The problem tho is that there are not appropriate functions exposed from `Message` that allow you to meaningfully act upon the contained `EndpointSecurityAPI`.

We should chat a bit about a good path forward here.

kallsyms: after some discussion, I implemented the "basic" solution in [61701251e786dbb22800cecc64ef5769c5f93795](https://github.com/google/santa/pull/1237/commits/61701251e786dbb22800cecc64ef5769c5f93795) (exposing an ESAPI() method on `Message` and taking a `Message&` in `InformFromESEvent`).

I experimented making this interface nicer with some message subclasses that expose ESAPI wrapper methods as appropriate in https://github.com/kallsyms/santa/commit/2319da99c2a69ab54f21e3f4442e17b4199ea1fc which looks pretty nice to me, but is a separate concern from this PR.

mlw (approved):

PR #1244, opened 2023-11-29
Fix USB state issue in santactl status

### LINKED ISSUE #1217: santactl status discrepancy when running as root/non-root
When USB blocking/remounting is configured via sync server, there can be a discrepancy between what `santactl status` reports when run as root vs. non-root.

The `santactl status` command [reads `SNTConfigurator.blockUSBMount`](https://github.com/google/santa/blob/main/Source/santactl/Commands/SNTCommandStatus.m#L194) which first reads from [`syncState` then falls back to `configState`](https://github.com/google/santa/blob/d2cbddd3fbca7c9da49d531f35ca6617b406190f/Source/common/SNTConfigurator.m#L972) if it was not configured via sync. The `SNTConfigurator` singleton init loads `syncState` by [calling `readSyncStateFromDisk`](https://github.com/google/santa/blob/d2cbddd3fbca7c9da49d531f35ca6617b406190f/Source/common/SNTConfigurator.m#L248), but this bails immediately if [running with `euid != 0`](https://github.com/google/santa/blob/d2cbddd3fbca7c9da49d531f35ca6617b406190f/Source/common/SNTConfigurator.m#L1077).

Found this while working on #1214.

mlw:
Fixes #1217


Source/santad/SNTExecutionController.mm
```diff
@@ -265,15 +265,15 @@ - (void)validateExecEvent:(const Message &)esMsg postAction:(bool (^)(SNTAction)
       absl::ReaderMutexLock lock(&self->_entitlementFilterMutex);
 
       if (teamID && self->_entitlementsTeamIDFilter.count(std::string(teamID)) > 0) {
-        LOGD(@"Dropping entitlement logging for configured TeamID: %s", teamID);
+        // Dropping entitlement logging for configured TeamID
```
kallsyms, line 268: Unrelated changes?

kallsyms (approved):

Source/santad/SNTExecutionController.mm
mlw: Yeah, basically just didn't want these debug logs anymore since they're kinda noisy (and didn't feel like doing a whole PR cycle for just removing the debug prints).

pmarkowsky (approved):
LGTM.

PR #1249, opened 2023-12-04
reorder e2e tests

kallsyms:
Reorder "provisioning profile only" tests to run before the sync server is started. With #1217 fixed, `santactl` gets the proper sync server usb block state which was causing these tests to fail.


pmarkowsky (approved):
LGTM

PR #1256, opened 2023-12-13
Fix issue with drop count calculations

mlw:
The previous method had an issue where state wasn't properly carried over when values were flushed after metrics collection.


pmarkowsky (approved):
LGTM.

Issue #1255, opened 2023-12-12
Preflight `clean_sync` has no effect when no rules are sent.


np5:
See: https://github.com/google/santa/blob/2e69370524ebebe88e74830b696ceba18e890020/Source/santad/DataLayer/SNTRuleTable.m#L408-L411

I think I understand the logic behind this. If there is a bug in the sync server, we do not want to wipe all the rules. It is also easy to fix on the server by just adding a single rule. It might be just a documentation issue ?

pmarkowsky:
Correct if a sync server sends zero rules we simply return out of the function without deleting the rules in the database. If you look at the lines below that check you'll see that we start a database transaction. It's in that transaction that we delete the rules from the database. 

Can you elaborate on what you're looking to do that would require sending zero rules? 

As you pointed out if you wanted to achieve this you could send a `ALLOWLIST` rule for something like a platform binary in which case, you'd achieve the same effect. 

In general sending zero rules could be especially troublesome in lockdown mode as the client would effectively only be able to run critical system binaries or any static rules it has.

np5:
The only use-case is for the test devices that are always being re-enrolled. The risks clearly outweigh the benefits. This is why I suggested maybe just adding a warning in the docs.

pmarkowsky:
This is addressed via #1259. Marking as closed.

PR #1259, opened 2023-12-21
Added documentation to clarify clean sync with zero rule behavior

pmarkowsky:
Add documentation to address #1255.

### DIFF
--- a/Source/gui/SNTBinaryMessageWindowController.h
+++ b/Source/gui/SNTBinaryMessageWindowController.h
@@ -21,7 +21,7 @@
 ///
 ///  Controller for a single message window.
 ///
-@interface SNTBinaryMessageWindowController : SNTMessageWindowController <NSWindowDelegate>
+@interface SNTBinaryMessageWindowController : SNTMessageWindowController
 
 - (instancetype)initWithEvent:(SNTStoredEvent *)event
                     customMsg:(NSString *)message
--- a/Source/gui/SNTBinaryMessageWindowController.h
+++ b/Source/gui/SNTBinaryMessageWindowController.h
@@ -21,7 +21,7 @@
 ///
 ///  Controller for a single message window.
 ///
-@interface SNTBinaryMessageWindowController : SNTMessageWindowController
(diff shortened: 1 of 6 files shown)


Source/gui/SNTBinaryMessageWindowController.h
```diff
@@ -21,7 +21,7 @@
 ///
 ///  Controller for a single message window.
 ///
-@interface SNTBinaryMessageWindowController : SNTMessageWindowController
+@interface SNTBinaryMessageWindowController : SNTMessageWindowController <NSWindowDelegate>
```
russellhancox, line 24: SNTMessageWindowController already declares that it implements NSWindowDelegate.

docs/development/sync-protocol.md
```diff
@@ -319,6 +319,8 @@ If a clean sync was not requested by either the client or the sync service, then
 
 Santa applies rules idempotently and is designed to receive rules multiple times without issue.
 
+One caveat to be aware of is that when a clean sync is requested in the `preflight` stage, the client expects that at least one rule will be sent by the sync service in the `ruledownload` stage. If no rules are sent then the client is expected to keep its old set of rules prior to the client or server requesting a clean sync.
```
russellhancox, line 322:
```suggestion
One caveat to be aware of is that when a clean sync is requested in the `preflight` stage, the client expects that at least one rule will be sent by the sync service in the `ruledownload` stage. If no rules are sent then the client is expected to keep its old set of rules prior to the client or server requesting a clean sync and the client will continue to request a clean sync on all subsequent syncs until a successful sync completes that includes at least one rule.
```

Source/gui/SNTBinaryMessageWindowController.h
pmarkowsky: Apologies this was an accident based on VSCode and delirium from a bad stomach flu. 

Should be fixed.

mlw (approved):

mlw (approved):

PR #1271, opened 2024-01-13
Add back support for EnableForkAndExitLogging config key

### LINKED ISSUE #1269: The configuration key `EnableForkAndExitLogging` is broken
The configuration key `EnableForkAndExitLogging` is supposed to control whether or not Santa logs FORK/EXIT event types. It is supposed to default to `false`.

However, in `v2022.9`, support for the key was inadvertently removed. The effect is that FORK/EXIT events are ***always*** logged, regardless of the setting.

Support for this key should be added back with the original `false` default.

mlw:
This PR adds back support for this configuration key that was inadvertently removed in v2022.9.

The default value is meant to be `false`. However since v2022.9, this value has effectively been hardcoded to `true`. This change fixes the logic back to the original intention.

**IMPORTANT: This means that if you want `FORK`/`EXIT` events to continue to be logged, you must set the `EnableForkAndExitLogging` config key to `true`**

Fixes #1269

### DIFF
--- a/Source/santad/EventProviders/SNTEndpointSecurityRecorder.mm
+++ b/Source/santad/EventProviders/SNTEndpointSecurityRecorder.mm
@@ -134,6 +134,13 @@ es_file_t *GetTargetFileForPrefixTree(const es_message_t *msg) {
 
   [self.compilerController handleEvent:esMsg withLogger:self->_logger];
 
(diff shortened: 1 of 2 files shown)


pmarkowsky (approved):
LGTM.

PR #1275, opened 2024-01-19
Sync clean all

### LINKED ISSUE #1272: Clean syncs should only optionally remove transitive rules
Currently, when a clean sync is performed all rules in the local rules database are first deleted before adding the newly downloaded rules. This means that any transitive rules in the database are also removed which causes issues for deployments using compiler rules.

This behavior should be changed so that by default only non-transitive rules are deleted.

The following changes will be required (though likely not an exhaustive list):

1. `santactl sync --clean` will only remove non-transitive rules.
2. A new flag will be added to `santactl sync` to force the current behavior of removing all rules
3. A new key will be supported in the Preflight request to differentiate the new clean method
4. A new key will be supported in the Preflight response indicating the type of sync the server is expecting to perform

mlw:
**\*\*\*IMPORTANT: This PR introduces a change in behavior!\*\*\***

Currently, clean syncs result in the entire rules database being deleted and newly received rules being written. However this means that any transitive rules are also wiped out, causing pain for users.

This PR changes the meaning of clean syncs to instead only delete non-transitive rules before writing the newly received rules.

The following describes the new behavior:
1. `santactl sync --clean` will only remove non-transitive rules
2. `santactl sync --clean-all` is a new flag that will remove all rules (like `--clean` before this PR)
3. The preflight response key `clean_sync` is now deprecated
4. The preflight response key `clean_sync` is treated as only removing non-transitive rules
5. A new preflight response key is supported, `sync_type`
6. The `sync_type` preflight response key supports the following values
    * `normal` - Only new rules will be added, existing rules will not be removed
    * `clean` - Remove non-transitive rules before writing new rules
    * `clean_all` - Remove all rules before writing new rules
    * Not set - Same as `normal`
7. The server can still force a `clean` or `clean_all` action in the preflight response
8. The client can force a `clean_all` action even if the server only responded with `clean`
    * This is because from the server's perspective, this is the same operation. A user is allowed to dictate if all rules or only transitive rules should be removed (presuming the server doesn't force all rules)

Fixes #1272


pmarkowsky (dismissed):
LGTM.

docs/development/sync-protocol.md
```diff
@@ -156,7 +157,24 @@ The JSON object has the following keys:
...
+
+The rules for resolving the type of sync that will be performed are as follows:
+1. If the server responds with a `sync_type` of `clean`, a clean sync is performed (regardless of whether or not it was requested by the client), unless the client sync type state was `clean_all`, in which case a `clean_all` sync type is performed.
+2. If the server responded that it is performing a `clean_all` sync, a `clean all` is performed (regardless of whether or not it was requested by the client)
+3. Otherwise, a normal sync is performed
+
+A client that has a `clean` or `clean_all` sync type state set will continue to request a clean sync until it is satisfied by the server. A clean sync is never performed until the server indicates it is prepared to send the entire set of rules.
```
russellhancox, line 173: It's unclear to me what the second sentence is saying.

Source/santasyncservice/SNTSyncPostflight.m
```diff
@@ -42,11 +42,11 @@ - (BOOL)sync {
...
-  // Remove clean sync flag if we did a clean sync
-  if (self.syncState.cleanSync) {
-    [rop setSyncCleanRequired:NO
-                        reply:^{
-                        }];
+  // Remove clean sync flag if we did a clean or clean all sync
+  if (self.syncState.syncType != SNTSyncTypeNormal) {
+    [rop setSyncTypeRequired:SNTSyncTypeNormal
+                       reply:^{
+                       }];
```
russellhancox, line 49:
```suggestion
  if (self.syncState.syncType != SNTSyncTypeNormal) {
    [rop setSyncTypeRequired:SNTSyncTypeNormal reply:^{}];
```

russellhancox: Ack.

russellhancox (approved):

Issue #1276, opened 2024-01-22
Make santactl status report on the status of enableTransitiveRules even when not using a sync server


pmarkowsky:
Right now we only report the status of transitive rules when a Sync Server URL is set

https://github.com/google/santa/blob/3b7061ea62a902bda529e52c61312aa3dd1ce130/Source/santactl/Commands/SNTCommandStatus.m#L304-L314

We need to change this so that we're always reporting if transitive rules are enabled for users using either MDM only or standalone.

pmarkowsky:
Fixed in #1277

PR #1277, opened 2024-01-22
Make santactl status always print out transitive rule status if set 

pmarkowsky:
Fix for #1276 

Also fixes a typo in `SNTCommandRule.m`.

### DIFF
--- a/Source/santactl/Commands/SNTCommandStatus.m
+++ b/Source/santactl/Commands/SNTCommandStatus.m
@@ -262,6 +262,11 @@ REGISTER_COMMAND_NAME(@"status")
   } else {
     printf(">>> Daemon Info\n");
     printf("  %-25s | %s\n", "Mode", [clientMode UTF8String]);
+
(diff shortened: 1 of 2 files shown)


Source/santactl/Commands/SNTCommandStatus.m
```diff
@@ -310,6 +310,9 @@ - (void)runWithArguments:(NSArray *)arguments {
       printf("  %-25s | %s\n", "Push Notifications",
              (pushNotifications ? "Connected" : "Disconnected"));
       printf("  %-25s | %s\n", "Bundle Scanning", (enableBundles ? "Yes" : "No"));
+    }
+
+    if (enableTransitiveRules) {
```
mlw, line 315: I think we should move this up to `Daemon Info` above (like other things that can be configured via profile or sync server)

mlw (changes requested):

mlw (approved):

PR #1281, opened 2024-02-05
ProcessTree: integrate process tree throughout the event processing lifecycle (3/4)

kallsyms:
This PR:
* Instantiates a process tree in SantaDeps, and threads it through to the enricher and proto serializer.
* Adds a new `SNTEndpointSecurityTreeAwareClient` which clients subclass from to automatically have the tree updated before they run.
* Adds a `ProcessToken` to the core `Message` type passed through Santa, which causes the tree to automatically "hold on" to process information, even after the process' exit event is received. This means the clients can get process information out of the tree at any point, even in (delayed) async processing.
* Adds the basic configuration knobs to change which annotations are enabled (and if none are, disables the tree entirely).

N.B. Due to the last bullet above, while this code does technically begin to interact with event processing, the tree remains entirely disabled unless the config key is set.


Source/santad/EventProviders/SNTEndpointSecurityTreeAwareClient.mm
```diff
@@ -0,0 +1,104 @@
...
+  }
+  if (events.find(ES_EVENT_TYPE_NOTIFY_EXIT) == events.end()) {
+    eventsWithLifecycle.insert(ES_EVENT_TYPE_NOTIFY_EXIT);
+    _addedEvents[ES_EVENT_TYPE_NOTIFY_EXIT] = true;
+  }
+
+  return [super subscribe:eventsWithLifecycle];
+}
+
+- (bool)handleContextMessage:(Message &)esMsg {
```
mlw, line 66: Add some comments about how this enforces specific event types are subscribed and will prevent messages from flowing to subclasses if those events were not directly subscribed.

mlw (changes requested):

Source/santad/EventProviders/EndpointSecurity/Enricher.h
```diff
@@ -32,7 +32,7 @@ enum class EnrichOptions {
 
 class Enricher {
  public:
-  Enricher();
+  Enricher(std::shared_ptr<process_tree::ProcessTree> pt);
```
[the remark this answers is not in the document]
kallsyms: Makes sense to me - changed and test changes reverted.

Source/santad/EventProviders/SNTEndpointSecurityClient.mm
```diff
@@ -142,6 +146,10 @@ - (void)establishClientOrDie {
     // sequence numbers are processed in order.
     self->_metrics->UpdateEventStats(self->_processor, esMsg.operator->());
 
+    if ([self handleContextMessage:esMsg]) {
+      return;
```
mlw, line 150: We probably want a call to record metrics here.

Source/santad/EventProviders/SNTEndpointSecurityTreeAwareClient.mm
```diff
@@ -0,0 +1,114 @@
...
+                    (std::shared_ptr<santa::santad::process_tree::ProcessTree>)processTree {
+  self = [super initWithESAPI:std::move(esApi) metrics:std::move(metrics) processor:processor];
+  if (self) {
+    _processTree = std::move(processTree);
+    _addedEvents.resize(ES_EVENT_TYPE_LAST, false);
+  }
+  return self;
+}
+
+// ES guarantees logical consistency within a client (e.g. forks always preceed exits),
```
mlw, line 49: preceed --> precede

mlw (changes requested):

mlw (approved):

PR #1285, opened 2024-02-08
Respect fail closed on deadlines

### LINKED ISSUE #1283: deadline exceeded in monitor mode still ends up denying execution
When deadlines are exceeded here:

https://github.com/google/santa/blob/e8db89c57c41dde73fa447bca206faf9075eb484/Source/santad/EventProviders/SNTEndpointSecurityClient.mm#L296-L299

The process is still denied execution even if Santa is in monitor mode.

Also: these denials are not logged in the normal ALLOW/DENY logging (that end up in `/var/db/santa/santa.log*` files.

mlw:
Santa attempts to respond to events before system deadlines in case event processing took too long for some reason. This previously only resulted in events being denied.

This PR changes two things related to the default respons Santa will make before deadlines expire:

1. When in `Lockdown` mode, `FailClosed` configuration will now be respected
2. When in `Monitor` mode, Santa will ALLOW the events

Fixes #1283


kallsyms (approved):

Source/common/SNTConfigurator.m
```diff
@@ -604,8 +604,11 @@ - (void)setSyncServerClientMode:(SNTClientMode)newMode {
 
 - (BOOL)failClosed {
   NSNumber *n = self.configState[kFailClosedKey];
-  if (n) return [n boolValue];
-  return NO;
+  if ([n boolValue] && self.clientMode == SNTClientModeLockdown) {
+    return YES;
+  } else {
+    return NO;
+  }
```
russellhancox, line 611: `return [n boolValue] && self.clientMode == SNTClientModeLockdown`

russellhancox (approved):

russellhancox (approved):

PR #1296, opened 2024-02-20
ProcessTree: add the first annotation, originator (4/4)

kallsyms:
This PR adds the first annotator to the tree and to Santa, allowing processes to be labeled with which of a handful of potentially long-running "originators" created the process. Current tracked originators are `/usr/bin/login` and `/usr/sbin/cron`.

### DIFF
--- a/Source/santad/ProcessTree/annotations/originator.cc
+++ b/Source/santad/ProcessTree/annotations/originator.cc
@@ -13,8 +13,14 @@
 /// limitations under the License.
 #include "Source/santad/ProcessTree/annotations/originator.h"
 
+#include <memory>
 #include <optional>
+#include <string>
+#include <utility>
 
+#include "absl/container/flat_hash_map.h"
+#include "Source/santad/ProcessTree/process.h"
+#include "Source/santad/ProcessTree/process.h"
 #include "Source/santad/ProcessTree/process_tree.h"
 #include "Source/santad/ProcessTree/process_tree.pb.h"
 
--- a/Source/santad/ProcessTree/annotations/originator.cc
+++ b/Source/santad/ProcessTree/annotations/originator.cc
@@ -15,8 +15,8 @@
 
 #include <optional>
 
-#include "Source/santad/ProcessTree/process_tree.pb.h"
 #include "Source/santad/ProcessTree/process_tree.h"
+#include "Source/santad/ProcessTree/process_tree.pb.h"
 
 namespace ptpb = ::santa::pb::v1::process_tree;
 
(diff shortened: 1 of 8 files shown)


Source/santad/ProcessTree/annotations/originator.cc
```diff
@@ -0,0 +1,61 @@
...
+namespace santa::santad::process_tree {
+
+void OriginatorAnnotator::AnnotateFork(ProcessTree &tree, const Process &parent,
+                                       const Process &child) {
+  // "Base case". Propagate existing annotations down to descendants.
+  if (auto annotation = tree.GetAnnotation<OriginatorAnnotator>(parent)) {
+    tree.AnnotateProcess(child, std::move(*annotation));
+  }
+}
+
```
mlw, line 32: Outside of scope for this PR, but I think this makes it clear in the future having access to codesign info would be very powerful for macOS annotators. Has there been any thought to how this information could be propagated, maybe in a `Process` object?

Source/santad/ProcessTree/annotations/originator.cc
```diff
@@ -0,0 +1,61 @@
...
+void OriginatorAnnotator::AnnotateExec(ProcessTree &tree,
+                                       const Process &orig_process,
+                                       const Process &new_process) {
+  if (auto annotation = tree.GetAnnotation<OriginatorAnnotator>(orig_process)) {
+    tree.AnnotateProcess(new_process, std::move(*annotation));
+    return;
+  }
+
```
mlw, line 40: I think we should consider making this a map. Something like

```
absl::flat_hash_map<std::string, ::santa::pb::v1::process_tree::Annotations::Originator>
```

Would make it easier to extend and get rid of the near-duplicate calls to `AnnotateProcess`.

Source/santad/ProcessTree/process_tree.proto
```diff
@@ -3,4 +3,11 @@ syntax = "proto3";
 package santa.pb.v1.process_tree;
 
 message Annotations {
+  enum Originator {
+    UNSPECIFIED = 0;
+    LOGIN = 1;
+    CRON = 2;
+  }
+
+  Originator ancestor = 1;
```
mlw, line 12: Was there a particular reason to not name this `originator`? Given `ancestor` is more generic, could potentially step on toes of future annotators.

Source/santad/SantadDeps.mm
```diff
@@ -159,8 +160,11 @@
   std::vector<std::unique_ptr<process_tree::Annotator>> annotators;
 
   for (NSString *annotation in [configurator enabledProcessAnnotations]) {
-    // TODO(nickmg): add annotation name switch
-    (void)annotation;
+    if ([annotation isEqualToString:@"Originator"]) {
+      annotators.emplace_back(std::make_unique<process_tree::OriginatorAnnotator>());
+    } else {
+      LOGE(@"Unrecognized process annotation %@", annotation);
```
mlw, line 166: super nit: `LOGW`

Source/santad/SantadDeps.mm
```diff
@@ -159,8 +160,11 @@
   std::vector<std::unique_ptr<process_tree::Annotator>> annotators;
 
   for (NSString *annotation in [configurator enabledProcessAnnotations]) {
-    // TODO(nickmg): add annotation name switch
-    (void)annotation;
+    if ([annotation isEqualToString:@"Originator"]) {
```
mlw, line 163: nit: Should maybe do something like this to be a bit more flexible:
```
[[annotation lowercaseString] isEqualToString:@"originator"]
```

mlw (changes requested):

Source/santad/ProcessTree/process_tree.proto
kallsyms: Nope, just an oversight from earlier revisions.

Source/santad/ProcessTree/annotations/originator.cc
kallsyms: Yep so the thought right now is that any non-OS-agnostic stuff (e.g. macOS codesigning) will be in an `agent_specific` or `os_specific` struct in `Process` (defined in the `process_tree` OS-specific headers) which anything can then get/set. This implementation will change slightly when the internal changes are migrated out (notably the fact that everything is a proto for serialization) but the idea remains the same.

mlw (approved):

PR #1298, opened 2024-02-26
Refactor rule and count lookups

mlw:
This change streamlines rule lookups and rule count lookups. The goal is to make it easier to add supported rule types by reducing the number of interface changes required.

This PR is not intended to introduce/change any functionality.

### DIFF
--- a/Source/santactl/Commands/SNTCommandStatus.m
+++ b/Source/santactl/Commands/SNTCommandStatus.m
@@ -91,23 +91,14 @@ REGISTER_COMMAND_NAME(@"status")
   }];
 
   // Database counts
-  __block int64_t eventCount = -1;
-  __block int64_t binaryRuleCount = -1;
-  __block int64_t certRuleCount = -1;
-  __block int64_t teamIDRuleCount = -1;
-  __block int64_t signingIDRuleCount = -1;
-  __block int64_t compilerRuleCount = -1;
-  __block int64_t transitiveRuleCount = -1;
-  [rop databaseRuleCounts:^(int64_t binary, int64_t certificate, int64_t compiler,
-                            int64_t transitive, int64_t teamID, int64_t signingID) {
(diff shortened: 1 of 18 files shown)


Source/santactl/Commands/SNTCommandStatus.m
```diff
@@ -91,23 +91,14 @@ - (void)runWithArguments:(NSArray *)arguments {
...
-  [rop databaseRuleCounts:^(int64_t binary, int64_t certificate, int64_t compiler,
-                            int64_t transitive, int64_t teamID, int64_t signingID) {
-    binaryRuleCount = binary;
-    certRuleCount = certificate;
-    teamIDRuleCount = teamID;
-    signingIDRuleCount = signingID;
-    compilerRuleCount = compiler;
-    transitiveRuleCount = transitive;
+  __block struct RuleCounts ruleCounts;
+  memset(&ruleCounts, NSUIntegerMax, sizeof(ruleCounts));
```
russellhancox, line 95:
```suggestion
  __block struct RuleCounts ruleCounts = {0};
```

russellhancox: Unless I'm missing a reason this is initializing to NSUintegerMax instead of 0.

mlw: Previously, all counts were initialized to -1. This kinda made sense because the values were ints (even though data was obtained as an `NSUInteger`). However if there was some XPC failure, the default of -1 would get displayed indicating an error. 

I made the types of the struct match the method signatures, but I now think that isn't quite right. The query returns a signed `long`, and that gets converted to an `NSUInteger`. I think I should instead fix the rule count methods to return long and then make this struct match those same types. Then we can go back to the `-1` default as well.

russellhancox (approved):

PR #1299, opened 2024-02-28
Creating transitive rules for rename events should fallback to destination path

mlw:
When a transitive rule is being created during a RENAME event, currently only the source path is used. However, the rename operation may have already succeeded and the path no longer exists. This change makes it so that the destination path is used in the event the source no longer exists.


pmarkowsky (approved):
LGTM.

PR #1300, opened 2024-03-01
Added clean flags for JSON rule import

### LINKED ISSUE #1293: `santactl rule` should allow clearing ruleset
### Discussed in https://github.com/google/santa/discussions/1292

<div type='discussions-op-text'>

<sup>Originally posted by **jumpsdefcon** February 14, 2024</sup>
So, was doing some testing with an agent that is not using a sync server, and was looking to manual import the json rule list.  I got the list to import fine, but when I try to import a different rule list I still see the full list of rules in the database.  Do I need to script out a removal line by line and then re-import the new list or perhaps script removal and reinstall of the agent?

I would have expected the import to purge the current list and then import the rules in the new config file.</div>

We should add a `--clean` flag that can both delete all rules on its own or clean before importing new when combined with the `--import` flag.

pmarkowsky:
This PR adds `--clean` and `--clean-all` flags to the `santactl rule` command to allow clearing the rule database when importing rules via JSON.

Fixes #1293

### DIFF
--- a/Source/santactl/Commands/SNTCommandRule.m
+++ b/Source/santactl/Commands/SNTCommandRule.m
@@ -45,58 +45,65 @@ REGISTER_COMMAND_NAME(@"rule")
 }
 
 + (NSString *)longHelpText {
-  return (@"Usage: santactl rule [options]\n"
-          @"  One of:\n"
-          @"    --allow: add to allow\n"
-          @"    --block: add to block\n"
-          @"    --silent-block: add to silent block\n"
-          @"    --compiler: allow and mark as a compiler\n"
-          @"    --remove: remove existing rule\n"
-          @"    --check: check for an existing rule\n"
-          @"    --import {path}: import rules from a JSON file\n"
-          @"    --export {path}: export rules to a JSON file\n"
-          @"\n"
-          @"  One of:\n"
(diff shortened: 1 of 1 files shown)


Source/santactl/Commands/SNTCommandRule.m
```diff
@@ -181,6 +189,16 @@ - (void)runWithArguments:(NSArray *)arguments {
         [self printErrorUsageAndExit:@"--import requires an argument"];
       }
       jsonFilePath = arguments[i];
+    } else if ([arg caseInsensitiveCompare:@"--clean"] == NSOrderedSame) {
+      if (!importRules) {
```
mlw, line 193: It might be nice to consider moving this check below this chunk of code checking arguments. As written, you cannot do `santactl rule --clean --import ...` which might be confusing.

Maybe something below like:

```
if (!importRules && cleanupType != SNTRuleCleanupNone) {
    // [self printErrorUsageAndExit...]
}
```

If order does matter for some reason and you'd like to keep as is, might be worth documenting the requirement in the usage string.

mlw (changes requested):

mlw (approved):

Issue #1200, opened 2023-10-13
Support for advanced rulesets, including evaluation of flags, via Common Expression Language


keeleysam:
(discussed IRL with @mlw, opening an issue for discussion)

Santa's process based rulesets currently apply either as an allow or block to all invocations of a executable.  This works well for many use cases, but there are times when one may wish to have a rule which allows execution only under certain conditions.  

This to mind because of a presentation by @r3ggi about using previously granted permissions of Electron applications though injection of JavaScript, which is possible on many applications though the use of `--inspect` flags.  https://www.electronjs.org/docs/latest/api/command-line-switches#--inspecthostport

In this case, most organizations will allow the use of Electron apps, but to mitigate this issue may want to block execution if a `--inspect` flag is in the parameters upon launch.

Likewise, organizations might look to use Santa to mitigate vulnerabilities similar to cURL's CVE-2023-38545, which might be exploitable on macOS versions older than 14.0.1 if `/usr/bin/curl` is launched with a `–-socks5-hostname` flag or `socks5h://` specified for a `-x`, `--proxy`, or `--preproxy` flag.

There are other cases where this functionality could be useful.  For example, organizations may want to be sure that Chromium based browsers aren't able to be launched with flags like `--force-devtools-available`, `--remote-allow-origins *`, `--remote-debugging-address 0.0.0.0` or `--unsafely-treat-insecure-origin-as-secure`.

These are just examples of where being able to write an advanced rule would help with flags on a process, but there are potentially many other use cases where being able to craft an advanced rule would help.

I propose that https://github.com/google/cel-spec be utilized for these types of rules, and there is a C++ library available as well at https://github.com/google/cel-cpp.  CEL is very performant and flexible, and utilized heavily in public-facing Google products such as Conditional IAM in GCP.

For how the rulesets might be structured, this should be a new ruletype, and probably be _more_ specific than the current `BINARY` type.  All existing rules should be able to be implemented as advanced rules, though there should be no need to do so.

As an example, where a current rule looks like this:
```
  {
    "identifier": "EQHXZ8M8AV",
    "rule_type": "TEAMID",
    "policy": "ALLOWLIST",
    "custom_msg": "Allow Software Google's Team ID",
    "creation_time": 1576623399.151607
  }
```
an advanced version of the same rule could look like:
```
  {
    "rule_text": "process.team_id == \"EQHXZ8M8AV\"",
    "rule_type": "ADVANCED",
    "policy": "ALLOWLIST",
[...]

PR #1304, opened 2024-03-08
Update SNTPolicyProcessor to use a map

pmarkowsky:
This PR refactors SNTPolicyProcessor to use a flat_hash_map instead of the switch statement. 

This is being done for two reasons: 

1. To make it easier to add rules in the future or change the states
2. In preparation for more advanced rules e.g. something to solve #1200


Source/santad/SNTPolicyProcessor.mm
```diff
@@ -44,6 +45,123 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
...
+      {{SNTRuleTypeTeamID, SNTRuleStateAllow}, SNTEventStateAllowTeamID},
+      {{SNTRuleTypeTeamID, SNTRuleStateSilentBlock}, SNTEventStateBlockTeamID},
+      {{SNTRuleTypeTeamID, SNTRuleStateBlock}, SNTEventStateBlockTeamID},
+    };
+
+  auto iterator = decisions.find(std::pair<SNTRuleType, SNTRuleState>{rule.type, rule.state});
+  if (iterator != decisions.end()) {
+    cd.decision = iterator->second;
+  } else {
+    // We don't have a matching rule type, rule state for this.
```
russellhancox, line 79: Expand this to note that these two states aren't final and could be overridden by "scope" rules.

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -44,6 +45,123 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
...
+    if (mode == SNTClientModeMonitor) {
+      cd.decision = SNTEventStateAllowUnknown;
+    } else {
+      cd.decision = SNTEventStateBlockUnknown;
+    }
+  }
+
+  if (rule.state == SNTRuleStateSilentBlock) {
```
russellhancox, line 87: This if/else if chain would be more readable as a switch

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -44,6 +45,123 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
...
+             raise:@"Invalid compiler rule type"
+            format:@"updateCachedDecisionForRule: Unexpected compiler rule type: %ld", rule.type];
+          break;
+      }
+    }
+  } else if (rule.state == SNTRuleStateAllowTransitive) {
+    // If transitive rules are enabled, then SNTRuleStateAllowTransitive rules
+    // become SNTEventStateAllowTransitive decisions.  Otherwise, we treat the
+    // rule as if it were SNTRuleStateUnknown.
+    if (!enableTransitiveRules) {
```
russellhancox, line 109: Should this be `if (enableTransitiveRules)`?

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -44,6 +45,123 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
...
+      cd.signingID = nil;
+    }
+  }
+
+  NSDictionary *entitlements =
+    csInfo.signingInformation[(__bridge NSString *)kSecCodeInfoEntitlementsDict];
+
+  if (entitlementsFilterCallback) {
+    cd.entitlements = entitlementsFilterCallback(entitlements);
+    cd.entitlementsFiltered = (cd.entitlements.count == entitlements.count);
```
russellhancox, line 157:
```suggestion
    cd.entitlementsFiltered = (cd.entitlements.count != entitlements.count);
```

russellhancox (changes requested):

Source/santad/SNTPolicyProcessor.mm
pmarkowsky: Nope the map takes care of this for us, but the comment should probably be updated. This is to fallback in the event that we're not enabling transitive rules but have transitive states from the database.

russellhancox: Ack. The extra comment makes it clearer

russellhancox (approved):

russellhancox (dismissed):

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -44,6 +45,128 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
...
+      // become SNTEventStateAllowTransitive decisions.  Otherwise, we treat the
+      // rule as if it were SNTRuleStateUnknown.
+      if (!enableTransitiveRules) {
+        // check operating mode.
+        if (mode == SNTClientModeMonitor) {
+          cd.decision = SNTEventStateAllowUnknown;
+        } else {
+          cd.decision = SNTEventStateBlockUnknown;
+        }
+      }
```
mlw, line 119: There is a delta here with the existing handling for transitive rules. The new code has a second, inner if-check for the mode which doesn't currently happen. The new code also isn't setting the decision to `SNTEventStateAllowTransitive` properly when transitive rules are enabled.

Existing code:
```
if ([self.configurator enableTransitiveRules]) {
  cd.decision = SNTEventStateAllowTransitive;
  return cd;
} else {
  rule.state = SNTRuleStateUnknown;
}
```

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -44,6 +45,128 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
   return self;
 }
 
+// This method applies the rules to the cached decision object.
+inline SNTCachedDecision *updateCachedDecisionForRule(SNTCachedDecision *cd,
```
mlw, line 49: I think we should consider a small change in this interface. As implemented, an in-band signal is used (via `SNTEventStateBlockUnknown` or `SNTEventStateAllowUnknown`) to indicate further processing is needed in the caller.

Since we mutate the provided `SNTCachedDecision`, there isn't much need to return it as well. We could isntead return a boolean as to whether or not a decision was able to be obtained:

```
static BOOL updateCachedDecisionForRule(
    SNTCachedDecision *cd,
    SNTRule *rule,
    // SNTClientMode mode, Note: This parameter wouldn't be needed anymore
    bool enableTransitiveRules)
```

I think this will help simplify the logic and be easier to maintain.

If we go this route, it might be good to change the function name. Maybe `decisionForRule`? Or some name that also makes it clear the provided `SNTCachedDecision` is still mutated would be nice.

(Additional comments in this function are made assuming this change is made)

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -44,6 +45,128 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
...
+  if (iterator != decisions.end()) {
+    cd.decision = iterator->second;
+  } else {
+    // We don't have a matching rule type, rule state for this.
+    // These states aren't final and may be changed by scope rules.
+    if (mode == SNTClientModeMonitor) {
+      cd.decision = SNTEventStateAllowUnknown;
+    } else {
+      cd.decision = SNTEventStateBlockUnknown;
+    }
```
mlw, line 86: (If the interface change mentioned previously is made...) This whole if/else block (and the enclosing else block) can be deleted

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -44,6 +45,128 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
...
+inline SNTCachedDecision *updateCachedDecisionForRule(SNTCachedDecision *cd, 
+                                                      SNTRule *rule,
+                                                      SNTClientMode mode,
+                                                      bool enableTransitiveRules) {
+  static const auto decisions =
+    absl::flat_hash_map<std::pair<SNTRuleType, SNTRuleState>, SNTEventState>{
+      {{SNTRuleTypeCDHash, SNTRuleStateAllow}, SNTEventStateAllowCDHash},
+      {{SNTRuleTypeCDHash, SNTRuleStateAllowCompiler}, SNTEventStateAllowCompiler},
+      {{SNTRuleTypeCDHash, SNTRuleStateAllowTransitive}, SNTEventStateAllowTransitive},
+      {{SNTRuleTypeCDHash, SNTRuleStateBlock}, SNTEventStateBlockCDHash},
```
mlw, line 58: I think CDHash is missing a row for silent block

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -44,6 +45,128 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
...
+    // We don't have a matching rule type, rule state for this.
+    // These states aren't final and may be changed by scope rules.
+    if (mode == SNTClientModeMonitor) {
+      cd.decision = SNTEventStateAllowUnknown;
+    } else {
+      cd.decision = SNTEventStateBlockUnknown;
+    }
+  }
+
+  switch (rule.state) {
```
mlw, line 89: (If the interface change mentioned previously is made...) The `break` statements should be changed to `return YES;`

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -44,6 +45,128 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
...
+    default: break;
+  }
+
+  if (rule.customMsg) {
+    cd.customMsg = rule.customMsg;
+  }
+
+  if (rule.customURL) {
+    cd.customURL = rule.customURL;
+  }
```
mlw, line 130: (If the interface change mentioned previously is made...) Setting these two properties should be moved up above the `switch` (just after the map lookup) so that they are set before any return

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -44,6 +45,128 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
...
+
+  return cd;
```
mlw, line 132: (If the interface change mentioned previously is made...) Replace with `return NO;`

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -138,116 +238,9 @@ - (nonnull SNTCachedDecision *)decisionForFileInfo:(nonnull SNTFileInfo *)fileIn
...
-            cd.decision = SNTEventStateBlockTeamID;
-            return cd;
-          default: break;
-        }
-        break;
-
-      default: break;
+    cd = updateCachedDecisionForRule(cd, rule, mode, self.configurator.enableTransitiveRules);
+    if (cd.decision != SNTEventStateBlockUnknown || cd.decision != SNTEventStateAllowUnknown) {
+      return cd;
```
mlw, line 243: (If the interface change mentioned previously is made...) this would become:
```
if (updateCachedDecisionForRule(cd, rule, mode, self.configurator.enableTransitiveRules)) {
  return cd;
}
```

mlw (changes requested):

Source/santad/SNTPolicyProcessor.mm
pmarkowsky: The issue with doing this is that the special case rules that are changed based on configuration may require setting the `customURL` and `customMsg` in multiple branches or unsetting them. 

I think its simpler with the keeping the default as a break.

Source/santad/SNTPolicyProcessor.mm
pmarkowsky: Changed the interface to return a BOOL but also changed the function back to an objective-c method.

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -44,6 +45,119 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
...
+//
+// It returns YES if the decision was made, NO if the decision was not made.
+- (BOOL)decision:(SNTCachedDecision *)cd
+              forRule:(SNTRule *)rule
+  withTransitiveRules:(BOOL)enableTransitiveRules {
+  static const auto decisions =
+    absl::flat_hash_map<std::pair<SNTRuleType, SNTRuleState>, SNTEventState>{
+      {{SNTRuleTypeCDHash, SNTRuleStateAllow}, SNTEventStateAllowCDHash},
+      {{SNTRuleTypeCDHash, SNTRuleStateAllowCompiler}, SNTEventStateAllowCompiler},
+      {{SNTRuleTypeCDHash, SNTRuleStateAllowTransitive}, SNTEventStateAllowTransitive},
```
mlw, line 58: I think `SNTRuleStateAllowTransitive` is only valid for `SNTRuleTypeBinary`, not `SNTRuleTypeCDHash`

Source/santad/SNTPolicyProcessorTest.m
```diff
@@ -0,0 +1,564 @@
...
+
+#import "Source/santad/SNTPolicyProcessor.h"
+
+@interface SNTPolicyProcessorTest : XCTestCase
+@property SNTPolicyProcessor *processor;
+@end
+
+@implementation SNTPolicyProcessorTest
+- (void)setUp {
+  _processor = [[SNTPolicyProcessor alloc] init];
```
mlw, line 31: style nit: `self.processor`

Source/santad/SNTPolicyProcessorTest.m
```diff
@@ -0,0 +1,564 @@
...
+  SNTCachedDecision *cd = [[SNTCachedDecision alloc] init];
+  cd.sha256 = rule.identifier;
+
+  [self.processor decision:cd forRule:rule withTransitiveRules:YES];
+
+  XCTAssertEqualObjects(cd.customMsg, @"Custom Message");
+  XCTAssertEqualObjects(cd.customURL, @"https://example.com");
+}
+
+@end
```
mlw, line 564: style nit: missing new line

mlw (changes requested):

Source/santad/SNTPolicyProcessor.mm
pmarkowsky: Wasn't sure if we wanted to add that. Meant to call that out in comments. Will remove and add in a later PR.

mlw (approved):

Source/santad/SNTPolicyProcessor.h
```diff
@@ -60,4 +61,13 @@
...
                                        identifiers:(nonnull SNTRuleIdentifiers *)identifiers;
 
+
+///
+/// Updates a decision for a given file and agent configuration.
+///
+/// Returns YES if the decision requires no futher processing NO otherwise.
+- (BOOL) decision:(nonnull SNTCachedDecision *) cd 
+          forRule:(nonnull SNTRule *) rule
+          withTransitiveRules:(BOOL) transitive;
```
russellhancox, line 71:
```suggestion
- (BOOL)decision:(nonnull SNTCachedDecision *) cd 
                forRule:(nonnull SNTRule *) rule
    withTransitiveRules:(BOOL) transitive;
```

russellhancox: I'm not certain that clang-format will agree with the last line but there should be no space between the return type and method name

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -44,6 +45,118 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
   return self;
 }
 
+// This method applies the rules to the cached decision object.
+//
+// It returns YES if the decision was made, NO if the decision was not made.
+- (BOOL)decision:(SNTCachedDecision *)cd
+              forRule:(SNTRule *)rule
+  withTransitiveRules:(BOOL)enableTransitiveRules {
```
russellhancox, line 53:
```suggestion
- (BOOL)decision:(SNTCachedDecision *)cd
                forRule:(SNTRule *)rule
    withTransitiveRules:(BOOL)enableTransitiveRules {
```

russellhancox (changes requested):

russellhancox (approved):

russellhancox (approved):

mlw (approved):

PR #1312, opened 2024-03-20
Overrides disabled when running tests unless explicitly enabled

mlw:
This change disables applying the config overrides file when running tests by default. You can override this behavior by setting the environment variable `ENABLE_CONFIG_OVERRIDES`. E.g.:

```
bazel test --test_env=ENABLE_CONFIG_OVERRIDES=1 ...other test args...
```

### DIFF
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -1234,6 +1234,7 @@ static NSString *const kSyncTypeRequired = @"SyncTypeRequired";
     }
   }
 #ifdef DEBUG
+  BOOL overridesApplied = NO;
   NSDictionary *overrides = [NSDictionary dictionaryWithContentsOfFile:kConfigOverrideFilePath];
   for (NSString *key in overrides) {
     id obj = overrides[key];
@@ -1242,11 +1243,18 @@ static NSString *const kSyncTypeRequired = @"SyncTypeRequired";
          ![obj isKindOfClass:[NSString class]])) {
       continue;
     }
+
     forcedConfig[key] = obj;
+    overridesApplied = YES;
+
     if (self.forcedConfigKeyTypes[key] == [NSRegularExpression class]) {
       NSString *pattern = [obj isKindOfClass:[NSString class]] ? obj : nil;
       forcedConfig[key] = [self expressionForPattern:pattern];
     }
+  }
+
+  if (overridesApplied) {
+    NSLog(@"WARNING: Running with overrides applied!");
   }
 #endif
   return forcedConfig;
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -1227,7 +1227,7 @@ static NSString *const kSyncTypeRequired = @"SyncTypeRequired";
 #ifdef DEBUG
   if ([[[NSProcessInfo processInfo] processName] isEqualToString:@"xctest"] &&
       ![[[NSProcessInfo processInfo] environment] objectForKey:@"ENABLE_CONFIG_OVERRIDES"]) {
-    // By default, config overrides are not applied when runnings tests to help
+    // By default, config overrides are not applied when running tests to help
     // mitigate potential issues due to unexpected config values. This behavior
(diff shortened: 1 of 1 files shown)


Source/common/SNTConfigurator.m
```diff
@@ -1242,13 +1245,36 @@ - (NSMutableDictionary *)readForcedConfig {
...
+    overridesApplied = YES;
+
     if (self.forcedConfigKeyTypes[key] == [NSRegularExpression class]) {
       NSString *pattern = [obj isKindOfClass:[NSString class]] ? obj : nil;
       forcedConfig[key] = [self expressionForPattern:pattern];
     }
   }
+
+  if (overridesApplied) {
+    NSLog(@"WARNING: Running with overrides applied!");
```
pmarkowsky, line 1259: Should this be an asl_log like we do in SNTLogging?

Source/common/SNTConfigurator.m
```diff
@@ -1222,18 +1222,21 @@ - (NSRegularExpression *)expressionForPattern:(NSString *)pattern {
...
-    // Create the regex objects now
-    if (self.forcedConfigKeyTypes[key] == [NSRegularExpression class]) {
-      NSString *pattern = [obj isKindOfClass:[NSString class]] ? obj : nil;
-      forcedConfig[key] = [self expressionForPattern:pattern];
-    }
-  }
+- (void)applyOverrides:(NSMutableDictionary *)forcedConfig {
+  // Overrides should only be applied under debug builds.
 #ifdef DEBUG
+  if ([[[NSProcessInfo processInfo] environment] objectForKey:@"BAZEL_TEST"] &&
```
pmarkowsky, line 1228: Not sure we need to explicitly check for `BAZEL_TEST`, rather than just having this gated by `ENABLE_CONFIG_OVERRIDES`.

Just using `ENABLE_CONFIG_OVERRIDES` would mean that we'd need to do the following:

It would just mean that we'd need to do either:

* `bazel run --define ENABLE_CONFIG_OVERRIDES=1 //:reload`
* `bazel test --define ENABLE_CONFIG_OVVERRIDES=1 //test:target`

Both of which could be added to the .bazelrc file e.g.

```
# Maintains the old behavior
run:debug --define=ENABLE_CONFIG_OVERRIDES=1
# Gives us a new config to toggle this on / off
run:with_overrides --define ENABLE_CONFIG_OVVERRIDES=1
test:with_overrides --define ENABLE_CONFIG_OVVERRIDES=1
```

So we could do `bazel test --config=with_overrides //test:target`

Source/common/SNTConfigurator.m
mlw: Since we unfortunately can't include SNTLogging within the configurator as things are currently setup, and since this print is only for local debug builds and doesn't meed to conform to any structure, I went with the simple route here. Was there any sort of structure/data you were hoping for as is currently emitted by something like `LOGW`?

Source/common/SNTConfigurator.m
mlw: I'm not sure I like this approach as it requires re-compilation to toggle whichever mode you want. It also changes how local debug builds currently work by default which is to apply the overrides, without have another compilation or adding extra build flags.

The current impl has the benefit of not changing the existing workflow for local debug builds (enabling overrides) as well as doing the right thing by default (disabling overrides) when running tests. All without needing recompilation.

Source/common/SNTConfigurator.m
pmarkowsky: Yeah pretty much was looking to keep things uniform and include the warning flag, also wanted to make sure the logs turned up in the same place for debugging purposes.

mlw: As discussed, opting to just remove the log instead of recreating all of that logic just for debug builds.

Source/common/SNTConfigurator.m
pmarkowsky: Yeah as we talked more I agree here. This should be sufficent.

pmarkowsky (approved):
LGTM.

pmarkowsky (approved):
LGTM

PR #788, opened 2022-04-19
santa/windows: Update buttons to use push to better stand out

radsec:
TO: @russellhancox @tburgin 

### CHANGES:

Modifies the MessageWindow to match AboutWindow by using a `push` button type vs `roundTextured` button type to use the default button color scheme. This is to better guide users to engage with the sync/allowlist engine button link.

#### Presently

present-santa-block-modal

#### Proposed

santa-block-modal

PR #1316, opened 2024-03-26
[Bug] Restore `default button type` to `MessageWindow` for blocked events

radsec:
[Bug] Restore default button type to MessageWindow for blocked events

### CONTEXT
At some point in macOS 14 with the latest few agent revisions, the MessageWindow needs a hidden titlebar in order for the default button to show.

This is more of a stop-gap measure until this can be rewritten in SwiftUI, which I don't mind taking in the coming weeks to help contribute that change back. 
@russellhancox --> #1006 --> I can probs take the MessageWindow in the coming weeks.

For now, this UI bug can be quickly restored without much effort and decrease UX issues/confusion reported from our users.

### CHANGES
- Add hidden title bar to the MessageWindow box so the default button from this prior PR, #788, works again
- Update other SwiftUI buttons to inherit the same behavior
  - Keep UX uniformity with this change
### TESTING
- from latest `dev/main branch`
Screenshot 2024-03-25 at 15 28 48

### DIFF
--- a/Source/gui/Resources/MessageWindow.xib
+++ b/Source/gui/Resources/MessageWindow.xib
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="20037" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="32700.99.1234" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
     <dependencies>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="20037"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="22689"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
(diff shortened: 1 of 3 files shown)


Source/gui/SNTBinaryMessageWindowController.m
```diff
@@ -92,6 +92,8 @@ - (void)loadWindow {
     NSString *eventDetailText = [[SNTConfigurator configurator] eventDetailText];
     if (eventDetailText) {
       [self.openEventButton setTitle:eventDetailText];
+      [self.openEventButton setKeyEquivalent:@"\r"];
+      [self.openEventButton setKeyEquivalentModifierMask:!NSEventModifierFlagCommand];
```
russellhancox, line 96: We'd prefer if this _did_ require Command to be held. On the off-chance that a block dialog appears in the middle of someone typing, we don't want them to inadvertently dismiss the dialog without being able to see it. Cmd+Enter makes this significantly less likely.

russellhancox:
> This is more of a stop-gap measure until this can be rewritten in SwiftUI, which I don't mind taking in the coming weeks to help contribute that change back.

If you have the time and inclination, have at it, we'd appreciate it! I ran into issues trying to get all the various field and buttons to align properly with SwiftUI in a way that matches the current dialog (e.g. some fields aren't always populated and the "Open" button doesn't always exist, but we still want the UI to look consistent and centered in those cases) so I put it down and moved on to other things.

russellhancox (changes requested):

Source/gui/SNTBinaryMessageWindowController.m
radsec: Sounds good, I can incorporate that shortly! thanks for reviewing

Source/gui/Resources/MessageWindow.xib
```diff
@@ -241,16 +241,17 @@
...
-                        <constraints>
-                            <constraint firstAttribute="width" priority="900" constant="112" id="Pec-Pa-4aZ"/>
-                        </constraints>
                         <buttonCell key="cell" type="push" title="Open..." bezelStyle="rounded" alignment="center" state="on" borderStyle="border" imageScaling="proportionallyDown" inset="2" id="X1b-TF-1TL">
                             <behavior key="behavior" pushIn="YES" lightByBackground="YES" lightByGray="YES"/>
                             <font key="font" metaFont="system"/>
                             <string key="keyEquivalent" base64-UTF8="YES">
 DQ
 </string>
+                            <modifierMask key="keyEquivalentModifierMask" command="YES"/>
```
radsec, line 250: require the cmd + enter

Source/gui/SNTBinaryMessageWindowController.m
```diff
@@ -92,6 +92,9 @@ - (void)loadWindow {
     NSString *eventDetailText = [[SNTConfigurator configurator] eventDetailText];
     if (eventDetailText) {
       [self.openEventButton setTitle:eventDetailText];
+      // Require the button keyEquivalent set to be CMD + Return
+      [self.openEventButton setKeyEquivalent:@"\r"]; // Return Key
+      [self.openEventButton setKeyEquivalentModifierMask:NSCommandKeyMask]; // Command Key
```
radsec, line 97: require cmd + enter from PR feedback

russellhancox: Use `NSEventModifierFlagCommand`

radsec: gotcha...lols, really moving the button backwards with that. for ref: https://developer.apple.com/documentation/appkit/nseventmodifierflags/nseventmodifierflagcommand.

russellhancox (approved):

russellhancox:
Thanks for the contribution!

Issue #1318, opened 2024-04-05
CDHash serialization issue in fileinfo


np5:
There are issues with the CD Hash in the `santactl fileinfo` command output:

```
Path                   : /Applications/Firefox Developer Edition.app/Contents/MacOS/firefox
SHA-256                : e089ce663ef57a9be064f9ad8c49428930ce03083b19d5949460b14a5ce8571e
SHA-1                  : b9e82bdb5d90ec039df945d79a61c09a3fab9457
Bundle Name            : Firefox Developer Edition
Bundle Version         : 12524.4.3
Bundle Version Str     : 125.0
Download Timestamp     : 2024/04/04 07:02:21 +0200
Team ID                : 43AQ936H96
Signing ID             : org.mozilla.firefoxdeveloperedition
CDHash                 : {length = 20, bytes = 0xa71fd6b275762f87ad27f0f8cbe76961025370e3}
Type                   : Executable (arm64, x86_64)
Code-signed            : Yes
Rule                   : Could not communicate with daemon
```

It seems that the binary data is passed directly to the text output.

There is also a related hard crash in the json output:

```
/Applications/Santa.app/Contents/MacOS/santactl fileinfo --json /Applications/1Password.app
[
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'Invalid type in JSON write (__NSCFData)'
*** First throw call stack:
(
	0   CoreFoundation                      0x000000019a872ccc __exceptionPreprocess + 176
	1   libobjc.A.dylib                     0x000000019a35a788 objc_exception_throw + 60
	2   Foundation                          0x000000019b9734f4 _writeJSONValue + 788
	3   Foundation                          0x000000019b973e08 ___writeJSONObject_block_invoke + 388
	4   CoreFoundation                      0x000000019a7c0904 __NSDICTIONARY_IS_CALLING_OUT_TO_A_BLOCK__ + 24
	5   CoreFoundation                      0x000000019a7ffcc4 -[__NSDictionaryM enumerateKeysAndObjectsWithOptions:usingBlock:] + 288
	6   Foundation                          0x000000019b97372c _writeJSONObject + 504
	7   Foundation                          0x000000019c0f2e68 -[_NSJSONWriter dataWithRootObject:options:] + 88
	8   Foundation                          0x000000019b9730cc +[NSJSONSerialization dataWithJSONObject:options:error:] + 112
[...]

PR #1319, opened 2024-04-05
Fix CDHash value and formatting in santactl fileinfo

pmarkowsky:
This is a fix for #1318.


russellhancox:
Hold off on this, I'd rather get accessors into MOLCodesignChecker that we can use everywhere we need them

PR #1331, opened 2024-04-29
Grammar: Updated readme accordingly

Mpro256:
changed to It is named Santa because it keeps track of naughty or nice binaries.


russellhancox:
Both are grammatically correct and the original reads closer to the nursery rhyme.

Issue #1278, opened 2024-01-22
Big Sur Style Icon for Santa


legalizenukes:
Santa's current icon still adheres to the pre-Big Sur style and should be updated to match the new MacOS design language. I've made such an icon and have attached the .png file below.

I did my best to maintain the same style as the current icon but I can make changes as necessary to better suit the intended look.

The sled icon used is from [here](https://thenounproject.com/icon/sled-833974/).

![Santa](https://github.com/google/santa/assets/118522035/d49d7b21-f41a-47f8-a94b-ffb0546ef435)

PR #1334, opened 2024-04-30
Change App Icon to match Big Sur (and on) Design Theme

Mpro256:
Added icons to correct sizes.  

NOTE: The readme file and the docs use the santa 256px image which i changed to 1024px image. They both look the same to me.

This uses the icon designed by *A Deleted User* on issue #1278 .


Mpro256:
On github they say they are 1024x1024 but thats just a transparent backround so there is no effect.

Mpro256:
Still Waiting on response!

russellhancox:
Sorry for the delay in responding.

I'm uncomfortable approving this because the icon comes from a user that no longer exists and we have no way to ensure that this can be merged without copyright or licensing issues. The fact that the sled itself is creative-commons licensed does not mean that the whole icon is.

If you'd like to create an entirely new icon, we'd be open to reviewing an update as long as there are no issues with licensing and the new icon keeps with the original theme (bonus points if it takes the existing icon and just updates the icon style to match the OS better, instead of a full-on replacement)

PR #1335, opened 2024-05-01
Testing: update E2E to use JIT runners

kallsyms:
This PR updates the E2E test pipeline to register ephemeral just in time runners via the GitHub API and dynamically inject the runner token into the VM. This means VM images no longer have any keys inside of them, and they won't expire if unused.

This also splits the start_vm script into 2, one to update the image (if applicable) and the other to register a runner and start the VM.

### DIFF
--- a/Testing/integration/VM/setup.sh
+++ b/Testing/integration/VM/setup.sh
@@ -34,13 +34,13 @@ fi
 softwareupdate --install-rosetta --agree-to-license
 
 # Add a LaunchAgent to start the mounted runner
-cat | sudo tee ${HOME}/Library/LaunchAgents/runner.plist << EOF
+tee ${HOME}/Library/LaunchAgents/runner.plist << EOF
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
     <key>Label</key>
-    <string>runner</string>
+    <string>com.google.santa.e2erunner</string>
     <key>ProgramArguments</key>
     <array>
         <string>/Volumes/init/run.sh</string>
--- a/Testing/integration/VM/setup.sh
+++ b/Testing/integration/VM/setup.sh
@@ -33,16 +33,23 @@ fi
 # Install rosetta (for test binaries)
 softwareupdate --install-rosetta --agree-to-license
 
-# Install actions runner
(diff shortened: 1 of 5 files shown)


Testing/integration/VM/setup.sh
```diff
@@ -33,16 +33,23 @@ fi
...
-pushd ~/actions-runner
-curl -o actions-runner-osx-arm64-2.296.0.tar.gz -L https://github.com/actions/runner/releases/download/v2.296.0/actions-runner-osx-arm64-2.296.0.tar.gz
-echo 'e358086b924d2e8d8abf50beec57ee7a3bb0c7d412f13abc51380f1b1894d776  actions-runner-osx-arm64-2.296.0.tar.gz' | shasum -a 256 -c
-tar xzf ./actions-runner-osx-arm64-2.296.0.tar.gz
-./config.sh --url https://github.com/google/santa
-./svc.sh install
-./svc.sh start
-popd
+# Add a LaunchAgent to start the mounted runner
+cat | sudo tee ${HOME}/Library/LaunchAgents/runner.plist << EOF
```
russellhancox, line 37:
```suggestion
sudo tee ${HOME}/Library/LaunchAgents/runner.plist <<EOF
```

But also: why is sudo used? ~/Library/LaunchAgents is always writeable by the owning user.

Testing/integration/VM/setup.sh
```diff
@@ -33,16 +33,23 @@ fi
...
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>Label</key>
+    <string>runner</string>
```
russellhancox, line 43: There are odd cases where launchd will get upset about a Label that isn't a reverse-DNS label. Perhaps `com.google.santa.e2erunner`?

russellhancox (changes requested):

Testing/integration/VM/setup.sh
kallsyms: Was originally putting this in /System - leftover from that. Changed.

pmarkowsky (commented):
Overall looks good. 

Some nits.

Testing/integration/VM/setup.sh
kallsyms: Ah, good to know. Changed.

russellhancox (approved):

Issue #1336, opened 2024-05-03
Missing variable sequences in EventDetailURL


disenchant:
I might just be missing something but if I understand it correctly, `santad` will on a regular basis talk to the sync server and uploads the events in the "Event Upload" stage of the sync protocol to the server. Now if I want to allow a user to vote for the allow listing of a signing ID (e.g. `EQHXZ8M8AV:com.google.Chrome`) I can't really do that because the `EventDetailURL` configuration key does not allow for the signing ID to be included in the link. This would mean that unless there was already a sync after an application was blocked, I can't map e.g. the value of `%bundle_or_file_identifier%` to a signing ID in my logs and therefore only offer the user to allowlist a specific hash but not a signing ID as the mapping information would only be available to me after the next sync.

Is this actually the case or is there a way that I'm missing to include the signing ID (or at least team ID and bundle ID in two separate values) in the `EventDetailURL`?

If it's not possible I'd recommend that the [available sequences](https://santa.dev/deployment/configuration.html#eventdetailurl) which can be used in `EventDetailURL` get extended by at least the following:
* `team_id`
* `file_bundle_id`
* `signing_id` (should also cover `platform:...`)

And optional / nice-to-have would also be the following:

* `file_bundle_path`
* `file_name`
* `signing_chain` (format of this one TBD or maybe even split it into multiple sequences)

disenchant:
I don't have the dev setup ready to check it out right now but maybe just adding the following in [`Source/common/SNTBlockMessage.m:180`](https://github.com/google/santa/blob/main/Source/common/SNTBlockMessage.m#L180) would already be enough to address this:

```
      ^{ return event.teamID; },                 @"%team_id%",
      ^{ return event.fileBundleID; },           @"%file_bundle_id%",
      ^{ return event.signingID; },              @"%signing_id%",
```

russellhancox:
Need to document the additions in https://santa.dev/deployment/configuration.html#eventdetailurl

PR #1337, opened 2024-05-03
SNTBlockMessage: add more template options

tburgin:
Add the following template keys to SNTBlockMessage:
* `%file_bundle_id%`  The bundle id of the binary, if any.
* `%team_id%` The Team ID if present in the signature information.
* `%signing_id%` The Signing ID if present in the signature information.
* `%cdhash%`  If signed, the CDHash.

Also refactor `SNTFileAccessEvent` slightly. Make it a subclass of `SNTStoredEvent` to avoid redefining common properties.

### DIFF
--- a/Source/common/SNTBlockMessage.m
+++ b/Source/common/SNTBlockMessage.m
@@ -145,6 +145,10 @@
 //   %file_identifier%           - The SHA-256 of the binary being executed.
 //   %bundle_or_file_identifier% - The hash of the bundle containing this file or the file itself,
 //                                 if no bundle hash is present.
+//   %file_bundle_id%            - The bundle id of the binary, if any.
+//   %team_id%                   - The Team ID if present in the signature information.
+//   %signing_id%                - The Signing ID if present in the signature information.
+//   %cdhash%                    - If signed, the CDHash.
 //   %username%                  - The executing user's name.
 //   %machine_id%                - The configured machine ID for this host.
 //   %hostname%                  - The machine's FQDN.
@@ -177,6 +181,10 @@
       ^{ return event.fileSHA256 ? event.fileBundleHash ?: event.fileSHA256 : nil; },
                                                  @"%bundle_or_file_identifier%",
       ^{ return event.executingUser; },          @"%username%",
+      ^{ return event.fileBundleID; },           @"%file_bundle_id%",
+      ^{ return event.teamID; },                 @"%team_id%",
+      ^{ return event.signingID; },              @"%signing_id%",
+      ^{ return event.cdhash; },                 @"%cdhash%",
       ^{ return config.machineID; },             @"%machine_id%",
       ^{ return [SNTSystemInfo longHostname]; }, @"%hostname%",
       ^{ return [SNTSystemInfo hardwareUUID]; }, @"%uuid%",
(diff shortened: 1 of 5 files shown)


tburgin:
#1336

Source/common/SNTBlockMessage.m
```diff
@@ -177,6 +181,10 @@ + (NSURL *)eventDetailURLForEvent:(SNTStoredEvent *)event customURL:(NSString *)
       ^{ return event.fileSHA256 ? event.fileBundleHash ?: event.fileSHA256 : nil; },
                                                  @"%bundle_or_file_identifier%",
       ^{ return event.executingUser; },          @"%username%",
+      ^{ return event.fileBundleID; },           @"%file_bundle_id%",
```
russellhancox, line 184: I think it would be best to call this one `%bundle_identifier%` to match the pattern of `file_identifier` and `bundle_or_file_identifier`

Source/common/SNTBlockMessage.m
```diff
@@ -177,6 +181,10 @@ + (NSURL *)eventDetailURLForEvent:(SNTStoredEvent *)event customURL:(NSString *)
       ^{ return event.fileSHA256 ? event.fileBundleHash ?: event.fileSHA256 : nil; },
```
russellhancox, line 181: Can we use this opportunity to change this dictionary to be initialized with literal syntax so we don't have to read the keys/values backwards?

russellhancox (changes requested):

Source/common/SNTBlockMessage.m
tburgin: Done, simplified a bit too. I don't know why the values were blocks? I have replaced them with values.

@mlw Any thoughts on this?

russellhancox: IIRC they were blocks so that we didn't call all of the properties and store their values in the dictionary when most of the values would not be used.

tburgin: Oh, to handle the event.xxx nil case.

Source/common/SNTBlockMessage.m
russellhancox: Doesn't sound like there's a great option, we've overloaded identifier. The existing `_identifier` values should really be `_hash` but renaming them now is a pain.

Source/common/SNTBlockMessage.m
tburgin: > IIRC they were blocks so that we didn't call all of the properties and store their values in the dictionary when most of the values would not be used.

All the blocks are executed, I don't think its to save reads from the event. I think it's just a way to get around having a nil value in a dict.

russellhancox (dismissed):

Source/common/SNTBlockMessage.m
tburgin: Yeah don't think its worth it, this isn't a hot code path. I just went with NSNull.

mlw: As mentioned above, my recollection was just about not storing nil in the dictionary. The change to always store NSNull and then check for that value is fine too.

Source/common/SNTBlockMessage.m
```diff
@@ -123,35 +127,76 @@ + (NSString *)stringFromHTML:(NSString *)html {
...
+  }];
 
   return formatStr;
 }
 
-// Returns either the generated URL for the passed in event, or an NSURL from the passed in custom
-// URL string. If the custom URL string is the string "null", nil will be returned. If no custom
-// URL is passed and there is no configured EventDetailURL template, nil will be returned.
-// The following "format strings" will be replaced in the URL, if they are present:
 //
```
mlw, line 142: Nit: Bit of an awkward comment now with only the replacement strings listed. Maybe add a small description?

mlw (approved):

mlw (approved):

PR #1342, opened 2024-05-09
santad: Bump QoS of notify handling queue

russellhancox:
The use of the background queue is a historical artifact from when Santa had its own kernel extension with separate in-kernel queues for processing AUTH & NOTIFY type events. With the move to ES and the larger number of event types that we now notify on, running at the background QoS carries a small risk that the thread processing these events is not given a chance to run often enough that the queue grows and increases memory usage.

### DIFF
--- a/Source/santad/EventProviders/SNTEndpointSecurityClient.mm
+++ b/Source/santad/EventProviders/SNTEndpointSecurityClient.mm
@@ -91,8 +91,7 @@ constexpr std::string_view kProtectedFiles[] = {"/private/var/db/santa/rules.db"
 
     _notifyQueue = dispatch_queue_create(
       "com.google.santa.daemon.notify_queue",
-      dispatch_queue_attr_make_with_qos_class(DISPATCH_QUEUE_CONCURRENT_WITH_AUTORELEASE_POOL,
-                                              QOS_CLASS_DEFAULT, 0));
+      DISPATCH_QUEUE_CONCURRENT_WITH_AUTORELEASE_POOL);
   }
   return self;
 }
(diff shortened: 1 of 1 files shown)


pmarkowsky (approved):
LGTM.

Source/santad/EventProviders/SNTEndpointSecurityClient.mm
```diff
@@ -92,7 +92,7 @@ - (instancetype)initWithESAPI:(std::shared_ptr<EndpointSecurityAPI>)esApi
     _notifyQueue = dispatch_queue_create(
       "com.google.santa.daemon.notify_queue",
       dispatch_queue_attr_make_with_qos_class(DISPATCH_QUEUE_CONCURRENT_WITH_AUTORELEASE_POOL,
-                                              QOS_CLASS_BACKGROUND, 0));
+                                              QOS_CLASS_DEFAULT, 0));
```
mlw, line 95: Documentation states that `QOS_CLASS_DEFAULT` shouldn't be directly specified. We should change this instead to just call:
```
 _notifyQueue = dispatch_queue_create(
      "com.google.santa.daemon.notify_queue",
      DISPATCH_QUEUE_CONCURRENT_WITH_AUTORELEASE_POOL);
```

https://github.com/apple-oss-distributions/libpthread/blob/d8c4e3c212553d3e0f5d76bb7d45a8acd61302dc/include/sys/qos.h#L78-L86

mlw (changes requested):

Source/santad/EventProviders/SNTEndpointSecurityClient.mm
russellhancox: Done

mlw (approved):

mlw (approved):

PR #1348, opened 2024-05-15
Add metric for when the file on disk is not the file being evaluated

mlw:
Adds a new metric that tracks cases where the file reported in EXEC events is not the same file currently on disk.

Test setup:
Create some binary named `t`. In one terminal, run the following to rapidly change `t` devno/inode pair:
```
cp t t1
cp t t2
while [[ true ]]; do rm ./t; cp ./t1 ./t; sleep 0; rm ./t; cp ./t2 ./t; sleep 0; done
```

In a second terminal, continuously execute the binary:
```
while [[ true ]]; do ./t > /dev/null; done
```

Sample `santactl metrics` output:
```
  Metric Name               | /santa/event_stat_change_count
  Description               | Count of times a stat info changed for a binary being evalauted
  Type                      | SNTMetricTypeCounter
  Field                     | step=MessageCreate,error=0
  Created                   | 2024-05-15T00:47:07.050Z
  Last Updated              | 2024-05-15T00:47:07.050Z
  Data                      | 70
  Field                     | step=MessageCreate,error=2
  Created                   | 2024-05-15T00:47:07.050Z
  Last Updated              | 2024-05-15T00:47:07.050Z
  Data                      | 62
  Field                     | step=NoChange,error=0
  Created                   | 2024-05-14T23:47:10.405Z
  Last Updated              | 2024-05-15T00:47:07.050Z
  Data                      | 1252973
  ```

### DIFF
--- a/Source/santad/EventProviders/EndpointSecurity/Message.mm
+++ b/Source/santad/EventProviders/EndpointSecurity/Message.mm
@@ -16,7 +16,6 @@
 
 #include <bsm/libbsm.h>
 #include <libproc.h>
-#include <sys/errno.h>
 #include <sys/stat.h>
 
 #include "Source/santad/EventProviders/EndpointSecurity/EndpointSecurityAPI.h"
@@ -63,17 +62,13 @@ void Message::UpdateStatState(santa::santad::StatChangeStep step) const {
       es_msg_->event.exec.target && es_msg_->event.exec.target->executable) {
     struct stat &es_sb = es_msg_->event.exec.target->executable->stat;
     struct stat sb;
-    errno = 0;
     int ret = stat(es_msg_->event.exec.target->executable->path.data, &sb);
     // If stat failed, or if devno/inode changed, update state.
     if (ret != 0 || es_sb.st_ino != sb.st_ino || es_sb.st_dev != sb.st_dev) {
       stat_change_step_ = step;
       // Determine the specific condition that failed for tracking purposes
-      if (ret != 0) {
-        stat_result_ = santa::santad::StatResult::kStatError;
-      } else {
-        stat_result_ = santa::santad::StatResult::kDevnoInodeMismatch;
-      }
+      stat_result_ = (ret != 0) ? santa::santad::StatResult::kStatError
+                                : santa::santad::StatResult::kDevnoInodeMismatch;
     }
   }
 }
--- a/Source/santad/EventProviders/EndpointSecurity/Message.mm
+++ b/Source/santad/EventProviders/EndpointSecurity/Message.mm
@@ -16,6 +16,8 @@
 
 #include <bsm/libbsm.h>
 #include <libproc.h>
+#include <sys/errno.h>
+#include <sys/stat.h>
 
 #include "Source/santad/EventProviders/EndpointSecurity/EndpointSecurityAPI.h"
 
@@ -24,6 +26,7 @@ namespace santa::santad::event_providers::endpoint_security {
 Message::Message(std::shared_ptr<EndpointSecurityAPI> esapi, const es_message_t *es_msg)
     : esapi_(std::move(esapi)), es_msg_(es_msg), process_token_(std::nullopt) {
   esapi_->RetainMessage(es_msg);
+  UpdateStatState(santa::santad::StatChangeStep::kMessageCreate);
 }
 
 Message::~Message() {
@@ -38,6 +41,8 @@ Message::Message(Message &&other) {
   other.es_msg_ = nullptr;
   process_token_ = std::move(other.process_token_);
   other.process_token_ = std::nullopt;
+  stat_change_step_ = other.stat_change_step_;
+  stat_error_ = other.stat_error_;
 }
 
 Message::Message(const Message &other) {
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: Source/santad/SNTPolicyProcessor.h [hunk 1]
```diff
@@ -51,6 +51,13 @@
                           (NSDictionary *_Nullable (^_Nonnull)(
                             const char *_Nullable teamID,
                             NSDictionary *_Nullable entitlements))entitlementsFilterCallback;
+- (nonnull SNTCachedDecision *)decisionForFileInfo:(nonnull SNTFileInfo *)fileInfo
+                                     targetProcess:(nonnull const es_process_t *)targetProc
+                          preCodesignCheckCallback:(void (^_Nullable)(void))preCodesignCheckCallback
```
tburgin, line 56: Adding an additional callback here is starting to get unwieldy as an interface. Consider refactoring these callbacks out into a delegate protocol. Then have a new concrete object that implements the protocol to handle the callbacks.

Source/santad/EventProviders/EndpointSecurity/Message.mm
```diff
@@ -38,13 +41,36 @@
...
+      // partially construct an es_message_t.
+      es_msg_->event.exec.target && es_msg_->event.exec.target->executable) {
+    struct stat &es_sb = es_msg_->event.exec.target->executable->stat;
+    struct stat sb;
+    errno = 0;
+    int ret = stat(es_msg_->event.exec.target->executable->path.data, &sb);
+    // If stat failed, or if devno/inode changed, update state.
+    if (ret != 0 || es_sb.st_ino != sb.st_ino || es_sb.st_dev != sb.st_dev) {
+      stat_change_step_ = step;
+      stat_error_ = errno;
```
tburgin, line 71: It would be helpful to tell a part a stat error, inode mismatch and dev mismatch. Currently error is 0 for both inode and dev mismatches. We don't want to overload errno with additional meaning. I can't think of a good approach besides splitting up these kinds of failures or adding a new dimension to the metric. Both are not great.

How concerned are we about stat failing? Maybe just don't log a metric if stat fails and only log mismatches, which is what we really care about?

Source/santad/SNTPolicyProcessor.h
mlw: This would be a pretty large refactor. Reckon i'm overall not a huge fan of the delegate pattern, but it's a fair point about multiple callbacks becoming unwieldy. Part of the problem is there are multiple paths into the PolicyProcessor. One from the main AUTH EXEC processing flow, another via XPC/santactl. Obviously delegate could handle this as well by jsut not setting the delegate property via the XPC route. But that feels odd too. Having the interfaces into the PolicyProcessor makes it more clear (to me) about a caller's intention, without having to search around for when/how random delegates get set.

I'm inclined to punt on the major refactor to delegates, but I'm open to other ideas of how to simplify this.

Source/santad/EventProviders/EndpointSecurity/Message.mm
mlw: My hunch would be that the biggest reason stat fails would be due to the file being deleted. This feels like that would be interesting information to capture..

But I could see grouping all stat errors together. Then we could change `stat_error_` to an enum type. Something like:

```
enum class StatResult {
  kOK,
  kStatError,
  kInodeDevnoMismatch,
}
```

So then it would just boil down to how much would we care to know the underlying error if we see a lot of them...

Source/santad/SNTPolicyProcessor.h
tburgin: At its core the ObjC delegate pattern isn't terrible in my opinion. It is ubiquitous and pretty well understood. But if we were to refactor maybe we punt and refactor in C++. There are similar "delegate/observer" patterns we could use.

https://chromium.googlesource.com/chromium/src/+/HEAD/docs/patterns/inversion-of-control.md

Source/santad/EventProviders/EndpointSecurity/Message.mm
tburgin: StatResult sounds good to me

Source/santad/EventProviders/EndpointSecurity/Message.mm
```diff
@@ -38,13 +41,41 @@
...
+void Message::UpdateStatState(santa::santad::StatChangeStep step) const {
+  // Only update state for AUTH EXEC events and if no previous change was detected
+  if (es_msg_->event_type == ES_EVENT_TYPE_AUTH_EXEC &&
+      stat_change_step_ == santa::santad::StatChangeStep::kNoChange &&
+      // Note: The following checks are required due to tests that only
+      // partially construct an es_message_t.
+      es_msg_->event.exec.target && es_msg_->event.exec.target->executable) {
+    struct stat &es_sb = es_msg_->event.exec.target->executable->stat;
+    struct stat sb;
+    errno = 0;
```
tburgin, line 66: Not used anymore

Source/santad/EventProviders/EndpointSecurity/Message.mm
```diff
@@ -16,6 +16,8 @@
 
 #include <bsm/libbsm.h>
 #include <libproc.h>
+#include <sys/errno.h>
```
tburgin, line 19: Not used anymore

Source/santad/EventProviders/EndpointSecurity/Message.mm
```diff
@@ -38,13 +41,41 @@
...
+      es_msg_->event.exec.target && es_msg_->event.exec.target->executable) {
+    struct stat &es_sb = es_msg_->event.exec.target->executable->stat;
+    struct stat sb;
+    errno = 0;
+    int ret = stat(es_msg_->event.exec.target->executable->path.data, &sb);
+    // If stat failed, or if devno/inode changed, update state.
+    if (ret != 0 || es_sb.st_ino != sb.st_ino || es_sb.st_dev != sb.st_dev) {
+      stat_change_step_ = step;
+      // Determine the specific condition that failed for tracking purposes
+      if (ret != 0) {
```
tburgin, line 72: Nit, if you like:

```
stat_result_ = (ret != 0) ? santa::santad::StatResult::kStatError
                          : santa::santad::StatResult::kDevnoInodeMismatch;
```

tburgin (approved):

tburgin (approved):

PR #1353, opened 2024-05-17
Add cdhash, teamID, signingID to the bundle events

### LINKED ISSUE #1352: Add Signing ID and Team ID to the Bundle Binary events
Since signing ID & team ID have been added to MOLCodesignChecker, would you consider adding them to the [Bundle Binary pseudo events](https://github.com/google/santa/blob/67883c5200ee22410086cbf0581be7c1c96d50d8/Source/santabundleservice/SNTBundleService.m#L227-L229) to send them to the sync server ? This way we could build signing ID rules for a whole bundle when approved.

np5:
Fix #1352

### DIFF
--- a/Source/santabundleservice/SNTBundleService.m
+++ b/Source/santabundleservice/SNTBundleService.m
@@ -226,6 +226,9 @@
 
       MOLCodesignChecker *cs = [fi codesignCheckerWithError:NULL];
       se.signingChain = cs.certificates;
+      se.teamID = cs.teamID;
+      se.signingID = cs.signingID;
+      se.cdhash = cs.cdhash;
 
       dispatch_sync(dispatch_get_main_queue(), ^{
         relatedEvents[se.fileSHA256] = se;
--- a/Source/santabundleservice/SNTBundleService.m
+++ b/Source/santabundleservice/SNTBundleService.m
@@ -226,9 +226,17 @@
 
       MOLCodesignChecker *cs = [fi codesignCheckerWithError:NULL];
       se.signingChain = cs.certificates;
-      se.teamID = cs.teamID;
-      se.signingID = cs.signingID;
       se.cdhash = cs.cdhash;
+      se.teamID = cs.teamID;
+      if (cs.signingID) {
+        if (cs.teamID) {
+          se.signingID = [NSString stringWithFormat:@"%@:%@", cs.teamID, cs.signingID];
+        } else {
+          if (cs.platformBinary) {
+            se.signingID = [NSString stringWithFormat:@"platform:%@", cs.signingID];
+          }
+        }
+      }
 
       dispatch_sync(dispatch_get_main_queue(), ^{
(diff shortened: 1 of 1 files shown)


Source/santabundleservice/SNTBundleService.m
```diff
@@ -226,6 +226,9 @@ - (NSDictionary *)generateEventsFromBinaries:(NSArray *)fis
 
       MOLCodesignChecker *cs = [fi codesignCheckerWithError:NULL];
       se.signingChain = cs.certificates;
+      se.teamID = cs.teamID;
+      se.signingID = cs.signingID;
```
pmarkowsky, line 230: You need to prefix the Signing ID with the Team ID e.g. `7Z6EMTD2C6:org.wireshark.Wireshark` and handle if this is a platform binary to be consistent with the rest of the codebase and the sync protocol docs.

If you look at MOLCodeSignChecker you'll see that this just returns [the raw signing ID](https://github.com/google/macops-molcodesignchecker/blob/5060bcc8baa90bae3b0ca705d14850328bbbec53/Source/MOLCodesignChecker/MOLCodesignChecker.m#L287-L289). 

```objc
- (NSString *)signingID {
  return self.signingInformation[(__bridge id)kSecCodeInfoIdentifier];
}
```

You need to format it like so:

https://github.com/google/santa/blob/9b184ed4fb9e4e720699b905f7b04f5c7b5cf018/Source/santactl/Commands/SNTCommandFileInfo.m#L385-L396

pmarkowsky (changes requested):
You need to fix the signingID formatting.

Source/santabundleservice/SNTBundleService.m
np5: I have posted a fix.

Source/santabundleservice/SNTBundleService.m
```diff
@@ -226,6 +226,17 @@ - (NSDictionary *)generateEventsFromBinaries:(NSArray *)fis
...
       se.signingChain = cs.certificates;
+      se.cdhash = cs.cdhash;
+      se.teamID = cs.teamID;
+      if (cs.signingID) {
+        if (cs.teamID) {
+          se.signingID = [NSString stringWithFormat:@"%@:%@", cs.teamID, cs.signingID];
+        } else {
+          if (cs.platformBinary) {
+            se.signingID = [NSString stringWithFormat:@"platform:%@", cs.signingID];
+          }
```
russellhancox, line 237:
```suggestion
        } else if (cs.platformBinary) {
          se.signingID = [NSString stringWithFormat:@"platform:%@", cs.signingID];
        }
```

russellhancox (changes requested):

Source/santabundleservice/SNTBundleService.m
np5: I have included this change in the last commit

PR #1358, opened 2024-05-21
Fix NSSecureCoding adoption in SNTFileAccessEvent

mlw:
This prevented the FAA dialog popups from displaying.

Caused by: #1337


tburgin (approved):
ooops, thanks!

PR #1359, opened 2024-05-28
sync: Add a protobuf for the existing sync protocol

russellhancox:
This PR is intended to have no impact on existing sync servers. The fields and enum values in the protobuf have been named such that their JSON equivalents match the existing constants we have in the codebase.

Adding this provides a few benefits:

1) The protobuf serves as canonical documentation of the protocol in a form that's much easier to read than the existing code.
2) Protobuf parsing of JSON is likely to be better than our hand-written version.
3) We can (in a later PR) add a configuration option to use binary encoding instead of JSON, saving network during syncs.
4) Servers written in other languages are easier to write and update as time goes on, especially as we extend the protocol.


tburgin:
Oh all but SNTSyncEventUpload are diff'd, it would be good to get that one too.

Source/santasyncservice/syncv1.proto
```diff
@@ -0,0 +1,225 @@
...
+  optional string allowed_path_regex = 11;
+  optional string blocked_path_regex = 12;
+
+  optional bool block_usb_mount = 13;
+  repeated string remount_usb_mode = 14;
+
+  optional string override_file_access_action = 15;
+
+  // These fields are deprecated forms of other fields and exist here solely for backwards compatibility
+  optional bool deprecated_enabled_transitive_whitelisting = 80 [json_name="enabled_transitive_whitelisting"];
```
tburgin, line 83: What about using the `[deprecated=true]` tag instead of the `deprecated_` prefix in the name?

tburgin, line 83: Starting at 80 feels a bit arbitrary, any reason for picking it? I can't find any guidance around creating a new enum with already deprecated fields.

I guess the other options are start the deprecated fields at 1 or interleave them next to the non-deprecated fields? To me the later feels the most natural, however it clutters up  the non-deprecated fields.

pmarkowsky: Agreed. Usually in the past I've worked on teams where we've made deprecated fields start at 1,000 or 2,000. 

Not only does it give you a lot of wiggle room it also makes it really clear that 4 digits means do not use.

pmarkowsky:
This is awesome. However given how strict proto3's JSON parsing can be. 

We should be really careful to make sure we're backwards compatible.

russellhancox:
> This is awesome. However given how strict proto3's JSON parsing can be.
> 
> We should be really careful to make sure we're backwards compatible.

All the existing unit tests pass and I've tested a bunch with our internal sync server. Given how strict our hand-written parser was, this _should_ work generally with other servers. We'll also get Moroz testing once this is merged and I'm planning to call this change out in the release notes as something to be very aware of.

Source/santasyncservice/SNTSyncEventUpload.mm
```diff
@@ -0,0 +1,183 @@
...
+#import "Source/common/String.h"
+#import "Source/santasyncservice/NSData+Zlib.h"
+#import "Source/santasyncservice/SNTSyncLogging.h"
+#import "Source/santasyncservice/SNTSyncState.h"
+
+#include <google/protobuf/arena.h>
+#include "Source/santasyncservice/syncv1.pb.h"
+namespace pbv1 = ::santa::sync::v1;
+
+using santa::common::NSStringToUTF8String;
```
tburgin, line 36: nit: avoid using `using`

https://google.github.io/styleguide/cppguide.html#Namespaces

Source/santasyncservice/SNTSyncEventUpload.mm
```diff
@@ -0,0 +1,183 @@
...
+    }
+    dispatch_semaphore_signal(sema);
+  }];
+  return (dispatch_semaphore_wait(sema, DISPATCH_TIME_FOREVER) == 0);
+}
+
+- (BOOL)uploadEvents:(NSArray *)events {
+  google::protobuf::Arena arena;
+  auto req = google::protobuf::Arena::Create<::pbv1::EventUploadRequest>(&arena);
+  auto uploadEvents = req->mutable_events();
```
tburgin, line 59: Avoid `auto` where it isn't readily apparent what the type is.

https://google.github.io/styleguide/cppguide.html#Type_deduction

I would argue that we should not use `auto` for these getters. I would like to see what types are coming out them, especially during review.

Source/santasyncservice/SNTSyncEventUpload.mm
russellhancox: using-directives are banned, using-declarations are allowed.

Source/santasyncservice/SNTSyncPreflight.mm
```diff
@@ -115,59 +118,76 @@ - (BOOL)sync {
...
+  auto value = resp.push_notification_full_sync_interval();
+  if (value == 0) {
+    value = resp.deprecated_fcm_full_sync_interval();
+  }
   self.syncState.pushNotificationsFullSyncInterval =
     (value < kDefaultFullSyncInterval) ? kDefaultPushNotificationsFullSyncInterval : value;
 
-  value = [EnsureType(resp[kFCMGlobalRuleSyncDeadline], [NSNumber class]) unsignedIntegerValue];
+  value = resp.push_notification_global_rule_sync_deadline();
+  if (value == 0) {
```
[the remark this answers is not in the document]
russellhancox: Done

Source/santasyncservice/SNTSyncPreflight.mm
```diff
@@ -115,59 +118,76 @@ - (BOOL)sync {
...
+      resp.deprecated_transitive_whitelisting_enabled());
+  self.syncState.enableAllEventUpload = @(resp.enable_all_event_upload());
+  self.syncState.disableUnknownEventUpload = @(resp.disable_unknown_event_upload());
+  self.syncState.eventBatchSize = resp.batch_size();
 
   // Don't let these go too low
-  NSUInteger value =
-    [EnsureType(resp[kFCMFullSyncInterval], [NSNumber class]) unsignedIntegerValue];
+  auto value = resp.push_notification_full_sync_interval();
+  if (value == 0) {
```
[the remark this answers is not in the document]
russellhancox: Done. Much better

Source/santasyncservice/SNTSyncEventUpload.mm
tburgin: Yeah I agree.

tburgin (approved):

tburgin (approved):

Issue #1235, opened 2023-11-15
santa daemon should have a mechanism to prevent being killed by users with root privileges.


izzh:
If I have root privileges, I can bypass the allowlist mechanism by repeatedly killing santa daemon in a loop.

```
#!bin/bash

while true; do
    ps aux | grep com.google.santa.daemon | grep -v grep | awk '{print $2}' | xargs kill -9
    if [ $? -eq 0 ]; then
        echo "kill santa daemon"
    fi
    sleep 0.5
done
```
```
santa@macos-13 Desktop % ./process_not_in_allowlist
### Before run the kill script
Santa

This application has been blocked from executing.

Path:       /Users/santa/Desktop/process_not_in_allowlist
Identifier: 0b742eae49b08af3858f17e95aa80afa82a18a178a671fc23256f12d58286894
Parent:     zsh (2021)

More info:
https://santa/blockables/E6C3C62D-8A40-5B56-99CA-191EACBCE8FD/0b742eae49b08af3858f17e95aa80afa82a18a178a671fc23256f12d58286894

### After run the kill script
santa@macos-13 Desktop % ./process_not_in_allowlist
malicious
^C
santa@macos-13 Desktop % ./process_not_in_allowlist
malicious
^C
santa@macos-13 Desktop % ./process_not_in_allowlist
malicious
^C
santa@macos-13 Desktop % 
```

I have observed other security software that cannot be killed even if I have root privileges.
```
santa@macos-13 Desktop % ps aux |grep falcon
root               375   5.5  0.7 36057876 228704   ??  Rs   Mon04PM 124:01.29 /Library/SystemExtensions/EA9DEA93-1AAA-4A86-9DC4-9CA95609D798/com.crowdstrike.falcon.Agent.systemextension/Contents/MacOS/com.crowdstrike.falcon.Agent
santa@macos-13 Desktop % sudo kill -9 375
Password:
kill: 375: Operation not permitted
santa@macos-13 Desktop ~ % 
```
Is it possible to implement a feature to control whether santa daemon can be killed through a configuration field? thanks～

mlw:
Santa currently has some basic tamper resistance features, but we should expand upon these, including better signal handling. It is something we'd like to get to eventually.

[...]

PR #1360, opened 2024-05-28
santad: Add signal auth to tamper resistence.

russellhancox:
Root users can currently kill the `com.google.santa.daemon` process. It will be immediately restarted by sysextd but this opens a very brief window where protection is lost. Hooking `ES_EVENT_TYPE_AUTH_SIGNAL` and blocking all signals to the santad process prevents this and it doesn't interfere with upgrades.

### DIFF
--- a/Source/santad/EventProviders/SNTEndpointSecurityTamperResistance.mm
+++ b/Source/santad/EventProviders/SNTEndpointSecurityTamperResistance.mm
@@ -14,9 +14,10 @@
 
 #import "Source/santad/EventProviders/SNTEndpointSecurityTamperResistance.h"
 
-#include <bsm/libbsm.h>
 #include <EndpointSecurity/ESTypes.h>
+#include <bsm/libbsm.h>
 #include <string.h>
+#include <algorithm>
 
 #import "Source/common/SNTLogging.h"
 #include "Source/santad/DataLayer/WatchItemPolicy.h"
@@ -88,13 +89,11 @@ static constexpr std::string_view kSantaKextIdentifier = "com.google.santa-drive
 
     case ES_EVENT_TYPE_AUTH_SIGNAL: {
       // Only block signals sent to us and not from launchd.
-      if (audit_token_to_pid(esMsg->event.signal.target->audit_token) != getpid() ||
-          audit_token_to_pid(esMsg->process->audit_token) == 1) {
-        result = ES_AUTH_RESULT_ALLOW;
-        break;
+      if (audit_token_to_pid(esMsg->event.signal.target->audit_token) == getpid() &&
+          audit_token_to_pid(esMsg->process->audit_token) != 1) {
+        LOGW(@"Preventing attempt to kill Santa daemon");
+        result = ES_AUTH_RESULT_DENY;
       }
-      LOGW(@"Preventing attempt to kill Santa daemon");
-      result = ES_AUTH_RESULT_DENY;
       break;
     }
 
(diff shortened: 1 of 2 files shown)


Source/santad/EventProviders/SNTEndpointSecurityTamperResistance.mm
```diff
@@ -85,6 +85,16 @@ - (void)handleMessage:(Message &&)esMsg
...
     }
 
+    case ES_EVENT_TYPE_AUTH_SIGNAL: {
+      if (esMsg->event.signal.target->session_id != getpid()) {
+        result = ES_AUTH_RESULT_ALLOW;
+        break;
+      }
+      LOGW(@"Preventing attempt to kill Santa daemon");
+      result = ES_AUTH_RESULT_DENY;
+      break;
```
pmarkowsky, line 95: Probably need a check to see if the signal's coming from launchd. It's not perfect but otherwise won't this prevent us from doing easy upgrades.

russellhancox: Upgrades seem to work fine as-is (the unload is handled by the sysext framework) but this does cause a delay during reboot/shutdown, which I'm investigating workarounds for.

mlw:
#1235

Source/santad/EventProviders/SNTEndpointSecurityTamperResistance.mm
russellhancox: Updated approach so this no longer applies.

Source/santad/EventProviders/SNTEndpointSecurityTamperResistance.mm
```diff
@@ -85,6 +86,51 @@ - (void)handleMessage:(Message &&)esMsg
...
+    }
+
+    case ES_EVENT_TYPE_AUTH_EXEC: {
+      // When not running a debug build, prevent attempts to kill Santa
+      // by launchctl commands.
```
mlw, line 103: We chatted briefly about this yesterday, but here's a quick sorta-pseudo-code approach I was mentioning. I don't expect this to be a hot path at all, but inverting these checks will reduce the number of string comparisons done:

```
es_auth_result_t ValidateLaunchctl(const Message &esMsg) {
  es_string_token_t exec_path = esMsg->event.exec.target->executable->path;
  if (strncmp(exec_path.data, "/bin/launchctl", exec_path.length) != 0) {
    return ES_AUTH_RESULT_ALLOW;
  }

  // Ensure at least 1 argument is provided to the command
  std::shared_ptr<EndpointSecurityAPI> esApi = esMsg.ESAPI();
  uint32_t argCount = esApi->ExecArgCount(&esMsg->event.exec);
  if (argCount < 2) {
    return ES_AUTH_RESULT_ALLOW;
  }

  // Check for some allowed subcommands
  es_string_token_t arg = esApi->ExecArg(&esMsg->event.exec, 1);
  if (arg.data in ["list" || "print"]) {
    return ES_AUTH_RESULT_ALLOW;
  }

  // Check whether com.google.santa.daemon is in the argument list.
  // launchctl no longer accepts PIDs to operate on.
  for (int i = 2; i < argCount; i++) {
    es_string_token_t arg = esApi->ExecArg(&esMsg->event.exec, i);
    if (strnstr(arg.data, "com.google.santa.daemon", arg.length) != NULL) {
      return ES_AUTH_RESULT_DENY;
    }
  }

  return ES_AUTH_RESULT_ALLOW;
}
```

pmarkowsky: To make sure I'm reading this right at a high level the flow is:

```mermaid
graph TD
A[Execution ESMsg is received] --> B{Is the command /bin/launchctl?}
B --> |No| X[Return ES_AUTH_RESULT_ALLOW]
B --> |Yes| C{Does this have at least 1 argument?}
C --> |No|X
C --> |Yes|D{Is the first argument\n a safe subcommand?}
D --> |Yes|X
D --> |No|E{Is the command targeting\n the Santa daemon \naka com.google.santa.daemon?}
E --> |Yes|Y[Return ES_AUTH_RESULT_DENY]
E --> |No|X
```

I imagine that rather than checking the path in the binary we'd use the Signing ID so that someone can't copy the binary or other trickery.

mlw: Flow LGTM.

> I imagine that rather than checking the path in the binary we'd use the Signing ID so that someone can't copy the binary or other trickery.

Fine with checking "platform binary + SID", but also it shouldn't much matter either way. You can't hard link this binary, symlinks would resolve to the real path in ES, and launch constraints ___should___ prevent any other copy trickery.

Source/santad/EventProviders/SNTEndpointSecurityTamperResistance.mm
```diff
@@ -120,15 +142,54 @@ - (void)enable {
...
+  uint32_t argCount = esApi->ExecArgCount(&esMsg->event.exec);
+  if (argCount < 2) {
+    return ES_AUTH_RESULT_ALLOW;
+  }
+
+  // Check for some allowed subcommands
+  es_string_token_t arg = esApi->ExecArg(&esMsg->event.exec, 1);
+  static const std::unordered_set<std::string> safe_commands{
+    "print",
+    "procinfo",
```
mlw, line 177: Need to add the `list` subcommand.

Source/santad/EventProviders/SNTEndpointSecurityTamperResistance.mm
pmarkowsky: >  launch constraints should prevent any other copy trickery.

Yeah, tbh I'm not entirely sure what the limits are with launch constraints but given that we have a strong primitive like code signing seems like an easy way to be sure.

russellhancox: Using the signing ID would mean unmuting all paths, which makes this _much_ more expensive. Given the limitations on copying/linking launchctl, I think this is good enough as-is.

Source/santad/EventProviders/SNTEndpointSecurityTamperResistance.mm
russellhancox: I've added a few of those. All of them will ultimately be allowed if santa isn't referenced so adding them to the safe list just skips the arg lookup and speeds things up a very minor amount.

mlw (approved):

PR #1361, opened 2024-05-30
docs: Document new EventDetailURL keys

### LINKED ISSUE #1336: Missing variable sequences in EventDetailURL

russellhancox:
Fixes #1336


russellhancox:
The markdown-check failure is wrong: it's reporting that https://bazel.build/ is not valid.

mlw (dismissed):

docs/development/sync-protocol.md
```diff
@@ -390,7 +390,7 @@ downloading if the rules need to be downloaded in multiple batches.
 | policy | YES | string | Identifies the action to perform in response to the rule matching (must be one of the examples) | "ALLOWLIST","ALLOWLIST_COMPILER", "BLOCKLIST", "REMOVE",  "SILENT_BLOCKLIST" |
 | rule\_type | YES | string | Identifies the type of rule (must be one of the examples) | "BINARY", "CERTIFICATE", "SIGNINGID", "TEAMID", "CDHASH"  |
 | custom\_msg | NO | string | A custom message to display when the rule matches | "Hello" |
-| custom\_url | NO | string | A custom URL to use for the open button when the rule matches | http://lmgtfy.app/?q=dont+download+malware |
+| custom\_url | NO | string | A custom URL to use for the open button when the rule matches | http://example.com |
```
pmarkowsky, line 393: You could also use https://letmegooglethat.com/?q=don%27t+download+malware

russellhancox: I could, but this is less snarky

mlw (approved):

PR #1362, opened 2024-05-30
Update check-markdown workflow to only check modified files 

pmarkowsky:
* Update check-markdown workflow to only check modified files 
* Ignore bazel.build

### DIFF
--- a/docs/development/building.md
+++ b/docs/development/building.md
@@ -5,6 +5,7 @@ parent: Development
 
 # Building
 
+<!-- markdown-link-check-disable-next-line -->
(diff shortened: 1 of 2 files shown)


russellhancox (approved):

docs/development/building.md
```diff
@@ -5,6 +5,7 @@ parent: Development
 
 # Building
 
+<!-- markdown-link-check-disable-next-line -->
```
russellhancox, line 8: With the switch to Lychee, is this still needed?

russellhancox (approved):

PR #1364, opened 2024-06-03
sync: Add option to sync using binary protos

russellhancox:


Source/common/SNTConfigurator.m
mlw, line 1: Should we go ahead and setup the KVO handling for this key to handle changes?

russellhancox: Done

mlw (approved):

PR #1365, opened 2024-06-10
Add a Signing ID Format Helper

pmarkowsky:
This PR adds a helper function to format the Signing IDs in Santa and to ensure they're always handled and formatted the same way.

It's marked draft until we're sure that it's used everywhere it should be and that folks are happy with it.

### DIFF
--- a/Source/common/SigningIDHelpers.m
+++ b/Source/common/SigningIDHelpers.m
@@ -16,7 +16,7 @@
 #import "Source/common/SNTLogging.h"
 
 NSString *FormatSigningID(MOLCodesignChecker *csc) {
-  if (csc == nil || !csc.signingID) {
+  if (!csc.signingID.length) {
     LOGD(@"unable to format signing ID as it's missing");
     return nil;
   }
--- a/Source/common/SigningIDHelpers.m
+++ b/Source/common/SigningIDHelpers.m
@@ -21,7 +21,7 @@ NSString *FormatSigningID(MOLCodesignChecker *csc) {
     return nil;
   }
 
-  if (csc.teamID == nil) {
+  if (!csc.teamID.length) {
     if (csc.platformBinary) {
       return [NSString stringWithFormat:@"%@:%@", @"platform", csc.signingID];
     } else {
--- a/Source/common/SigningIDHelpers.m
+++ b/Source/common/SigningIDHelpers.m
@@ -1,0 +1,31 @@
+/// Copyright 2024 Google LLC
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///     https://www.apache.org/licenses/LICENSE-2.0
+///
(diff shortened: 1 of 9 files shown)


Source/common/SigningIDHelpers.m
```diff
@@ -0,0 +1,31 @@
...
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+#import "Source/common/SigningIDHelpers.h"
+#import "Source/common/SNTLogging.h"
+
+NSString *FormatSigningID(MOLCodesignChecker *csc) {
+  if (csc.signingID) {
```
russellhancox, line 19: I'd invert this and return early to reduce indentation

pmarkowsky: Agreed.

Source/common/SigningIDHelpers.m
```diff
@@ -0,0 +1,34 @@
...
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+#import "Source/common/SigningIDHelpers.h"
+#import "Source/common/SNTLogging.h"
+
+NSString *FormatSigningID(MOLCodesignChecker *csc) {
+  if (csc == nil || !csc.signingID) {
```
russellhancox, line 19: Accessing the signingID property of csc is safe even if csc is nil (it will just return nil) and checking the length covers both the case of signingID being nil or it being an empty string

```suggestion
  if (!csc.signingID.length) {
```

mlw (commented):
I think it would be more powerful if things like SID were made into their own type. Having a helper is nice, but there's no format enforcement mechanism by keeping it a string.

Source/common/SigningIDHelpers.m
```diff
@@ -0,0 +1,34 @@
...
+#import "Source/common/SigningIDHelpers.h"
+#import "Source/common/SNTLogging.h"
+
+NSString *FormatSigningID(MOLCodesignChecker *csc) {
+  if (!csc.signingID.length) {
+    LOGD(@"unable to format signing ID as it's missing");
+    return nil;
+  }
+
+  if (csc.teamID == nil) {
```
russellhancox, line 24:
```suggestion
  if (!csc.teamID.length) {
```

russellhancox:
> I think it would be more powerful if things like SID were made into their own type. Having a helper is nice, but there's no format enforcement mechanism by keeping it a string.

Agree in principle but they'll be strings coming in from MOLCodesignChecker and need to be converted back to strings in several other places (syncservice, event database) so the type is only going to ensure that the code in this helper is used in the few places it should be and will make this change significantly bigger for a pretty small gain.

russellhancox (approved):

PR #1370, opened 2024-06-18
WIP - proposals for mapping login/logout events into Santa telemetry

mlw:
Sourcing some feedback. Two proposals outlined here.

Method 1: Breaking each event into its own type. Since each type is independent there is no ability to map in data to events that wouldn't otherwise support (e.g. logout events don't have failure messages). But this does lead to some duplication and/or extremely similar message types.

Method 2: Top level messages are shared with an enum specifying the specific action (e.g. login/logout attach/detach, etc.). This reduces the number of message types and is maybe easier to read but supports invalid mappings.

My preference is Method 1, it's a much cleaner approach, but looking for feedback. I'm also curious if anyone has more experience with how protos are more broadly defined and how well each of these methods align with the standard way of doing things.

### DIFF
--- a/Source/common/santa.proto
+++ b/Source/common/santa.proto
@@ -538,6 +538,229 @@ message FileAccess {
   optional PolicyDecision policy_decision = 6;
 }
 
+////////////////////
+//
+// New Shared Types
+//
+////////////////////
+
+message GraphicalSession {
+  optional uint32 id = 1;
+}
+
+message Address {
+  optional bytes address = 1;
+
+  enum AddressType {
+    ADDRESS_TYPE_UNKNOWN = 0;
+    ADDRESS_TYPE_IPV4 = 1;
+    ADDRESS_TYPE_IPV6 = 2;
+    ADDRESS_TYPE_NAMED_SOCKET = 3;
+  }
+
+  AddressType type = 2;
+}
+
+////////////////////
+//
+// START OPTION 1
+//
+////////////////////
+
+message LoginWindowSessionLogin {
+  optional UserInfo user = 1;
+  optional GraphicalSession graphical_session = 2;
+}
+
+message LoginWindowSessionLogout {
+  optional UserInfo user = 1;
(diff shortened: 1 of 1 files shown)


pmarkowsky:
My vote is for **Method 1** it's more friendly as it's adding a new field and any old parsers based on the old proto without that field will just ignore it.  

The ergonomics on using the **method 1** types feel more tailored to what you want to do.

Source/common/santa.proto
```diff
@@ -538,6 +538,229 @@ message FileAccess {
...
+}
+
+message ScreenSharing {
+  oneof event {
+    ScreenSharingAttach attach = 1;
+    ScreenSharingDetach detach = 2;
+  }
+}
+
+message OpenSSHLogin {
```
russellhancox, line 639: I think SSHLogin, SSHLogout and SSH (or SSHEvent) would be good here, the "Open" feels needless

Source/common/santa.proto
```diff
@@ -538,6 +538,229 @@ message FileAccess {
...
+//
+// New Shared Types
+//
+////////////////////
+
+message GraphicalSession {
+  optional uint32 id = 1;
+}
+
+message Address {
```
russellhancox, line 551: Perhaps SocketAddress?

russellhancox (commented):
I much prefer Method 1.

Source/common/santa.proto
mlw: Mentioning for posterity - this is specific to OpenSSH that ships on the system, not SSH generically. I don't have a strong preference here and think its fair to drop the prefix, but there is a chance for confusion. This is maybe more just a note to make sure we're clear in documentation about the limitation of this event.

russellhancox: Sticking to OpenSSH is fine, I retract my comment

mlw:
superseded by #1371

PR #1371, opened 2024-06-20
Login/logout events

mlw:
This PR sets is the first step in adding support for the following event types:

```
ES_EVENT_TYPE_NOTIFY_LW_SESSION_LOGIN
ES_EVENT_TYPE_NOTIFY_LW_SESSION_LOGOUT
ES_EVENT_TYPE_NOTIFY_LW_SESSION_LOCK
ES_EVENT_TYPE_NOTIFY_LW_SESSION_UNLOCK
ES_EVENT_TYPE_NOTIFY_SCREENSHARING_ATTACH
ES_EVENT_TYPE_NOTIFY_SCREENSHARING_DETACH
ES_EVENT_TYPE_NOTIFY_OPENSSH_LOGIN
ES_EVENT_TYPE_NOTIFY_OPENSSH_LOGOUT
ES_EVENT_TYPE_NOTIFY_LOGIN_LOGIN
ES_EVENT_TYPE_NOTIFY_LOGIN_LOGOUT
```

The string and protobuf serializer implementations have been relegated into a separate branch for a future PR to try and keep this already large PR a bit smaller. So while this branch will subscribe to the new events, nothing is actually logged yet.


Source/santad/Logs/EndpointSecurity/Serializers/BasicString.mm
```diff
@@ -168,6 +179,64 @@ static inline void AppendUserGroup(std::string &str, const audit_token_t &tok,
...
+}
+
+static inline void AppendSocketAddress(std::string &str, es_address_type_t type,
+                                       es_string_token_t addr) {
+  str.append("|address_type=");
+  switch (type) {
+    case ES_ADDRESS_TYPE_NONE: str.append("none"); break;
+    case ES_ADDRESS_TYPE_IPV4: str.append("ipv4"); break;
+    case ES_ADDRESS_TYPE_IPV6: str.append("ipv6"); break;
+    case ES_ADDRESS_TYPE_NAMED_SOCKET: str.append("named_socket"); break;
```
russellhancox, line 213: WDYT about using `unix` or `uds` here instead of `named_socket`?

russellhancox (approved):

Source/common/santa.proto
```diff
@@ -565,6 +694,10 @@ message SantaMessage {
     Allowlist allowlist = 20;
     FileAccess file_access = 21;
     CodesigningInvalidated codesigning_invalidated = 22;
+    LoginWindowSession login_window_session = 23;
+    LoginLogout login_logout = 24;
+    ScreenSharing screen_sharing = 25;
+    OpenSSH ssh = 26;
```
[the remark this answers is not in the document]
mlw: Good call, will change

Source/santad/Logs/EndpointSecurity/Serializers/BasicString.mm
mlw: I'm fine with using the full `unix_domain_socket` name, will change.

tburgin: ES calls it `named_socket`, I think matching that makes the most sense.
https://developer.apple.com/documentation/endpointsecurity/es_address_type_t?language=objc

If we were to deviate, maybe `local_socket`?
From `man 2 socket`: `PF_LOCAL        Host-internal protocols, formerly called PF_UNIX`

tburgin (approved):

Source/santad/Logs/EndpointSecurity/Serializers/BasicString.mm
pmarkowsky: I agree staying close to the ES data here makes sense. Though it'll require some documentation.

mlw: Keeping the name to match closely with ES.

Source/santad/Logs/EndpointSecurity/Serializers/ProtobufTest.mm
```diff
@@ -93,19 +94,35 @@ JsonPrintOptions DefaultJsonPrintOptions() {
   return options;
 }
 
-NSString *EventTypeToFilename(es_event_type_t eventType) {
```
mlw, line 96: Note: Some of the diff here was due to the full serializers being implemented and tested. While the changes here to use variants are used yet in this PR, they will be used in the next PR. It became a bit too unwieldy to break this out so it was left in.

russellhancox (changes requested):

MODULE.bazel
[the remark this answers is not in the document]
mlw: This was added in automatically by the fix.sh script. For some reason it didn't like the extra newline. I got it to keep the sections by adding comments tho...

russellhancox (approved):

PR #1388, opened 2024-07-08
fileinfo: add --filter-inclusive

tburgin:
`--filter-inclusive`: If multiple filters are specified, they must all be true for the file to be displayed.

I also added a TODO to make `-printInfoForFile:` testable.


russellhancox (approved):

Source/santactl/Commands/SNTCommandFileInfo.m
```diff
@@ -189,6 +190,8 @@ + (NSString *)longHelpText {
                 @"              case-insensitive regular expression which must match anywhere in\n"
                 @"              the keyed property value for the file's info to be displayed.\n"
                 @"              You may specify multiple filters by repeating this flag.\n"
+                @"    --filter-inclusive: If multiple filters are specified, they must all be \n"
```
mlw, line 193: It'd be nice to mention here, or maybe above in the `--filter` help text that the default is for any** supplied filter to be true to be considered a match

tburgin: Good idea, done.

mlw (approved):

PR #1389, opened 2024-07-09
sync: Handle parse errors, make some preflight fields optional

russellhancox:


Source/santasyncservice/SNTSyncPreflight.mm
```diff
@@ -135,7 +135,11 @@ - (BOOL)sync {
       resp.deprecated_transitive_whitelisting_enabled());
   self.syncState.enableAllEventUpload = @(resp.enable_all_event_upload());
   self.syncState.disableUnknownEventUpload = @(resp.disable_unknown_event_upload());
-  self.syncState.eventBatchSize = resp.batch_size();
+
+  self.syncState.eventBatchSize = 50;
```
mlw, line 139: Is this some known/previous default? Can we make it a constant and comment it?

mlw (changes requested):

Source/santasyncservice/SNTSyncPreflight.mm
russellhancox: Done

Source/santasyncservice/syncv1.proto
```diff
@@ -54,7 +54,11 @@ message PreflightRequest {
 }
 
 message PreflightResponse {
```
pmarkowsky, line 56: Long as you're in here. We should add an optional machine_id field for GRPC.

russellhancox: As discussed, let's do this in a separate PR

PR #1390, opened 2024-07-09
Add machine_id to facilitate a GRPC version of the sync protocol

pmarkowsky:
This PR adds a `machine_id` field to all of the Request messages in the proto version of the sync protocol. 

This is to enable a GRPC based version of the sync protocol.

### DIFF
--- a/Source/santasyncservice/syncv1.proto
+++ b/Source/santasyncservice/syncv1.proto
@@ -199,7 +199,7 @@ message Event {
 
 message EventUploadRequest {
   repeated Event events = 1;
-  // The UUID of the machine that generated this event.
+  // The UUID of the machine where the event(s) occurred
   string machine_id = 2;
 }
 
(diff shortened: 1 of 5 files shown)


Source/santasyncservice/syncv1.proto
```diff
@@ -197,6 +199,8 @@ message Event {
 
 message EventUploadRequest {
   repeated Event events = 1;
+  // The UUID of the machine that generated this event.
```
russellhancox, line 202: nit: There could be multiple events. Maybe the comment can be `The UUID of the machine making the request` for all of the request types?

pmarkowsky: What about `The UUID of the machine where the event(s) occurred`?

russellhancox: That works too

russellhancox (approved):

PR #1392, opened 2024-07-10
Handle non-200 HTTP responses in SNTSyncStage performRequest

bugos:
If we receive a non-200 HTTP response, we should return an error instead of parsing the response to an empty protobuf message.

### DIFF
--- a/Source/santasyncservice/SNTSyncStage.mm
+++ b/Source/santasyncservice/SNTSyncStage.mm
@@ -220,7 +220,7 @@ using santa::NSStringToUTF8String;
                     timeout:(NSTimeInterval)timeout {
   NSError *error;
   NSData *data = [self dataFromRequest:request timeout:timeout error:&error];
-  if (*error != NULL) {
+  if (error) {
     SLOGE(@"Error performing request: %@", error.localizedDescription);
     return error;
   }
--- a/Source/santasyncservice/SNTSyncStage.mm
+++ b/Source/santasyncservice/SNTSyncStage.mm
@@ -159,9 +159,11 @@ using santa::NSStringToUTF8String;
                                  }];
 }
 
-- (NSData *)dataFromRequest:(NSURLRequest *)request timeout:(NSTimeInterval)timeout {
+- (NSData *)dataFromRequest:(NSURLRequest *)request
+                    timeout:(NSTimeInterval)timeout
+                      error:(NSError **)error {
   NSHTTPURLResponse *response;
-  NSError *error;
+  NSError *requestError;
   NSData *data;
 
   int maxAttempts = 5;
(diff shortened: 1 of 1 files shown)


russellhancox (approved):

Source/santasyncservice/SNTSyncStage.mm
```diff
@@ -211,8 +218,19 @@ - (NSData *)dataFromRequest:(NSURLRequest *)request timeout:(NSTimeInterval)time
 - (NSError *)performRequest:(NSURLRequest *)request
                 intoMessage:(google::protobuf::Message *)message
                     timeout:(NSTimeInterval)timeout {
-  NSData *data = [self dataFromRequest:request timeout:timeout];
-  if (data.length == 0) return nil;
+  NSError *error;
+  NSData *data = [self dataFromRequest:request timeout:timeout error:&error];
+  if (*error != NULL) {
```
russellhancox, line 223: s/NULL/nil

russellhancox (changes requested):

Source/santasyncservice/SNTSyncStage.mm
tburgin, line 223: No need to deference the `NSError *`.
`if (error) {`

mlw (approved):

Source/santasyncservice/SNTSyncStage.mm
mlw: Went with @tburgin 's suggestion

mlw: Fixed

tburgin (approved):

Source/santasyncservice/SNTSyncStage.mm
```diff
@@ -211,8 +218,19 @@ - (NSData *)dataFromRequest:(NSURLRequest *)request timeout:(NSTimeInterval)time
...
-  NSData *data = [self dataFromRequest:request timeout:timeout];
-  if (data.length == 0) return nil;
+  NSError *error;
+  NSData *data = [self dataFromRequest:request timeout:timeout error:&error];
+  if (error) {
+    SLOGE(@"Error performing request: %@", error.localizedDescription);
+    return error;
+  }
+  if (data.length == 0) {
+    NSString *errStr = @"Response is empty";
```
russellhancox, line 228: - [ ]

Source/santasyncservice/SNTSyncStage.mm
```diff
@@ -211,8 +218,19 @@ - (NSData *)dataFromRequest:(NSURLRequest *)request timeout:(NSTimeInterval)time
...
+    SLOGE(@"%@", errStr);
+    return [NSError errorWithDomain:@"com.google.santa.syncservice"
```
russellhancox, line 230: Need to fix the test to account for this: SNTSyncTest.mm line 263

mlw: Will fix in an immediate followup PR

russellhancox (approved):

PR #1393, opened 2024-07-10
common: Remove debug log when signing ID is missing.

russellhancox:
This log line gets spit out in santactl fileinfo output in a way that makes the command harder to use for various things


pmarkowsky (approved):

tburgin (approved):
Maybe just wait to merge until pull 1392 lands so we don't have to rebase there.

PR #1394, opened 2024-07-10
Allow empty data for 200 responses

mlw:
Followup to #1392

### DIFF
--- a/Source/santasyncservice/SNTSyncStage.mm
+++ b/Source/santasyncservice/SNTSyncStage.mm
@@ -225,11 +225,7 @@ using santa::NSStringToUTF8String;
     return error;
   }
   if (data.length == 0) {
-    NSString *errStr = @"Response is empty";
-    SLOGE(@"%@", errStr);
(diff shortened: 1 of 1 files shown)


russellhancox (approved):

Source/santasyncservice/SNTSyncStage.mm
```diff
@@ -225,11 +225,7 @@ - (NSError *)performRequest:(NSURLRequest *)request
     return error;
   }
   if (data.length == 0) {
-    NSString *errStr = @"Response is empty";
-    SLOGE(@"%@", errStr);
-    return [NSError errorWithDomain:@"com.google.santa.syncservice"
-                               code:4
-                           userInfo:@{NSLocalizedDescriptionKey : errStr}];
+    return nil;
```
tburgin, line 228: nit: `if (data.length == 0) return nil;`

mlw: I'm anti this!  Let's debate offline :)

tburgin (approved):

PR #1395, opened 2024-07-16
Improve handling of sync response default values

mlw:
This change makes it so that nearly all sync protocol response fields are not required, and defines action taken by Santa when given keys are not set. It additionally standardizes more configuration into enum fields.

### DIFF
--- a/Source/santasyncservice/syncv1.proto
+++ b/Source/santasyncservice/syncv1.proto
@@ -56,6 +56,44 @@ message PreflightRequest {
   string machine_id = 18;
 }
 
+enum SyncType {
+  option allow_alias = true;
+
+  // An unspecified SyncType will fallback to default "NORMAL" handling
+  SYNC_TYPE_UNSPECIFIED = 0;
+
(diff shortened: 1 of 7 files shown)


Source/santasyncservice/syncv1.proto
```diff
@@ -56,39 +56,71 @@ message PreflightRequest {
...
+
+  // An unspecified SyncType will fallback to default "NORMAL" handling
+  SYNC_TYPE_UNSPECIFIED = 0;
+
+  // Standard, progressive sync
+  NORMAL = 1;
+  normal = 1 [deprecated=true];
+
+  // Delete all non-transitive, previously received rules before applying the newly received rules.
+  CLEAN = 2; // [json_name="CLEAN"],
```
russellhancox, line 70:
```suggestion
  CLEAN = 2;
```

russellhancox (changes requested):

russellhancox (approved):

PR #1396, opened 2024-07-16
GUI: Update activation policy for binary blocks

russellhancox:


tburgin (approved):

mlw (approved):

tburgin (changes requested):
Switching the `NSApp.activationPolicy` quickly can result in a leaked icon on the dock...let's think about this some more.

Source/gui/SNTNotificationManager.m
```diff
@@ -125,6 +127,8 @@ - (void)queueMessage:(SNTMessageWindowController *)pendingMsg {
   [self.pendingNotifications addObject:pendingMsg];
 
   if (!self.currentWindowController) {
+    // Add app to Cmd+Tab and Dock.
+    NSApp.activationPolicy = NSApplicationActivationPolicyRegular;
```
tburgin, line 131: I think `queueMessage:` is called from not-the-main-thread. Does NSApp care?

tburgin: To that point, `self.pendingNotifications` seems to be accessed and written to on different threads. We should also fix that.

russellhancox: NSApp doesn't seem to care but the threading issue is a bit surprising. It looks like only queueMessage: is called from a non-main thread so I've moved much of it into a dispatch block.

tburgin: Nice.

tburgin (approved):

mlw (approved):

PR #1401, opened 2024-07-25
At last! A Big Sur Santa Logo.

Mpro256:
This was designed by me (czcx) in Photoshop, using the old logo, removing the edges, and making it the big sur style.
I have done this before, except using someone elses logo, this time it is mine.

I hope you like it!


Mpro256:
Note: The red I used is darker, which I think I like more then the previous logos brighter red. Let me know your thoughts on the color swap as well.

russellhancox:
This is pretty good, I have a few minor requests:

1. Please provide the SVG file so that if we choose to make changes in the future it will be easier
2. Please make the icon a little sharper. The previous icon is a little blurry because it's 10 years old and was made before Retina displays were commonplace and unfortunately I lost the original SVG file.
3. I don't have strong feelings about the color change but nearly all of the Apple icons have a gradient to them (usually getting darker toward the bottom). Could we have something similar, going from the brighter red of the original down to the darker red of the new icon? I think this would make it "pop" a little more than it does at present.

disenchant:
@russellhancox looking for something like this? 😃 

santa-sleigh

Mpro256:
Sorry for the EXTREMELY late response. Here is the .svg file.

356355983-7c8ab51e-7964-4bc5-af31-99eb7df9fd8c
(tysm @disenchant!)

PR #1402, opened 2024-07-25
sync: Handle missing error string for abnormal statuses

russellhancox:


Source/santasyncservice/SNTSyncStage.mm
```diff
@@ -202,7 +202,9 @@ - (NSData *)dataFromRequest:(NSURLRequest *)request
       errStr = [NSHTTPURLResponse localizedStringForStatusCode:response.statusCode];
     } else {
       code = (long)requestError.code;
+      if (code == 0) code = response.StatusCode;
       errStr = requestError.localizedDescription;
+      if (!errStr) errStr = @"Unknown Error";
     }
     LOGE(@"HTTP Response: %ld %@", code, errStr);
     if (error != NULL) {
```
tburgin, line 210: As a backstop we should also check `errStr` is not nil before creating `@{NSLocalizedDescriptionKey : errStr}`.

Source/santasyncservice/SNTSyncStage.mm
```diff
@@ -202,7 +202,9 @@ - (NSData *)dataFromRequest:(NSURLRequest *)request
       errStr = [NSHTTPURLResponse localizedStringForStatusCode:response.statusCode];
     } else {
       code = (long)requestError.code;
+      if (code == 0) code = response.StatusCode;
       errStr = requestError.localizedDescription;
+      if (!errStr) errStr = @"Unknown Error";
```
tburgin, line 207: If we are setting `response.StatusCode`, why not also get its `[NSHTTPURLResponse localizedStringForStatusCode:response.statusCode]`

tburgin (approved):

PR #1405, opened 2024-07-29
santad: Synchronize access to metric callback array

### LINKED ISSUE #1403: Metric set export doesn't synchronize access to registered callbacks
The `SNTMetricSet` `export` method doesn't synchronize access to the `_callbacks` array. This can lead to a crash if the array is modified while being iterated.

This has been seen during Santa startup.

1. Santad.mm - calls `metrics->StartPoll();` which will trigger `[metric_set export]` in the very near future
2. Santad.mm - then inits `SNTEndpointSecurityDeviceManager` and `SNTEndpointSecurityFileAccessAuthorizer` which both register a callback via `[[SNTMetricSet sharedInstance] registerCallback:...`
3. Metrics.mm - timer fires which triggers `[metric_set export]`
4. SNTMetricSet.m - `export` iterates callbacks without a lock

If the iteration in step 4 (which occurs on a separate thread) happens at the same time as one of the callbacks being registered in step 2, then Santa will likely crash.

As currently written, this race only occurs during Santa startup and is normally rare. If Santa does crash, it will be restarted by launchd and hopefully not lose the race again.

russellhancox:
Fixes #1403, I think

### DIFF
--- a/Source/common/SNTMetricSet.m
+++ b/Source/common/SNTMetricSet.m
@@ -605,10 +605,15 @@ NSString *SNTMetricMakeStringFromMetricType(SNTMetricType metricType) {
 
 /** Export current state of the SNTMetricSet as an NSDictionary. */
 - (NSDictionary *)export {
-  NSDictionary *exported = nil;
+  NSDictionary *exported;
+
+  NSArray *callbacks;
+  @synchronized(self) {
+    callbacks = [_callbacks mutableCopy];
+  }
 
   // Invoke callbacks to ensure metrics are up to date.
-  for (void (^cb)(void) in _callbacks) {
+  for (void (^cb)(void) in callbacks) {
     cb();
   }
 


tburgin (approved):

Source/common/SNTMetricSet.m
```diff
@@ -605,10 +605,15 @@ - (void)addConstantBooleanWithName:(NSString *)name
...
+  NSArray *callbacks;
+  @synchronized(self) {
+    callbacks = [_callbacks mutableCopy];
+  }
 
   // Invoke callbacks to ensure metrics are up to date.
-  for (void (^cb)(void) in _callbacks) {
+  for (void (^cb)(void) in callbacks) {
     cb();
   }
```
pmarkowsky, line 618: Shouldn't this be:

```c++
@synchronized(self) {
  callbacks = [_callbacks mutableCopy];
  
   for (void (^cb)(void) in callbacks) {
    cb();
   }
}
```

russellhancox: That would hold a lock while we're calling each callback with no control over what they're doing, so not only could they take quite a while to run they could themselves try to grab locks of some kind. The race seems to be in adding callbacks to the array while we're iterating it. Copying the array avoids this problem as long as the callbacks themselves don't change (which they shouldn't be able to as blocks).

PR #1406, opened 2024-07-30
sync: Upgrade from SCNetworkReachability -> nw_path_monitor

russellhancox:
SCNetworkReachability is deprecated. It also causes occasional crashes.

### DIFF
--- a/Source/santasyncservice/SNTSyncFCM.m
+++ b/Source/santasyncservice/SNTSyncFCM.m
@@ -14,7 +14,7 @@
 
 #import "Source/santasyncservice/SNTSyncFCM.h"
 
-#import <SystemConfiguration/SystemConfiguration.h>
+#import <Network/Network.h>
 
 #import <MOLAuthenticatingURLSession/MOLAuthenticatingURLSession.h>
 
@@ -80,7 +80,7 @@ static const uint32_t kDefaultConnectDelayMaxSeconds = 10;
 @property(copy, nonatomic) SNTSyncFCMMessageHandler messageHandler;
 
 /**  Is used throughout the class to reconnect to FCM after a connection loss. */
-@property SCNetworkReachabilityRef reachability;
+@property nw_path_monitor_t pathMonitor;
 
 /**  FCM client identities. */
 @property(nonatomic, readonly) NSString *project;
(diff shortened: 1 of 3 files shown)


Source/santasyncservice/SNTSyncFCM.m
```diff
@@ -212,24 +203,13 @@ - (void)reachabilityRestored {
...
-  if (self.reachability) return;
   LOGD(@"Reachability started.");
-  self.reachability =
-    SCNetworkReachabilityCreateWithName(kCFAllocatorDefault, _connectComponents.host.UTF8String);
-  SCNetworkReachabilityContext context = {.info = (__bridge void *)self};
-  if (SCNetworkReachabilitySetCallback(self.reachability, reachabilityHandler, &context)) {
-    SCNetworkReachabilitySetDispatchQueue(
-      self.reachability, dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0));
-  }
+  nw_path_monitor_start(self.pathMonitor);
```
tburgin, line 207: Do we know what happens if `nw_path_monitor_start` is called multiple times?

Source/santasyncservice/SNTSyncFCM.m
```diff
@@ -212,24 +203,13 @@ - (void)reachabilityRestored {
...
 }
 
 /**  Stop listening for network state changes. */
 - (void)stopReachability {
-  if (self.reachability) {
-    SCNetworkReachabilitySetDispatchQueue(self.reachability, NULL);
-    if (self.reachability) CFRelease(self.reachability);
-    self.reachability = NULL;
-  }
+  nw_path_monitor_cancel(self.pathMonitor);
```
tburgin, line 212: Same Q here.

Source/santasyncservice/SNTSyncFCM.m
russellhancox: Good question, it's not documented anywhere so I tested and it turns out the monitor starts multiple times :-(

I've updated the code to only create the monitor in startReachability and to cancel it out in stopReachability

Source/santasyncservice/SNTSyncFCM.m
russellhancox: This is also undocumented and the behavior is also weird. It seems that if you "overcancel" the monitor gets broken and can't be started. I've changed this too.

Source/santasyncservice/SNTSyncFCM.m
```diff
@@ -212,24 +192,27 @@ - (void)reachabilityRestored {
 
 /**  Start listening for network state changes on a background thread. */
 - (void)startReachability {
-  if (self.reachability) return;
+  if (!self.pathMonitor) return;
```
tburgin, line 195: Shouldn't this be:

```
if (self.pathMonitor) return;
```

Source/santasyncservice/SNTSyncManager.m
```diff
@@ -413,34 +391,29 @@ - (void)setReachable:(BOOL)reachable {
...
-    SCNetworkReachabilityContext context = {
-      .info = (__bridge_retained void *)self,
-      .release = (void (*)(const void *))CFBridgingRelease,
-    };
-    if (SCNetworkReachabilitySetCallback(self->_reachability, reachabilityHandler, &context)) {
-      SCNetworkReachabilitySetDispatchQueue(
-        self->_reachability, dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0));
-    } else {
-      [self stopReachability];
+  self.pathMonitor = nw_path_monitor_create();
```
tburgin, line 396: Missing: `if (self.pathMonitor) return;`

Source/santasyncservice/SNTSyncFCM.m
russellhancox: Yeah, bad copy-pasta

Source/santasyncservice/SNTSyncManager.m
russellhancox: Done

tburgin (approved):

PR #1409, opened 2024-08-01
sync: Send a hash of all database rules at beginning and end of sync

russellhancox:


Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -589,4 +591,24 @@ - (BOOL)fillError:(NSError **)error code:(SNTRuleTableError)code message:(NSStri
   return rules;
 }
 
+- (NSString *)hashOfHashes {
+  __block CC_SHA256_CTX sha;
+  CC_SHA256_Init(&sha);
+
+  [self inDatabase:^(FMDatabase *db) {
+    FMResultSet *rs = [db executeQuery:@"SELECT * FROM rules"];
```
pmarkowsky, line 599: You might need to exclude transitive rules here as it'll change your hash of hashes and the server will have no idea that these exist.

mlw: Also any rules added manually while a sync server didn't exist (though I don't think we track such thing today). However this is likely niche...

mlw: Another thought - there's been requests and talks in the past about per-user rules. It would be good to have a bit of an idea about how something like those might be represented here if they're ever implemented.

russellhancox: @pmarkowsky Good call, I've excluded transitive rules

@mlw If a sync server is added the first sync afterward is a clean sync so the hashes shouldn't matter. For per-user rules those would be implemented on the server side anyway, so that _shouldn't_ matter here. The purpose of this is not necessarily to have the server calculate the hash and compare them but to store the hash sent in a postflight and then validate that the one sent in a preflight matches the stored value.

pmarkowsky: @mlw I'm assuming you're talking about the rumblings in the past that folks would like to block execution by the user e.g. root can run X but not user peterm. 

That'd have to be implemented in the client but is an extension to rules similar to #1200. 

Re: Hash of Hashes. Do you need to sort by identifier and policy to create a stable hash? Like if I get a rule out of order for some reason i'll end up with a different hash even though the set of policies is the same.

russellhancox: > Do you need to sort by identifier and policy to create a stable hash? Like if I get a rule out of order for some reason i'll end up with a different hash even though the set of policies is the same.

Right now this hashes based on insertion order, which is stable on the client at least. The server could ensure this is stable by always ordering by creation timestamp but I'm not really expecting servers to calculate this in the same way as the client but instead just check the hash doesn't change during preflight.

pmarkowsky: >Right now this hashes based on insertion order, which is stable on the client at least. The server could ensure this is stable by always ordering by creation timestamp but I'm not really expecting servers to calculate this in the same way as the client but instead just check the hash doesn't change during preflight.

That's good to know.  

Not saying you should do this as the hash is probably the right tradeoff for now, but previously, I looked into using an xorfilter/bloomfilter rather than a hash to avoid worrying about ordering e.g. this single header one https://github.com/FastFilter/xor_singleheader.

The advantage to this is that you're creating a set membership data structure so order becomes irrelevant. 

The process would just be:

1. Hash rules (policy, state, identifier) to a 64-bit integer (e.g. use murmur3)
2. Insert into the xorfilter
3. Serialize and ship the xorfilter to the server in postflight
4. Rehydrate the xorfilter on the server and then for all rules per user check if they're in the supplied xorfilter

Alternatively you could also receive a xorfilter from the server of all the expected rules and as you're scanning the Santa rule db check if those rules are supposed to be in the client DB. This would let you know exactly which rules were spurious.

russellhancox: I see 3 problems with that approach

1. Serializing and de-serializing the filter may require reimplementing the serialization - the Go implementation in particular doesn't support this
2. This sounds far more expensive on the server-side; what I've proposed requires saving a single extra field in postflight and doing a simple 64-byte string comparison during preflight. This would require looping over all rules in postflight to check that they're in the filter.
3. This will not detect extra rules that have been added. Reversing the process _would_ find extra rules (which is better, imo) but then not detect missing rules.

russellhancox: @mlw Gotcha. I don't think that should be an issue, whether we implement that as an extra field in a rule or via CEL, we would add the new field(s) to those included in each rules' hash.

Source/common/SNTFileInfo.mm
```diff
@@ -165,7 +164,7 @@ - (instancetype)initWithPath:(NSString *)path {
 - (void)hashSHA1:(NSString **)sha1 SHA256:(NSString **)sha256 {
   const int MAX_CHUNK_SIZE = 256 * 1024;  // 256 KB
   const size_t chunkSize = _fileSize > MAX_CHUNK_SIZE ? MAX_CHUNK_SIZE : _fileSize;
-  char *chunk = malloc(chunkSize);
+  char *chunk = (char *)malloc(chunkSize);
```
tburgin, line 167: Now that this is an .mm file, use c++ style castings.
https://google.github.io/styleguide/cppguide.html#Casting

Source/common/SNTRule.mm
```diff
@@ -228,8 +229,8 @@ - (instancetype)initWithCoder:(NSCoder *)decoder {
   self = [super init];
   if (self) {
     _identifier = DECODE(NSString, @"identifier");
-    _state = [DECODE(NSNumber, @"state") intValue];
-    _type = [DECODE(NSNumber, @"type") intValue];
+    _state = (SNTRuleState)[DECODE(NSNumber, @"state") intValue];
```
tburgin, line 232: c++ style casts

Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -589,4 +591,25 @@ - (BOOL)fillError:(NSError **)error code:(SNTRuleTableError)code message:(NSStri
...
+  __block CC_SHA256_CTX sha;
+  CC_SHA256_Init(&sha);
+
+  [self inDatabase:^(FMDatabase *db) {
+    FMResultSet *rs =
+      [db executeQuery:@"SELECT * FROM rules WHERE type!=?", @(SNTRuleStateAllowTransitive)];
+    while ([rs next]) {
+      SNTRule *r = [self ruleFromResultSet:rs];
+      NSString *digest = r.digest;
+      CC_SHA256_Update(&sha, digest.UTF8String, (CC_LONG)digest.length);
```
tburgin, line 604: If unicode characters are present, `digest.UTF8String` could be less than `digest.length`. Since these are hex digests that should never happen, but for robustness / completeness `NSStringToUTF8StringView` or `-lengthOfBytesUsingEncoding:` should be used.

Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -589,4 +591,22 @@ - (BOOL)fillError:(NSError **)error code:(SNTRuleTableError)code message:(NSStri
...
+- (NSString *)hashOfHashes {
+  __block uint64_t crc = 0LL;
+  [self inDatabase:^(FMDatabase *db) {
+    FMResultSet *rs =
+      [db executeQuery:@"SELECT identifier, state, type, timestamp FROM rules WHERE type!=?",
+                       @(SNTRuleStateAllowTransitive)];
+    while ([rs next]) {
+      char digest[128];
+      uint32_t len = snprintf(
+        digest, 128, "%s:%d:%d:%u", [rs stringForColumn:@"identifier"].UTF8String,
```
mlw, line 603: nit: `s/128/sizeof(digest)/`

mlw:
What do you think about adding the hash of hashes value to `santactl status` output? feels like it might be useful for debugging / sanity check purposes.

PR #1422, opened 2024-08-31
Fix: Change uint64 fields in syncv1.proto to uint32 for backwards compatibility

pmarkowsky:
Change the uint64 fields in the syncv1.proto to uint32 to ensure backwards compatibility.

This also updates the SNTSyncEventUpload code to use the uint32 values and updates sync protocol docs to define the integer fields as `uint32` explicitly.

### DIFF
--- a/Source/santad/EventProviders/SNTEndpointSecurityAuthorizer.mm
+++ b/Source/santad/EventProviders/SNTEndpointSecurityAuthorizer.mm
@@ -101,6 +101,7 @@ using santa::Message;
                               }];
 }
 
+// Entry point not async
 - (void)handleMessage:(Message &&)esMsg
    recordEventMetrics:(void (^)(EventDisposition))recordEventMetrics {
   if (unlikely(esMsg->event_type != ES_EVENT_TYPE_AUTH_EXEC)) {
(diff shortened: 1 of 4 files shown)


russellhancox (approved):

pmarkowsky:
Tested this with Moroz and it works.

clreinki:
Thank you so much - 2024.6 and .7 wouldn't work with Moroz and we were pulling our hair out trying to figure out why.  So glad this was fixed!

russellhancox:
Hey @clreinki! I'm sorry to hear you've been having trouble and glad the fix is working. 

I'd like to know if there's some way we could have communicated the possibility of this better to have saved you the frustration. We'd mentioned in the release notes for .6 that syncing had been changed and to open an issue if syncing seemed to not work any more. Do you look through the notes when downloading a release? If you do, was this mention not clear enough? Is there a better way we could have called this out that might have helped you?

clreinki:
No worries, @russellhancox.  Yes, I read the release notes but I was still trying to narrow down whether it was a Santa issue or a Moroz one.  We also use a customized build of Moroz so had to make sure our modifications weren't the issue.  Really, it boils down to we hadn't figured out where the failure was occurring yet.  I just didn't wanna be that person who opens a issue without all the relevant info!

russellhancox:
Thank you for the update, that makes sense.

> I just didn't wanna be that person who opens a issue without all the relevant info!

We appreciate that attitude 😄

PR #1428, opened 2024-10-02
Remove default root label values

bugos:


Source/santad/SNTApplicationCoreMetrics.mm
```diff
@@ -107,22 +107,17 @@ static void RegisterMemoryAndCPUMetrics(SNTMetricSet *metricSet) {
 }
 
 static void RegisterHostnameAndUsernameLabels(SNTMetricSet *metricSet) {
+  // Set default values for some system labels.
   NSString *hostname = [NSProcessInfo processInfo].hostName;
-
   [metricSet addRootLabel:@"host_name" value:hostname];
```
gnoack, line 112: Where do the root labels flow - does this break existing users?

PR #1439, opened 2024-12-02
Add null check in ProcessToken constuctor/destructor

bugos:


Source/santad/ProcessTree/process_tree.cc
```diff
@@ -308,9 +308,17 @@ Tokens
...
+  }
   tree_->RetainProcess(pids);
 }
 
-ProcessToken::~ProcessToken() { tree_->ReleaseProcess(pids_); }
+ProcessToken::~ProcessToken() {
+  if (tree_ == nullptr) {
+    return;
+  }
+  tree_->ReleaseProcess(pids_);
```
kubalaguna, line 321: If tree is nullptr, wouldn't this line result in null pointer exception? What we observe now is that we actually do execute ReleaseProcess method

kubalaguna: as discussed offline, this is fine

